PR #351, opened 2015-01-26
Issue349 storage dynamics

mwetter:
This replaces https://github.com/lbl-srg/modelica-buildings/pull/350. It contains all code of https://github.com/lbl-srg/modelica-buildings/pull/350 and additional changes.


Buildings/Fluid/Storage/StratifiedEnhanced.mo
```diff
@@ -52,6 +52,11 @@ The model requires at least 4 fluid segments. Hence, set <code>nSeg</code> to 4
 </html>", revisions="<html>
 <ul>
 <li>
+January 25, 2015, by Filip Jorissen:<br/>
```
Mathadon: @mwetter this still needs to be removed?

mwetter: @Mathadon It is fine to have both revision entries as they show the history of the changes.

Mathadon: @mwetter in that case the comment should be updated since ```tau``` is no longer propagated but set to 0?

PR #362, opened 2015-02-13
Fix typos

tbeu:
See https://github.com/vlajos/misspell_fixer


mwetter:
@tbeu : Thanks for this pull request! Would you mind running the same code on
https://github.com/iea-annex60/modelica-annex60
as many of these files will be overwritten the next time we merge `Annex60` into the `Buildings` library.

PR #367, opened 2015-02-27
Issue360 replaceable per

mwetter:
Merge replaceable performance data to `master`


mwetter:
This change has been added to https://github.com/iea-annex60/modelica-annex60/pull/153

PR #390, opened 2015-03-23
Added new icon that will be used by the models generated with OpenStudio...

mbonvini:
I created a new icon that can be extended by the models automatically generated with the openstudio to modelica ruby package.


mbonvini:
now on separate branch from commit 
de20b8fd106df5745af14c2c3271640521f64119

Issue #415, opened 2015-05-11
One phase resistive load model under electrical package (OpenModelica)


s-and:
Hi,
First of all, thank you very much for providing this library. I have been exploring the electrical package of this library and would like to especially thank Marco for his help so far.

There are several examples not working under OpenModelica (https://test.openmodelica.org/libraries/Buildings/BuildModelRecursive.html). I suspect that several of them involve the one phase resistive load model. For instance, to get the storage model working, I replaced the resistive load with a capacitive one and it worked.

I wonder what makes one phase resistive load model incompatible with OpenModelica as the rest of the one phase load models seem to work.

Thanks.

mbonvini:
Regarding some of the errors flagged by OpenModelica there is a ticket open that once fixed should address the problems related with index reduction (https://trac.openmodelica.org/OpenModelica/ticket/3305). The ticket has high priority so hopefully it will be fixed soon. However, if in the meantime this error appears a quick solution is setting this option in OM

```
--replaceHomotopy=actual
```

Regarding the difference between resistive and capacitive there isn't really a conceptual difference since the `CapacitiveLoad` model extends the same model used by the `ResistiveLoad` model, and actually it the capacitive one adds a layer of complexity. However this may lead to a possible workaround that makes the model work with OM.

I'll let you know if my idea works.

mbonvini:
@s-and - Thanks for this hint, I think I have a solution to the problem and soon should be in the master branch.

s-and:
Thanks for taking care of this.

mbonvini:
closed with https://github.com/lbl-srg/modelica-buildings/pull/416

PR #416, opened 2015-05-15
Tentative to fix issue #415. Adding a dummy interface for resistive models

mbonvini:
The simple addition of a new interface for resistive models (basically an empty wrapper of the basic Load interface) seems to help the OpenModelica parser.

This is supposed to fix issue #415.

### DIFF
--- a/Buildings/Electrical/AC/OnePhase/Loads/Resistive.mo
+++ b/Buildings/Electrical/AC/OnePhase/Loads/Resistive.mo
@@ -159,8 +159,10 @@ The choices are between a null current or the linearized model.
 </html>", revisions="<html>
 <ul>
 <li>May 14, 2015, by Marco Bonvini:<br/>
-Changed parent class to <code>ResistiveLoad</code> in order
-to help openmodelica parsing the model.
+Changed parent class to <a href=\"modelica://Buildings.Electrical.Interfaces.ResistiveLoad\">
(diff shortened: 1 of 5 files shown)


mbonvini:
@mwetter: the few edits I've done seem to help OpenModelica parse the models containing resistive loads (a lot). 
I tested it on my local OM installation (just on few of them) and it seems to work.
If you agree and we merge this into the master branch we can test it on the OM RecursiveBuild test.

mwetter:
@mbonvini :
Before merging, please address the following:


1. In the comments, 
"Changed parent class to <code>ResistiveLoad</code> in order
to help openmodelica parsing the model."
add a link to the OpenModelica ticket. Otherwise, one cannot reproduce what this fix is for, and consequently it will be very hard to remove it while being sure that nothing breaks.
Also, use the full modelica name, using 
```
<href=\"modelica://Buildings...
```
  Otherwise, renaming the class may not change the link above.

2. Also update the revision section of Buildings/Electrical/Interfaces/ResistiveLoad.mo
Simply writing "limitation" is not clear. The comment must say what did not work to allow checking whether the model works once the fix is reverted when OpenModelica corrected the bug fix.

3. Run all unit tests to make sure everything still works.

mbonvini:
I added the details in the documentation and run the unit test, it works.

PR #440, opened 2015-07-08
Removed heat flow boundary condition model

### LINKED ISSUE #439: CFD examples and tutorial declares model for zero heat flow boundary condition
Some examples for the CFD model implement a prescribed heat flow rate boundary model and set `Q_flow=0`. This model can be removed because the connection semantics of Modelica ensures that `Q_flow=0`.

mwetter:
This will fix #439.

@zuowangda : Can you please review this change and let me know if it is OK. The regression tests give the same results.


zuowangda:
Yes, it is ok.

Issue #527, opened 2016-06-09
Remove assert statement in FlowSplitter_u


tsnouidui:
This issue is to remove an `assert()` statement in `FlowSplitter_u (Fluid.FMI)`. If this model is exported as an FMU, and is part of a system model (e.g. with other FMUs), it might trigger the `assert()` statement depending on the execution sequence of the FMUs.

mwetter:
Closed as it is on the master.

PR #528, opened 2016-06-09
Remove assert statement.

tsnouidui:
This addresses issue #527.

### DIFF
--- a/Buildings/Fluid/FMI/FlowSplitter_u.mo
+++ b/Buildings/Fluid/FMI/FlowSplitter_u.mo
@@ -67,12 +67,6 @@ equation
 
   // Conditional connector
   connect(bacPro_internal, inlet.backward);
-  // Stop if mass is not conserved in the system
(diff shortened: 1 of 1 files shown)


mwetter:
@tsnouidui Please add also revision notes in `FlowSplitter_u.mo`

Buildings/Fluid/FMI/FlowSplitter_u.mo
```diff
@@ -148,6 +148,13 @@ the model stops with an error.
 </html>", revisions="<html>
 <ul>
 <li>
+June 9, 2016, by Thierry S. Nouidui:
+</li>
```
mwetter: @tsnouidui This should be `<br/>` and not a new list item. See the other entries for formatting.

tsnouidui: @mwetter This is fixed now. I also checked whether html is validated.

PR #561, opened 2016-10-06
minor HTML fixes

thorade:
just some minor HTML fixes,
found using http://htmlhint.com/


mwetter:
@thorade Thanks for the changes.

PR #575, opened 2016-11-08
Updated 563-entropy to recreate 2nd law violation problem for low-lift chillers

rzpass:
Updated the Modelica 563-entropy to reproduce 2nd law violation problem. Changed two things:
1) The entropy balance was actually computed backward (need entropy out - entropy in), not opposite.
2) There was zero flow through the condensor in either chiller. Updated m_flow to m1_flow_nominal for both. Now problem is recreated.

-Rebecca


mwetter:
@rzpass 
Instead of putting the new files to the top-level directory, can you please move them to where the original files are, and commit the changes. You should then see on this pull request the changes that you did when you select "Files changed", e.g., a line-by-line diff that can be used to see what changed. Right now, it looks like you added 8 new files and didn't modify the old files.

mwetter:
I close this as it has been integrated in https://github.com/iea-annex60/modelica-annex60/issues/497, commit https://github.com/iea-annex60/modelica-annex60/commit/e67876cad58e618aff4eb27e067901c676ca18f7

Issue #614, opened 2017-01-05
RadiantSlab models


tsnouidui:
All models which use `Buildings.Fluid.HeatExchangers.RadiantSlabs.ParallelCircuitsSlab` cannot translate in JModelica.

tsnouidui:
This is now fixed and is already on the master.

PR #615, opened 2017-01-05
Refactor for JModelica translation.

tsnouidui:
This is for #614. This addresses  an `Array index out of bounds: 3, index expression: j` compilation issue in JModelica.
This however does not cause the model to compile as JModelica reports following error:
> Error: in file 'null':
> Semantic error at line 0, column 0: Index reduction failed: Maximum number of differentiations has been reached.
The refactoring is however better as it does not expose the parameter `nSta2` anymore.
The refactoring does not change any reference results.

### DIFF
--- a/Buildings/HeatTransfer/Data/OpaqueConstructions.mo
+++ b/Buildings/HeatTransfer/Data/OpaqueConstructions.mo
@@ -11,8 +11,9 @@ package OpaqueConstructions
       annotation (choicesAllMatching=true, Evaluate=false, Placement(transformation(extent={{60,60},{80,80}})));
    final parameter Real R(unit="m2.K/W")=sum(material[i].R for i in 1:nLay)
       "Thermal resistance per unit area";
-   parameter Integer nSta[nLay](each min=1)={material[i].nSta for i in 1:nLay}
-      "Number of states (do not overwrite, used to work around Dymola 2017 bug)";
+    parameter Integer nSta[nLay](each min=1) = {material[i].nSta for i in 1:nLay}
+      "Number of states (do not overwrite, used to work around Dymola 2017 bug)"
+      annotation (HideResult=true, Dialog(enable=false, tab="Advanced"));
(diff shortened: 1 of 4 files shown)


Buildings/HeatTransfer/Data/OpaqueConstructions.mo
```diff
@@ -11,7 +11,8 @@ package OpaqueConstructions
       annotation (choicesAllMatching=true, Evaluate=false, Placement(transformation(extent={{60,60},{80,80}})));
    final parameter Real R(unit="m2.K/W")=sum(material[i].R for i in 1:nLay)
       "Thermal resistance per unit area";
-
+   parameter Integer nSta2[nLay](each min=1)={material[i].nSta for i in 1:nLay}
```
mwetter: This should be `final` and be renamed to `nSta` as we introduced `nSta2` to avoid a name clash, but this is not needed here.

tsnouidui: This shouldn't be set to `final` as they will be modified in `Buildings.Fluid.HeatExchangers.RadiantSlabs.SingleCircuitSlab`.

mwetter: But then the GUI will show this parameter as something the user needs to enter. Users will be confused if the see the default assignment of `nSta2`. Setting it final would avoid this. Is there another work-around?

tsnouidui: I couldn't come up with an alternative . We could add a comment along the line "(do not overwrite, used to work around Dymola 2017 bug)"...

tsnouidui:
@mwetter I added a comment for this parameter and fixed the other issues. Please review and merge.

mwetter:
@tsnouidui If you are fine with my edits 06fee3b, then please merge to the master.

tsnouidui:
Looks good.

PR #646, opened 2017-02-01
Issue546 CYMDIST

### LINKED ISSUE #546: Develop a model for interfacing a grid simulator
This issue is to create a model which can interface a grid simulator.

tsnouidui:
This is for the CYMDIST export. This commits introduces a dependency on Python 3.4. Ubuntu 16.04 doesn't have Python 3.4 by default. To run the unit test I had to install Python 3.4 from source following http://devmartin.com/blog/2016/04/creating-a-virtual-environment-with-python3.4-on-ubuntu-16.04-xenial-xerus/.
I then made a symbolic link to `libpython3.4m.so` and `libpython3.4m.so.1.0`which is the Python installation folder so the binaries could be found (`sudo ln -s /path-to-python/libpython3.4m.so /usr/lib/libpython3.4.so`, and `sudo ln -s /path-to-python/libpython3.4m.so.1.0 /usr/lib/libpython3.4m.so.1.0`). This commit hasn't changed any reference results. This closes #546.


Buildings/Utilities/IO/Python27/Functions/BaseClasses/package.order
```diff
@@ -1 +1 @@
-exchange
+exchange
```
mwetter: Correct line ending.

Buildings/Utilities/IO/Python27/Functions/Examples/package.order
```diff
@@ -1 +1 @@
-Exchange
+Exchange
```
mwetter: Correct line ending

PR #659, opened 2017-02-17
Issue587 low lift heat pump

MassimoCimmino:
Removed unneeded Sphynx documentation files. Updated release notes.


mwetter:
@MassimoCimmino : The Thumbs.db files should be ignored and removed as well.

MassimoCimmino:
@mwetter : They should be out now.

PR #664, opened 2017-02-25
Utilize timeScale parameter (of MSL 3.2.2) for tables

tbeu:


mwetter:
@tbeu Thanks for the changes. I will test in https://github.com/lbl-srg/modelica-buildings/pull/667 and then merge.

PR #667, opened 2017-02-27
Tbeu time scale

mwetter:
Merge after CI testing of https://github.com/lbl-srg/modelica-buildings/pull/664


tbeu:
Why did the results change?

mwetter:
@tbeu These are minor changes, is it possible that the numerics changed slightly? I will rerun the tests and then merge. Thanks for the changes.

tbeu:
The C sources were not touched at all for the scaling. The scaled floating-point numbers may have influence on the numerics - though all table data looked non-problematic when inspecting them. See https://github.com/modelica/Modelica/issues/1473 for the corresponding MSL issue.

PR #669, opened 2017-03-03
Issue668 wrong units in refrigerant properties

MassimoCimmino:


mwetter:
Thanks, that was fast.

PR #692, opened 2017-03-30
Issue653 carnot dt

FelixBue:
Implementation according to your suggestions (hopefully) with correct names of the parameters, default values according to the discussion I had with Emerson and depending on cp_default.


mwetter:
@FelixBue : Thanks, I will integrate it into Buildings and then also merge to the `IBPSA` library.

Issue #716, opened 2017-04-17
add default component name


JayHuLBL:
We need to add default component name to all CDL blocks.

PR #721, opened 2017-04-18
Added default component name for #716

JayHuLBL:
- [x] added default component name


Buildings/Resources/Scripts/Dymola/Experimental/OpenBuildingControl/CDL/Discrete/Examples/DayType.mos
```diff
@@ -1,4 +1,4 @@
-simulateModel("Buildings.Experimental.OpenBuildingControl.CDL.Discrete.Examples.DayType", startTime=-1.8144e+06, stopTime=1.8144e+06, method="dassl", resultFile="DayType");
+simulateModel("Buildings.Experimental.OpenBuildingControl.CDL.Discrete.Examples.DayType", startTime=-1.8144e+06, stopTime=1.8144e+06, method="dassl", resultFile="DayType1");
```
mwetter: The `resultFile` name needs to be the same as the name that is used in the plot command. If there is a name conflict, you can use for example `Examples.DayType` (or any other string). Otherwise, the plot won't display the result.

PR #738, opened 2017-04-25
Issue609 cdl update for 732 734

JayHuLBL:
- [x] Moved `Ramp.mo` to `CDL.Sources` package
- [x] Updated `CDL.Sources.Pulse` and `CDL.Sources.BooleanPulse` so to avoid using % but using fraction
- [x] Removed `CDL.Sources.DutyCycle`
- [x] Updated reference results after removing `DutyCycle`
- [x] Removed initialization type `SteadyState` to avoid conducting global analysis for `der(x)=0`


mwetter:
This is ready to merge when all tests pass

PR #752, opened 2017-05-05
Fix typos

tbeu:


mwetter:
@tbeu Thanks for all the fixes!

PR #771, opened 2017-05-22
Initialize Python interpreter only once

tbeu:
Binaries need to be regenerated then.


mwetter:
@tsnouidui : could you please process this pull request and regenerate the binaries as I am currently not set up for Windows.

tsnouidui:
@mwetter Yes I will do.

tsnouidui:
@mwetter I can't see and checkout this branch? Will this need to be a branch on the modelica-buildings repo? Easiest might be for me to create a new branch on the repo and implement the changes since they are minor.

PR #851, opened 2017-07-18
Issue637 scalable models

JayHuLBL:


mwetter:
@JayHuLBL 
I started revising the models but will need to continue after the BS'17 conference.
Could you please in the meantime go over the models once more and make sure that all classes have a documentation, as various package.mo files, and the schedules, were missing the info section. I fixed some of them.
Also, regarding structure, I suggest we put them to 
```
Buildings
  Examples
    ScalableBenchmarks
      BuildingVAV  // Everything should be in this package
        Controls
        HVAC
        ...
```
This way, we can have other scalable benchmarks in packages that are parallel to `BuildingVAV`.

Also, it is not clear to me why there are so many models in `Experimental/ScalableModels/ThermalZones/BaseClasses`. If we need all these room models, then it is fine to keep them, but please explain in the info section why there are so many models that are quite similar. Is there anything specific you want to test? If so, please explain in the info section so that others understand the rationale.

Please use shorter names, e.g., `Experimental/ScalableModels/ScalableDemos/ScalableDemo_1Floor_1Zone` states 3 times that it is a scalable model.

Best would be to have all runnable models in a `Validation` or `Examples` package as we did with other parts of the library.

PR #878, opened 2017-08-29
Issue859 dx air cooled data center

YangyangFu:


mwetter:
@YangyangFu 
I merged your code and further edited it.
In general, please follow the naming convention, e.g., don't use MAT, OA etc but rather TMix, TOut etc.

Please don't simply use an integer for the mode as this is error prone if we ever change the mode. I changed this to using Types.CoolingModes.

I also renamed the models to avoid names such as ASECoolingModeController as "ASE" is not very descriptive for non-experts.

PR #892, opened 2017-09-02
Issue888 data center control error

YangyangFu:
@mwetter @zuowangda This pull request solves the control errors of a DX cooling system with an airside economizer in #892  . The control errors are caused by inappropriate sizing and dead band settings for the control signal.


mwetter:
@YangyangFu : Could you please integrate your change to the new branch `issue888_dataCenter_controlError_refactor` that I just created as the structure of the packages changed substantially today.

PR #894, opened 2017-09-06
Issue842 trim respond

JayHuLBL:
Created Trim&Respond logic according to G36 Part5.A.14.


Buildings/Controls/OBC/ASHRAE/G36/Atomic/TrimRespondLogic.mo
```diff
@@ -0,0 +1,313 @@
...
+    annotation (Placement(transformation(extent={{-180,-100},{-160,-80}})));
+  CDL.Continuous.Add difReqIgnReq(k1=-1)
+    "Difference between ignored request number and the real request number"
+    annotation (Placement(transformation(extent={{-80,-80},{-60,-60}})));
+  CDL.Continuous.Add add1
+    "Increase setpoint by amount of value defined from reset logic"
+    annotation (Placement(transformation(extent={{-20,50},{0,70}})));
+  CDL.Continuous.Add add2 "Net reset value"
+    annotation (Placement(transformation(extent={{80,-140},{100,-120}})));
+  CDL.Continuous.Min min
```
milicag: Is this block name allowed? There is also a ```max``` block in line 98. Those strings might cause ambiguity since they also represent a mathematical operation. I don't know how Modelica handles this.

mwetter: @JayHuLBL  @milicag :  It is better not to call the instance ``min`` or ``max``. Modelica allows it, but when we convert it to other languages, this could cause problems.

Issue #866, opened 2017-08-15
Multizone VAV AHU Supply fan control


JayHuLBL:
Sequence for multizone VAV AHU supply fan control (G36, Part5.N.1)

PR #895, opened 2017-09-06
Issue838 t sup air multi zone

JayHuLBL:
The sequence defines supply air temperature control for multilane AHU, according to G36, Part5.N.2


Buildings/Controls/OBC/ASHRAE/G36/Atomic/SupplyAirTempSet_MultiZone.mo
```diff
@@ -0,0 +1,330 @@
...
+        iconTransformation(extent={{-140,-100},{-100,-60}})));
+  CDL.Interfaces.RealOutput TSup(
+    final unit="K",
+    quantity="ThermodynamicTemperature")
+    "Supply air temperature"
+    annotation (Placement(transformation(extent={{140,0},{160,20}}),
+        iconTransformation(extent={{100,-10},{120,10}})));
+
+
+//   fixme: Need a Trim&Response logic to find the T_max
```
mwetter: Address or remove the fixme

Buildings/Controls/OBC/ASHRAE/G36/Atomic/SupplyAirTempSet_MultiZone.mo
```diff
@@ -0,0 +1,330 @@
...
+    annotation (Placement(transformation(extent={{40,-60},{60,-40}})));
+  CDL.Continuous.Limiter TDea(
+    uMax=24 + 273.15,
+    uMin=21 + 273.15)
+    "Limiter that outputs the dead band value for the supply air temperature"
+    annotation (Placement(transformation(extent={{40,80},{60,100}})));
+  CDL.Logical.Switch swi3 "Check output regarding supply fan status"
+    annotation (Placement(transformation(extent={{100,20},{120,0}})));
+  CDL.Integers.LessThreshold intLesThr(threshold=Constants.OperationModes.warUpInd)
+    "Check if operation mode index is less than warm-up mode index (4)"
```
mwetter: It is better to not state the numerical value. Otherwise, if we ever need to change `Constants.OperationModes`, the comment will be wrong, and we probably won't detect it

Buildings/Controls/OBC/ASHRAE/G36/Atomic/SupplyAirTempSet_MultiZone.mo
```diff
@@ -0,0 +1,330 @@
+within Buildings.Controls.OBC.ASHRAE.G36.Atomic;
+block SupplyAirTempSet_MultiZone
```
mwetter: As this is only for VAV systems, the block should have `VAV` in its name, or somewhere in its package name.

milicag: @JayHuLBL could this block have a name equivalent to the single zone one (without the fan speed)?

JayHuLBL:
@mwetter Your comments have been addressed:
- [x] renamed the block
- [x] added TrimRespond logic (the same one used in #866) and implemented in the `VAVMultiZoneSupTempSet` block

mwetter:
@JayHuLBL : Please make the following changes, then it is ready to merge:
- [ ] Rename `VAVMultiZoneSupTempSet` to `VAVMultiZoneTSupSet`, as there is also a block `VAVSingleZoneTSupSet`. Then, they have the same name, except for 'Single' and 'Multi'.
- [ ] Sometimes we use TrimRespond while the documentation uses "trim-response". Please use consistently either "respond" or "response", whatever term Guideline 36 uses.

PR #869, opened 2017-08-22
Issue735 data center chiller wse

mwetter:
This closes #735

PR #898, opened 2017-09-06
Issue735 data center chiller wse

YangyangFu:
@mwetter @zuowangda I've solved the problems in #869 . Two major revisions include:

1. Add a hysteresis for the on/off signal of the shutoff valves in a pump parallel model;
2. Revise the `tWai` settings in different controllers to avoid huge controllers in supply air temperature and chilled water supply temperature.


mwetter:
@YangyangFu : I am still revising and editing your models, and I also removed the `Sign` block. I also made some other revisions to the controls. Can you please wait with these changes until I send you a note to fix the control overshoot of the room air temperature. Otherwise we have many conflicts. Best is probably if I close in the meantime this pull request, and then you can integrate the bug fix for the control using my latest revision.

PR #893, opened 2017-09-03
Issue889 hum

YangyangFu:
@mwetter @zuowangda  This pull request addresses the issue in #889. I changed the humidifier model from `SteamHumidifier_X` to `SprayAirWasher_X`, and re-implemented the `ReheatOnOffControl` using state graph.

Issue #889, opened 2017-09-01
Humidifier in DataCenters.HVAC.AHUs.CoolingCoilHumidifyingHeating


mwetter:
@zuowangda : 
`Buildings.Applications.DataCenters.HVAC.AHUs.CoolingCoilHumidifyingHeating`
has a `SteamHumidifier_X`. Shouldn't this be a `SprayAirWasher_X` as the steam humidifier would add additional cooling need compared to the adiabatic spray air washer.

If so, can you please update the model and documentation.

mwetter:
@YangyangFu : Please integrate your changes based on `issue889_humidifiers_datacenter` and make a pull request against this branch. This branch contains the refactored `DataCenter` package.

PR #902, opened 2017-09-08
Issue889 humidifiers_datacenter

YangyangFu:
@mwetter @zuowangda This pull request is based on pull request #893 to replace the `SteamHumidifer` model by `SprayWasher` model. This issue is shown in #889. And the model `Buildings.Applications.DataCenters.ChillerCooled.Controls.Reheat` is re-implemented using state graph, which makes this model easier to be understood.


YangyangFu:
@mwetter Dr. Wetter, you need merge this branch so that I can update #869 and #898 , which use the AHU model updated in this issue.

Issue #883, opened 2017-08-30
Restructure G36 package


mwetter:
Restructure the `G36` package to
```
G36
  TerminalUnit
    CoolingOnly.mo
    Subsequences
      xxx
  AHU
    Coils
      HeatingCoolingValves.mo
    Economizer
      SingleZone.mo
      MultiZone.mo
      ReliefDamper.mo
      Subsequences
        DamperLimitsSingleZone.mo
        DamperLimitsMultiZone.mo
    Setpoints
      TemperatureSpeedSingleZone.mo
      TemperatureMultiZone.mo
      SpeedMultiZone.mo
      ReliefFan.mo
      OperationMode.mo [remove "Selector"]
      TrimAndRespond.mo
      MinimumOutdoorAirSingleZone.mo
      MinimumOutdoorAirMultiZone.mo
```
once the additional sequences are merged to the master branch.

PR #911, opened 2017-09-09
Issue883 G36 Package Refactor

milicag:
**Not merge ready, see below**

This PR introduces a new structure to G36 package according to #883.

I'm committing the refactor script, ```Buildings/temp_refactor.py```, which we should remove before merge. The script currently handles all G36 modules present on master, PR #894 and PR #895.

To do:
* [ ] Any additional future merges should be considered and added to the script.
* [ ] Please also manually delete any remaining empty old packages and address ```fixme``` in any new package info sections.
* [ ] Check model names and make sure they are consistent throughout, also make sure that validation models include an indicator of the control variable(s).

I'm creating this PR ahead of time due to vacation till Sept 18], in case we need it before that.


JayHuLBL:
@mwetter a new test branch `issue883_restructure_g36_test` was created. The `temp_refactor.py` was updated. 
Previously, the restructuring was failed for the reason that the partial model path is used in model `Buildings.Controls.OBC.ASHRAE.G36.Composite.EconomizerMultiZone`, `Buildings.Controls.OBC.ASHRAE.G36.Composite.EconomizerSingleZone`: 
`Controls.OBC.ASHRAE.G36.Atomic.EconEnableDisableMultiZone ecoEnaDis`, ....

The renaming implementation cannot find the `EconEnableDisableMultiZone` and cannot update it.

mwetter:
Closed as this is merged through https://github.com/ibpsa/modelica-ibpsa/pull/824

PR #949, opened 2017-09-19
Refactored models for #947

### LINKED ISSUE #947: reduce number of events in ChillerDXHeatingEconomizer
The model `Buildings.Air.Systems.SingleZone.VAV.Examples.ChillerDXHeatingEconomizer` has a lot of state events that can probably be reduced by replacing the timer with a boolean delay.

mwetter:
This closes #947 

The timer and inequality check has been removed, and the hysteresis with delay implementation has been simplified. Also, some models in the `Air` package have been deleted as they are now in `CDL`.


mwetter:
- [ ] The validation tests `EconomizerMultiZone_Disable` fails the validation due to different damper position at the first transition of the test.

mwetter:
See comments of commit for differences in `EconomizerMultiZone_Disable`.
@JayHuLBL : Can you please modify the unit test so that the `delOutDamOsc` in `EconEnableDisableMultiZone` has a delay, e.g., you may need to change the input signals of the test so that the economizer opens and closes during the test.

- [ ] The inequalities in `EconDamperPositionLimitsSingleZone` and in `EconDamperPositionLimitsMultiZone` that are used to check the mode and freeze protection stage should be changed to check for integer equalities to make sure we don't incur overhead for the solver.

Issue #987, opened 2017-10-02
Change THeaSet to TSupSet for economizer


mwetter:
@milicag : The economizer sequences had a control input `THeaSet`, which I renamed to `TSupSet` because the same signal is also used during cooling. Can you please verify that I updated all signals and documentation, including the graphics, and update what is needed. For example, 
`Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.Economizers.Subsequences.Modulation` uses `THeaSet` in the sequence diagram.

milicag:
@mwetter Sounds good.

milicag:
@mwetter The multizone economizer should use ```TSupSet```, since the multizone ```TSupSet``` is a single value (without differentiation between heating and cooling). For the single zone economizer the value should be ```THeaSet```, per our discussion with Brent - the output is  indicated as an economizer input in the ```SingleZone.SetPoints.VAVSupply``` sequence. In single zone this is important since the coils are controling the ```TSup``` to ```TCoo``` (blue line), but the economizer will attempt to keep it lower, at ```THea```. Would you like me to proceed with making necessary corrections in line with this?

milicag:
```MultiZone```/```SingleZone``` naming convention - should we use ```DamperModulation``` or just ```Modulation```? Same question for ```DamperLimits```.

mwetter:
@milicag : I agree with your suggestion for the economizer (single zone) input `THeaSet`.

But `SingleZone.VAVSupply` should keep two temperature outputs, one for heating and one for cooling, as they are used for the heating and cooling coil.

In the `Economizer` package, you can delete the word `Damper`

milicag:
@mwetter Sounds good, I'll create a PR with that edit. 

I was just looking at the single zone room/supply air temperature structure. We've now corrected the inputs to ```AHUs.Valves.HeatingAndCooling``` to be the zone heating and cooling setpoints, which is consistent with G36. It is still a little unclear what is a correct ```TSupSet``` for the ```TerminalUnits.Reheat.DamperValve```` in case of a single zone AHU. This would be easier to clarify with a single zone closed loop demo.

mwetter:
@milicag : For a single zone VAV, there is no need for a terminal unit with reheat and damper. Hence I don't understand your question.

milicag:
OK, thanks.

PR #988, opened 2017-10-02
Issue987 supply air temperature as an input to economizer sequences

milicag:
This addresses #987. It only includes changes to variable and model names.


milicag:

PR #998, opened 2017-10-05
Changing solver for JModelica verification

tsnouidui:
@mwetter I changed the solver for JModelica verification.


mwetter:
@tsnouidui 
Can you please revise the request: One `.mos` file changed, but 7 reference results have changes. This is not correct.

tsnouidui:
I used Dymola 2018 FD01 (beta) to generate the reference results and I think the new compiler backend leads to smaller number of system of equations.
Should I revert and use older compilers (2017 FD01 or 2018)?

mwetter:
Please use 2017FD01, as this is what is used on travis.

PR #1000, opened 2017-10-06
Updated models for JModelica

tsnouidui:
@mwetter @JayHuLBL  For JModelica verification, I changed the solver of the scalable benchmark examples to use `Cvode`.
The simulation is way faster.
For the 
`Buildings.Examples.ScalableBenchmarks.BuildingVAV.Examples.TwoFloor_TwoZone`
using `Radau`, the simulation time is 227s.
using `Cvode`, the simulation time is 107s
I then activated parallel execution in Dymola 2017 FD01 which led to a simulation time of 82s.
Note: sparse solvers are available for `Cvode`.
It might be a good idea to spend some time checking whether `Cvode` should be our default solver? It is the one used by default in JModelica, and the one initially embedded in Dymola's generated FMUs for co-simulation.


mwetter:
@tsnouidui : Thanks, we will look further into CVode. With the Guideline 36 case study, we have two large models that would be good test cases, one with few events and the other with events every 2 minutes.

PR #1004, opened 2017-10-12
Issue967 vav reheat cdl merge

mwetter:


mwetter:
This is ready to merge when all tests pass

PR #1039, opened 2017-11-06
Fix syntax error

tbeu:


mwetter:
Thanks for the pull request.

PR #1044, opened 2017-11-08
TMix freeze protection as a separate block

milicag:
This PR:
* Pulls out the TMix based freeze protection block (non-G36, Michel introduced it for the lack of a good solution in G36) from economizer controllers to a separate model/adds a validation block
* Adds this freeze protection block to economizer controller models with the corresponding validation test (in ```Controller_Disable```)
* Sets the default freeze protection to ```TMix``` based model and 
* Eliminates ```uZonSta``` from the multizone economizer, since only relevant to single zone systems
* Introduces improvements to formatting and result plots, updates validation results


milicag:
@mwetter This is ready for review, the tests have passed (I have canceled the last test manually, since formatting change only).

mwetter:

Issue #1118, opened 2018-01-26
PartialParallelElectricEIR misses constrainedby declaration


mwetter:
The instance `chi` in `Buildings.Applications.DataCenters.ChillerCooled.Equipment.BaseClasses.PartialParallelElectricEIR` needs the `constrainedby` keyword. Otherwise, parameters are not assigned if the chiller is redeclared.

PR #1119, opened 2018-01-26
Added constrainedby for #1118

mwetter:
@YangyangFu : Can you please review this change. (This also needs to be corrected in Fig. 3 of the paper.)


YangyangFu:
@mwetter Thanks, Michael. I've updated Figure 3.

PR #1117, opened 2018-01-26
Add missing constants in package.order

sjoelund:
This fixes OpenModelica refusing the open the master branch

PR #1120, opened 2018-01-26
Add missing constants in package.order

mwetter:
This integrates PR https://github.com/lbl-srg/modelica-buildings/pull/1117


mwetter:
This closes #1121 
The tests currently fail due to a bug in BuildingsPy that is being fixed now.

Issue #1137, opened 2018-03-07
G36 library variable names consistency, include VAV in the name


milicag:
Go through the variable names in the G36 primary HVAC control library and check for consistency, e.g. TSupSet vs. TSetSup, etc.

Include a substring 'VAV' in the top AHU sequence controller name.

mwetter:
Please also write a conversion script with all the changes. See `./Resources/Scripts/Dymola/ConvertBuildings_from*.mos`

milicag:
I will. Sorry for the delay, but this is just about to make the top of my list.

milicag:
Include: Fix typo in info section of Buildings.Utilities.IO.Files.CSVWriter: ```The parameter samplePeriod defines every how may seconds the inputs are saved to the file. ```

mwetter:
@milicag This typo needs to be fixed by making a pull request in the IBPSA library https://github.com/ibpsa/modelica-ibpsa  (which will periodically be merged into the Buildings library).

milicag:
@mwetter It looks like I can't push to the IBPSA library: 

```remote: Permission to ibpsa/modelica-ibpsa.git denied to milicag.```

**Comment by mwetter** [comment]
you can make a fork and pull request.

**Comment by milicag** [comment]
Did you mean a branch?

```
mg@w:~/repos/modelica-ibpsa/IBPSA$ git push
Username for 'https://github.com': milicag
Password for 'https://milicag@github.com': 
remote: Permission to ibpsa/modelica-ibpsa.git denied to milicag.
fatal: unable to access 'https://github.com/ibpsa/modelica-ibpsa.git/': The requested URL returned error: 403
mg@w:~/repos/modelica-ibpsa/IBPSA$ git branch
* issue985_fix_typo_in_csvwriter
  master
mg@w:~/repos/modelica-ibpsa/IBPSA$
```

mwetter:
No, I did mean a fork.

milicag:
This is resolved and merged.

PR #1139, opened 2018-03-13
Issue1137 Rename variables and add a VAV subpackage

milicag:
This PR addresses #1137:

* Improves the nomenclature within the ```G36_PR1``` package.
* Adds a ```VAV``` suffix to ```SingleZone``` and ```MultiZone``` packages within the ```G36_PR1``` package. 

I'm temporarily committing [the rename script](https://github.com/lbl-srg/modelica-buildings/pull/1139/files#diff-24fe46d23384cb41d3f0d9c235b3618d). Please review and comment on the suggested renames, upon which I can run the script and commit the edits.  Here is my reasoning behind suggested modifications:

1) Even if the model name hints at what the output value refers to (e.g. ```ZoneTemperatures``` output did not contain a ```Zon``` substring), I'd still prefer to spell it out for the purpose of using the same variable name throughout the G36 package. For example ```ZoneTemperatures```: ```THeaSet``` -change to> ```TRooHeaSet```.

2) Any airflow from the terminal unit to the zone should have a ```Dis``` substring.

3) For single zone use ```Roo``` instead of ```Zon```.

4) Use  dT for any differential temperatures.

5) I was conservative to allow long labels in an attempt to minimize ambiguity.

6) Always place ```Set``` at the end of the name, unless the variable is the minimum or the maximum limit, in which case one would use, for example, ```SetMin```.

7) Damper position vs. damper position limit, just an fyi: exhaust damper control signal generating sequences do not have the damper limits and thus generate the damper signal, unlike the rest of the dampers, which generate the damper position, since the physical damper position limits are provided as parameters set at commissioning (as discussed with Brent).


milicag:
Note that the package renaming script (```move_class``` from ```buildingspy```) did not rename the ```OpenModelica.compareVars``` scripts.

milicag:
@mwetter The PR includes the class rename to add ```VAV```. I'd like some feedback on the variable rename script (link in the PR text) before I run it though.

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+     TCooWinOpe TZonCooSetWinOpe \
+     THeaWinOpe TZonHeaSetWinOpe \     
+
+    # in Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Reheat.SetPoints.ActiveAirFlow
+
+     VCooMax VDisCooSetMax_flow \
+     VMin VDisSetMin_flow \
+     VHeaMax VDisHeaSetMax_flow \
+     VMinCon VDisConMin_flow \
+     AFlo ZonAre \
```
mwetter: I would keep this `AFlo` (which is also used in the thermal zone model)

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+     outAirPerAre VOut_perAre_flow \
+     outAirPerPer VOut_perPer_flow \
+     VOccMinAir occVDisMin_flow \
```
mwetter: Start with `V`, e.g., `VOccDisMin_flow`

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+     VActMin VActMin_flow \
+     VActHeaMin VActHeaMin_flow \
+     VActHeaMax VActHeaMax_flow \
+
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Reheat.SystemRequests
+  
+     TCooSet TZonCooSet \
+     VDis VDis_flow \
+     VDisSet VDisSet_flow \
+     TDis \
```
mwetter: Something missing after `TDis`

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+     cooSetDif_1 dTZon_ZonCooSet_1 \ # these are very long
+     cooSetDif_2 dTZon_ZonCooSet_2 \ 
+     disAirSetDif_1 dTDis_DisSet_1 \
```
mwetter: Do we need two times `Dis` in `Dis_Dis`?

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+     TDisSet TDisHeaSet \
+ 
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.ModeAndSetPoints
+    # (most of these need to be consistent with G36_PR1.TerminalUnits.SetPoints.ZoneTemperatures)
+     THeaOn TZonHeaOn \
+     TCooOn TZonCooOn \
+     THeaOff TZonHeaOff \
+     TCooOff TZonCooOff \
+     freProThrVal TZon_frePro_on \
+     freProEndVal TZon_frePro_off \
```
mwetter: Use camel case such as `TZonFreProOn`

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+     TDis \
+     cooSetDif_1 dTZon_ZonCooSet_1 \ # these are very long
+     cooSetDif_2 dTZon_ZonCooSet_2 \ 
+     disAirSetDif_1 dTDis_DisSet_1 \
+     disAirSetDif_2 dTDis_DisSet_2 \ 
+
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Reheat.DamperValves
+
+     dTDisMax dTDis_ZonSet_Max \
+     V_flow_nominal VNom_flow \
```
mwetter: Keep `V_flow_nominal` as we have `m_flow_nominal` all over `Buildings.Fluid`

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+     decSetDem_1 decTSetDem_1 \
+     decSetDem_2 decTSetDem_2 \
+     decSetDem_3 decTSetDem_3 \
+
+# Buildings.Controls.OBC.ASHRAE.G36_PR1.Generic
+
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.Generic.SetPoints.OperationMode
+     THeaSet occTZonHeaSet \
+     TCooSet occTZonCooSet \
+     TZon TZon \
```
mwetter: TZon TZon are the same

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+# Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone
+
+    # SetPoints.ExhaustDamper
+     uFan uSupFan \ 
+
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.SetPoints.OutsideAirFlow
+     AFlo RooAre \ # already occured
+     outAirPerAre VOut_perAre_flow \ # already occured
+     outAirPerPer VOut_perPer_flow \ # already occured
+     TZon TRoo \
```
mwetter: I would always use TZon, for both single and multizone VAV.

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+     AFlo RooAre \ # already occured
+     outAirPerAre VOut_perAre_flow \ # already occured
+     outAirPerPer VOut_perPer_flow \ # already occured
+     TZon TRoo \ 
+     
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.SetPoints.VAVSupply
+     TMax TSupSetMax \
+     TMin TSupSetMin \
+     TSetZon TRooSet \
+     TZon TRoo \
```
mwetter: Duplicate from above

Buildings/issue1137_rename_scripts_temp.txt
```diff
@@ -0,0 +1,136 @@
...
+
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.SetPoints.VAVSupplySignals
+     TSetSup TSupSet \
+     
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.SetPoints.VAVSupplyFan
+     VBox_flow VDis_flow \
+     sumVBox_flow sumVDis_flow \  
+
+    # Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.SetPoints.OutsideAirFlow
+     AFlo ZonAre \ # already occured
```
mwetter: Keep `AFlo`

Buildings/Controls/OBC/ASHRAE/G36_PR1/AHUs/MultiZoneVAV/Controller.mo
```diff
@@ -1,4 +1,4 @@
-within Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone;
+within Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZoneVAV;
```
mwetter: I would prefer `AHUs.VAV.MultiZone`

milicag:
@mwetter I was just about to append to the ```ConvertBuildings_from_5.0_to_6.0.mos```, but the latest versions in the conversion scripts on the master are 4.0 to 5.0. Should I add a new file?

mwetter:
@milicag : Yes, please add a new file.

mwetter:
@milicag : As we already have a package `Buildings.Air.Systems.SingleZone.VAV` we should for consistency call the G36 sequences
`....G36_PR1.AHUs.MultiZone.VAV` and `....G36_PR1.AHUs.SingleZone.VAV`

milicag:
Just in time.

Buildings/issue1137_rename_scripts_temp.txt
milicag: It's simply a T difference between ```TDis``` and ```TDisSet```. Let's ignore that it has the indexes (_1, _2) for now. ```dTDisSet``` is shorter and appealing, but it indicates only one of the variables defining the dT. ```dTDis_Set``` may be a bit too cryptic. Let me know if you have a preference.

mwetter: If it is a control error, you could call it `errTDis` (= set point minus measured quantity)

milicag: That fits well, thanks. Do you have any preferences on ```dTDis_ZonSet_Max```? It's the maximum offset.

mwetter: I would need to know more context but `dTDisZonSetMax` or `errTDis(Max?)` may be options

milicag:
@mwetter I've commited the class rename script as ```Buildings\move_classes_temp.py```. Please let me know if the script runs fine on your environment.

mwetter:
@milicag : I corrected the conversion script and also work on a bug fix in BuildingsPy (https://github.com/lbl-srg/BuildingsPy/pull/194) that caused a new `package.mo` to not be generated. I will need to do some more testing.

milicag:
@mwetter Thank you for creating an "in depth" script. You also modified the conversion script the same way - does that mean that any conversion should be explicit down to the last package in the tree?

mwetter:
@milicag There is no need to list the packages down to the last package in the tree, but
```
convertClass("Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone",
                       "Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV")
```
seems ambiguous as it moves `MultiZone` to `MultiZone.VAV`, but then `MultiZone` still exists. Hence, I added one more package level below.

Please try again with the latest version of `BuildingsPy` from the master.

- [ ] Note that `TerminalUnits` should also be moved.

milicag:
@mwetter For debugging purposes, for some reason the script runs the Multizone correctly and throws an error for the single zone:
```
OSError: [Errno 2] No such file or directory: 'Buildings/Controls/OBC/ASHRAE/G36_PR1/AHUs/SingleZone/VAV/Economizers'
```
I think there is an issue with building intermediary package. It is also remarkable that the script performs correctly if there is only one downstream layer of packages, but if there are several it apparently fails.

mwetter:
@milicag : I get no error from the script if I apply it to what is on this branch.

milicag:
@mwetter Ok, I just pulled fresh BuildingsPy and am reapplying the script.

milicag:
@mwetter I get that same error. Package order also does not get written.

milicag:
@mwetter That said, I also get a an error when I run the tests:
```
Traceback (most recent call last):
  File "../bin/runUnitTests.py", line 181, in <module>
    show_gui = args.show_gui)
  File "../bin/runUnitTests.py", line 71, in _runUnitTests
    ut = u.Tester(tool=tool)
TypeError: __init__() got an unexpected keyword argument 'tool'
```
Which I have not been able to diagnose, since u.Tester clearly has a ```tool``` kwarg. There may be something odd happening with my setup. Let me attempt the conversion on the laptop.

milicag:
@mwetter I get the same error on the laptop.

mwetter:
@milicag : This error with the unit tests indicate that you have some corrupt installation of BuildingsPy. Try to reinstall it again and make sure you have no old version on your computer.

I also updated the master branch of BuildingsPy so that `package.order` are recreated.

milicag:
@mwetter I just realized it. I'll make sure to have the correct installation.

milicag:
@mwetter ```pip install buildingspy``` is not installing the latest master for some reason. The script and the tests work fine upon installing locally using ```pip install .```. I'll deal with the variables tomorrow, since it's a touchy activity.

mwetter:
@milicag `pip install buildingspy` will install the latest official release from https://pypi.python.org/pypi/buildingspy/1.6.0 and not the master.

milicag:
@mwetter Please note that if tests are run with ```runUnitTests.py``` from master and the latest official release of ```buildingspy```, they will fail due to the missing ```tool``` kwarg. The refactor method ```move_class``` as well. Current ```buildingspy``` master works as expected.

milicag:
@mwetter Would you be able to take a look at the [variable rename script](https://github.com/lbl-srg/modelica-buildings/pull/1139/commits/c531daf564ba3f60c0904ebe63ea398b49b26c6b) and see if I am omitting anything in the methodology. Even if I perform the renames package by package, some of the validation models still fail after the rename. For example, after running the current script (with only a subset of the variables we'd like to rename), the validation model that fails is ```G36_PR1/TerminalUnits/Validation/ModeAndSetPoints```. From what I've seen in the previous variable renaming issues, the script was simply run on the entire library.

milicag:
@mwetter It took me a while. It is not trivial to change a name of a variable, that used to bear the same name in 2 models, into two different names (for the sake of clarity). Please let me know if this is not something I should be attempting to do and I will drop such renames. Example:
```
variable: TCooSet
model: G36_PR1.TerminalUnits.SetPoints.ZoneTemperatures - into TZonCooSet
model: G36_PR1.Generic.SetPoints.OperationMode - into TOccZonCooSet
```
Where the problems occur is when both of these models are used in a, for example model C. In that model the variable name (in this case TCooSet) gets overwritten with say TZonCooSet (for OperationMode connectors) and remains unchanged once the actual OperationMode rename to TOccZonCooSet occurs.

I suspected that if I did:
```
convertElement("Buildings.Controls.OBC.ASHRAE.G36_PR1.Generic.SetPoints.OperationMode", "TZonCooSet ", "TOccZonCooSet")
```
the ```name_of_the_OperationMode_instance.TZonCooSet``` would change to ```name_of_the_OperationMode_instance.TOccZonCooSet```, but the command didn't do it.

How do you usually go about this issue?

mwetter:
I think it makes sense for clarity to use
`G36_PR1.Generic.SetPoints.OperationMode.TOccZonCooSet` (because this model also has `TUnoZonCooSet`)

The `convertElement` function from the `.mos` script should correctly apply the rename in the class as it looks up the class name. But the `.sh` script does not look up the class name, it simply applies the `sed` command, and hence requires manual verification and edits if rename strings are ambiguous.

milicag:
For some reason the convertElement script didn't do anything on my Dymola. Maybe I'm not running it properly. I'll run by Thierry to check for any obvious omissions. 

Regarding the bash script, it may be worth writting a more inteligent variable rename script at some point, that relies on the default component name in order to figure out which strings should and should not be replaced in connectors.

milicag:
@mwetter I'm running into the following issue. The ```convertElement``` command works as expected on a single element, but if I run [an entire convert element script](https://github.com/lbl-srg/modelica-buildings/pull/1139/files#diff-93dee29647d163195f225411f2f25cc8) not all elements get converted. 

The workflow for the variable renames is:
1. The classes have already been renamed in this PR version of the library
2. Run ```convertElement``` script to rename any variables in instantiated classes
3. Run [the sh script](https://github.com/lbl-srg/modelica-buildings/pull/1139/files#diff-9445e06e6649dcd0c729846e67e04a17) to rename the variables in models, classes, scripts and reference results

I've also updated the dymola conversion script. Note that it differs from the script run in step 2 since it contains the class conversions as well, and ```convertElement``` needs to be run on the old paths.

milicag:
There is a need for an another rename:
```.../Validation/Valve.mo``` into ```...Validation/VAVSupplySignals_Valve.mo```, since the previous validation model name did not contain the name of the model validated.

milicag:
I created a new branch for #1137 and intend create a new PR.

PR #1154, opened 2018-03-26
Issue1153 vav single zone y fan

### LINKED ISSUE #1153: SingleZone.SetPoints.VAVSupply need not have hysteresis
In this controller, the hysteresis can be removed as the fan control signal can be formulated as a signal that is continuous in `uCoo`.

mwetter:
This closes #1153


JayHuLBL:
@mwetter no further comment. I think it is better than previously implementation.

PR #1184, opened 2018-06-06
Issue1170 facade

### LINKED ISSUE #1170: Shade and blind control sequences
Implement simplified facade control sequences, based on the energy plus implementation ([EnergyPlus InputOutputReference, v8.3.0, WindowProperty:ShadingControl, pg 205](https://energyplus.net/sites/default/files/pdfs/pdfs_v8.3.0/InputOutputReference.pdf)).

Window shading elements such as blinds, shades are controlled using a schedule and threshold values, with the following enable variables: TOut, TZon, beam+diffuse irradiation.

milicag:
This closes #1170. 

* Added a shading control enable/disable block and a test.


milicag:
@mwetter Please take a look to see if this needs any edits before I put together a slide for the review. Tests are still running, local tests pass.

Buildings/Controls/OBC/Facade/ShadingEnable.mo
```diff
@@ -0,0 +1,121 @@
+within Buildings.Controls.OBC.Facade;
```
mwetter: Rename to `Shade`

Buildings/Controls/OBC/Facade/ShadingEnable.mo
```diff
@@ -0,0 +1,121 @@
...
+    "Set to true if enabling based on solar irradiance, otherwise enabling based on temeprature"
+    annotation(Dialog(group = "Conditional"));
+  parameter Modelica.SIunits.Temperature TThr = 298.15
+    "Temperature threshold (either zone or outdoor air)"
+    annotation(Evaluate=true, Dialog(group="Hysteresis", enable = not use_solIrr));
+  parameter Modelica.SIunits.Irradiance irrThr = 1000
+    "Solar irradiance threshold"
+    annotation(Evaluate=true, Dialog(group="Hysteresis", enable = use_solIrr));
+
+  CDL.Interfaces.BooleanInput uEnable "Shading device enable schedule"
```
mwetter: Please remove this input. In many cases, such as if a shade is only to be deployed when solar irradiation is high, this is not needed and just requires additional input for no good reason. Such overwrites should be done outside.

Buildings/Controls/OBC/Facade/ShadingEnable.mo
```diff
@@ -0,0 +1,121 @@
...
+    annotation (Placement(transformation(extent={{-120,-60},{-80,-20}}),
+    iconTransformation(extent={{-140,-80},{-100,-40}})));
+  CDL.Interfaces.RealInput T if not use_solIrr
```
mwetter: It would be easier if there are two blocks, `Shade_T` and `Shade_H`, one for temperature and one for irradiation.

Buildings/Controls/OBC/Facade/ShadingEnable.mo
```diff
@@ -0,0 +1,121 @@
...
+    "Zone or oudoor air temperature"
+    annotation (Placement(transformation(extent={{-120,20},{-80,60}}),
+    iconTransformation(extent={{-140,40},{-100,80}})));
+  CDL.Interfaces.RealInput irr if use_solIrr
```
mwetter: Please call the radiation input `H`, as is used in Buildings.HeatTransfer.Windows.BaseClasses.WindowRadiation.

You also need to set the unit of `H`

Buildings/Controls/OBC/Facade/ShadingEnable.mo
```diff
@@ -0,0 +1,121 @@
...
+    "Total solar irradiance on horizontal or on window surface"
+    annotation (Placement(transformation(extent={{-120,-20},{-80,20}}),
+    iconTransformation(extent={{-140,-20},{-100,20}})));
+
+  CDL.Interfaces.BooleanOutput yShaEna
```
mwetter: Because the room model, and also the Spawn interface, uses a continuous signal from 0 to 1, please change this to a Real output. Then it is compatible with Buildings.ThermalZones.Detailed.Examples.MixedAirFreeResponse

Buildings/Controls/OBC/Facade/ShadingEnable.mo
```diff
@@ -0,0 +1,121 @@
...
+    "Shade/Blind/Glaze/Screen status signal"
+    annotation (Placement(transformation(extent={{100,-10},{120,10}}),
+     iconTransformation(extent={{100,-20},{140,20}})));
+
+protected
+  parameter Modelica.SIunits.TemperatureDifference TDiff = 1
```
mwetter: Put these parameters for the hysteresis on the Advanced tab so that they can be adjusted.

Buildings/Controls/OBC/Facade/package.mo
```diff
@@ -0,0 +1,34 @@
+within Buildings.Controls.OBC;
+package Facade "Building facade control sequences"
+annotation (Icon(coordinateSystem(extent={{-100,-100},{100,100}}), graphics={
```
mwetter: Info section is missing

Buildings/Controls/OBC/Facade/ShadingEnable.mo
```diff
@@ -0,0 +1,121 @@
+within Buildings.Controls.OBC.Facade;
+block ShadingEnable "Generic shading device enable/disable seqence"
+
+  parameter Boolean use_solIrr = false
+    "Set to true if enabling based on solar irradiance, otherwise enabling based on temeprature"
+    annotation(Dialog(group = "Conditional"));
+  parameter Modelica.SIunits.Temperature TThr = 298.15
```
mwetter: `TSet` would be clearer than `TThr` as it is a set point.

milicag: @mwetter Setpoint is something a controlled device tracks, whereas in this case it seems to be a pure activation/deactivation threshold, especially if it is the outdoor air temperature. E+ suggests there exist setpoint schedules, which are then followed by the angle of the blind.

mwetter: I think it is still fine to call it set point. If it is the room temperature, then there could be situations where opening and closing the shade will keep the room temperature around this temperature (but typically, the control gain is too small)

milicag:
@mwetter Supersimple version ready for review.

PR #1194, opened 2018-06-18
Fix obsoleted icon

beutlich:


mwetter:
@beutlich Thanks for the pull request. I also made the changes on the upstream IBPSA library.

PR #1213, opened 2018-07-18
Update CDL.Logical.Timer block to include accumulate true input time

### LINKED ISSUE #1212: add timer reset parameter of block CDL.Logical.Timer
Add boolean parameter `reset` to allow:
- `reset = true`: reset timer to zero when input becomes false
- `reset = false`: accumulate time of true input

JayHuLBL:
This closes #1212.


mwetter:
@JayHuLBL Please also update the revision notes.

PR #1217, opened 2018-07-20
Issue1216 extractor

JayHuLBL:
- [x] changed name of block `CDL.Routing.ExtractSignal` to `CDL.Routing.RealExtractSignal`
- [x] created new block '`CDL.Routing.RealExtractor`
- [x] updated the release note


mwetter:
@JayHuLBL : I will add an entry to the conversion script, which we always need to do for backward incompatible changes. The change in class name is backward incompatible as older models need to be updated.

mwetter:

PR #1230, opened 2018-09-07
corrected labeling error in the script and in models

milicag:


milicag:
@mwetter This should fix the issue with the comments. Please let me know if the conversion works as expected in your example.

PR #1277, opened 2018-11-26
Test and merge changes for segfault

mwetter:


mwetter:
@kbenne : Could this be closed? There seem to be many changes that are already on the target branch. Best would be to make PR for CMake to branch `issue1129_energyPlus_zone` and I will then delete `issue1129_energyPlus_zone_dll` as this is merged into `issue1129_energyPlus_zone`.

PR #1284, opened 2018-12-08
corrected comment inconsistency and added 2 test cases

### LINKED ISSUE #1282: Fix imprecise comments in the CDL hysteresis block
The comments that require fixing:

parameter Real uLow(start=0) "if y=true and u<=uLow, switch to y=false";
parameter Real uHigh(start=1) "if y=false and u>=uHigh, switch to y=true";

Suggested correction (please note the change to the inequality sign):

parameter Real uLow(start=0) "if y=true and u<uLow, switch to y=false";
parameter Real uHigh(start=1) "if y=false and u>uHigh, switch to y=true";

This change would be in line with the existing info section and the code implementation (y = not pre(y) and u > uHigh or pre(y) and u >= uLow):

When the output was false and the input becomes greater than parameter uHigh, the output switches to true.
When the output was true and the input becomes less than parameter uLow, the output switches to false.

milicag:
​Aligned comments with the code implementation and the info section.
Added two cases to tests for inputs that fall on chosen hysteresis bounds and slightly out of them.

Closes #1282​


mwetter:

PR #1301, opened 2018-12-18
modified the output variable of MultiMax block from yMax to y

kuzha:
The two models using this block have also been modified accordingly.


kuzha:
@mwetter Please have a review at the changes.

PR #1308, opened 2018-12-21
Issue1252 multi max

### LINKED ISSUE #1252: MultiMax change output from yMax to y
Change the name of the output connector for `CDL.Continuous.MultiMax` from `yMax` to `y`. This will allow replacing the block and keeping the connection.
Also, check `MultiMin` and do the same change if needed.

mwetter:
This closes #1252


Buildings/Controls/OBC/CDL/Continuous/MultiSum.mo
```diff
@@ -5,10 +5,11 @@ block MultiSum
   parameter Integer nin(min=0) = 0 "Number of input connections"
     annotation (Dialog(connectorSizing=true), HideResult=true);
   parameter Real k[nin]=fill(1, nin) "Input gains";
-  Interfaces.RealInput u[nin] "Connector of Real input signals"
-    annotation (Placement(transformation(extent={{-140,70},{-100,-70}})));
   Interfaces.RealOutput y "Connector of Real output signal"
     annotation (Placement(transformation(extent={{100,17},{134,-17}})));
+  Interfaces.RealInput u1 "Connector of Real input signal 1"
```
mwetter: @kuzha : This should be u[nin] "Connector of Real input signals" and listed before the output.

mwetter:
@kuzha : This still fails, see https://travis-ci.org/lbl-srg/modelica-buildings/jobs/471093516
Looks like you use the wrong variable in the mos script.

You can run regression tests of individual packages using a command such as
```
../bin/runUnitTests.py --single-package Buildings.Controls
```

mwetter:

PR #1355, opened 2019-02-08
Issue829 sun rise set

### LINKED ISSUE #829: Add block for sunrise and sunset to CDL
Add a block that outputs the next sunrise and sunset time. The block should be updated for the sunset at each sunset, and for the sunrise at each sunrise. This will give the building automation system sufficient time to schedule controls in the corner-case of the sun only setting for a few seconds (e.g., at or above the polar circle).

The outputs should be the next sunrise and sunset in units of the model time.

See the `BoundaryConditions` package for a block that could be used.

mwetter:
This closes #829.

This can be merged if https://github.com/lbl-srg/modelica-buildings/issues/829#issuecomment-461692799 is addressed and all tests pass.


mwetter:

Issue #1314, opened 2018-12-21
unused connectors in Buildings.Applications.DataCenters.ChillerCooled.Equipment


mwetter:
Quite a few models in `Buildings.Applications.DataCenters.ChillerCooled.Equipment` have inputs `trigger` and `y_reset_in` that are not connected to anything. It looks like they should have been removed in the base class.
@YangyangFu : can you please review these and make a pull request. Thanks.

Issue #1312, opened 2018-12-21
Set parameters final in PartialPlantParallel


mwetter:
@YangyangFu :

In `Buildings.Applications.DataCenters.ChillerCooled.Equipment.BaseClasses.PartialPlantParallel`, the assignment
```modelica
  extends Buildings.Applications.DataCenters.ChillerCooled.Equipment.BaseClasses.ValvesParameters(
    numVal = 2,
    m_flow_nominal = {m1_flow_nominal,m2_flow_nominal},
```
should be changed to
```modelica
  extends Buildings.Applications.DataCenters.ChillerCooled.Equipment.BaseClasses.ValvesParameters(
    final numVal = 2,
    final m_flow_nominal = {m1_flow_nominal,m2_flow_nominal},
```
because for example `l` is declared to have dimension `2`.

mwetter:
@YangyangFu : There is quite some confusion about "identical chillers". For example, 
`Buildings.Applications.DataCenters.ChillerCooled.Equipment.Validation.ElectricChillerParallel`
declares
```modelica
  parameter Buildings.Fluid.Chillers.Data.ElectricEIR.ElectricEIRChiller_McQuay_WSC_471kW_5_89COP_Vanes
    per1 "Chiller performance data"
    annotation (Placement(transformation(extent={{60,80},{80,100}})));
  parameter Buildings.Fluid.Chillers.Data.ElectricEIR.ElectricEIRChiller_York_YT_563kW_10_61COP_Vanes
    per2 "Chiller performance data"
    annotation (Placement(transformation(extent={{32,80},{52,100}})));
  Buildings.Applications.DataCenters.ChillerCooled.Equipment.ElectricChillerParallel chiPar(
    num=2,
    redeclare package Medium1 = Medium1,
    redeclare package Medium2 = Medium2,
    m1_flow_nominal=mEva_flow_nominal,
    m2_flow_nominal=mCon_flow_nominal,
    dp1_nominal=6000,
    dp2_nominal=6000,
    dpValve_nominal={6000,6000},
    per={per1,per2})
    "Identical chillers"
    annotation (Placement(transformation(extent={{-10,0},{10,20}})));
```
Hence, `chiPar` says they are identical, but then different performance records are used.
Can you please go through the package and sort out these inconsistencies.

PR #1356, opened 2019-02-08
Issue1311 1312 1314 data center

### LINKED ISSUE #1311: ElectricChillerParallel says chillers are identical, but parameters allows different records
@YangyangFu : 
The model `Buildings.Applications.DataCenters.ChillerCooled.Equipment.ElectricChillerParallel` states that the chillers are identical.
However, the data record `per[]` is an array, and also the valve data are an array. If the chillers really are identical, which I assume they are since there is only one value for `m1_flow_nominal` and `m2_flow_nominal`, the parameter `per` should be a scalar. Otherwise the user may configure them to be different chillers

mwetter:
This closes #1311, #1312, #1314


mwetter:

PR #1359, opened 2019-02-09
issue829 sunrise/sunset 2 new examples

kuzha:
@mwetter:
I added two new examples in the validation package: one starts from -12h; the other from 12h. The results look fine.

The two new examples are the same as the existing example except the start time and end time.


mwetter:
@kuzha : Please give the tests a better name than ...1 and ...2, such as ...NegativeStartTime and ...PositiveStartTime.

Buildings/Controls/OBC/CDL/Utilities/Validation/SunRiseSetNegativeStartTime.mo
```diff
@@ -27,15 +27,15 @@ equation
 annotation (
   experiment(StartTime=-43200, StopTime=259200, Tolerance=1e-06),
   __Dymola_Commands(
-  file="modelica://Buildings/Resources/Scripts/Dymola/Controls/OBC/CDL/Utilities/Validation/SunRiseSet1.mos"
+  file="modelica://Buildings/Resources/Scripts/Dymola/Controls/OBC/CDL/Utilities/Validation/SunRiseSetNegativeStartTime.mos"
         "Simulate and plot", executeCall=simulateModel(
```
mwetter: Please remove the `executeCall` (we don't use that anywhere else).

PR #1365, opened 2019-02-15
New package OutdoorLights

kuzha:
@mwetter:

The current implementation is: outdoor lights is controlled only by the sunUp information. I am not sure if other features need to be added.
The testing is also simple since there is no input for the block, with only three parameters like the SunRiseSet block.

Please review and comment.


Buildings/Controls/OBC/OutdoorLights/DaylightControlled.mo
```diff
@@ -0,0 +1,99 @@
...
+
+protected
+  CDL.Utilities.SunRiseSet sunRiseSet(
+  final lat=lat,
+  final lon=lon,
+  final timZon=timZon)
+   annotation (Placement(transformation(extent={{-20,-10},{0,10}})));
+
+equation
+  y = if sunRiseSet.sunUp then 0 else 1;
```
mwetter: While this is valid Modelica, it is not valid in OpenBuildingControl. See http://obc.lbl.gov/specification/cdl.html#equations

The CDL translator modelica-json won't translate such assignments.

kuzha: CDL blocks are used for the implementation. Please review.

Buildings/Controls/OBC/OutdoorLights/DaylightControlled.mo
```diff
@@ -12,10 +12,19 @@ protected
   final lat=lat,
   final lon=lon,
   final timZon=timZon)
-   annotation (Placement(transformation(extent={{-20,-10},{0,10}})));
+   annotation (Placement(transformation(extent={{-80,-10},{-60,10}})));
+  CDL.Logical.Not not1
```
mwetter: Every instance needs a short comment. Please update the file accordingly.

kuzha: Sorry I forgot again...Added already.

PR #1403, opened 2019-04-05
Corrected Latch and Toggle implementation

### LINKED ISSUE #1402: Correct implementation of CDL.Logical.Latch
Initially, if input `u0` is `false`, 

- if input `u1` is `true`, then output `y` should be `true`.
- if input `u1` is `false`, then output `y` should be `false`.

current implementation gives wrong output as below:
image

This issue will correct the implementation.

JayHuLBL:
This closes #1402.

- [x] corrected `CDL.Logical.Latch` and `CDL.Logical.Toggle` implementation that would cause wrong onput at initial stage.
- [x] checked classes that instantiated the `Latch` block, i.e. `ASHRAE.G36_PR1.Generic.SetPoints.OperationModel` and `ASHRAE.G36_PR1.TerminalUnits.Reheat.SystemRequests`
- [x] corrected validation model `ASHRAE.G36_PR1.Generic.SetPoints.Validation.OperationModel`
- [x] updated release note


mwetter:
@JayHuLBL Please address these points:
- [ ] In `Logical.Toggle`, the info section uses on and off. This should be `true` and `false`
- [ ] In `Logical.Toggle`, the info section uses 'clr', this should be `u0` (Maybe we should change the name of this input to `reset` in the `toggle` and `
- [ ] In `Logical.Toggle` please rewrite the info section. It currently is not clear. For example, it says "The block toggles output value whenever input turns ON." but this is not the case at t=3.75 in the figure.
- [ ] In `Logical.Latch` I believe the last sentence in "When the latch input u switches to true, the output y becomes true and remains true, even if u turns false, until the clear input u0 becomes true. When the clear input u0 becomes true, then the output y becomes false. " can be deleted as it is the same as "...until the clear input u0 becomes true".
- [ ] Are these blocks also in the Modelica Standard Library? The revision notes state "...based on the Modelica Standard Library" but I can't find them. If they are in the MSL, do they still have the same behavior as the ones in the MSL?
- [ ] Move the parameters that have in this revision in `Toggle` and `Latch` be made protected to the public section. Otherwise, this is not backward compatible. If you have a good reason why they should be protected, then this need to be added to the backward incompatible change list, and the conversion script needs to be updated.

Also, wouldn't it be clearer to rename `u0` to `clear`? This could be done in a separate issue.

JayHuLBL:
@mwetter It's ready for review when all tests passed.

mwetter:
- [ ] Reformat table in `Latch` to separate initial behavior from simulation behavior, similar to `Toggle`.
- [ ] In `Toogle`, the `pre(y)` and `y` columns are wrong (after I attempted to reformat the table without changing its meaning).
- [ ] Make sure the validation has a case for each scenario.

JayHuLBL:

mwetter:
@JayHuLBL 
I updated the conversion script, and started revising the documentation. Let's discuss as it is still not entirely clear. Also, please do the following:
- [ ] Add the new block `swiCleaInp1` in `Buildings.Controls.OBC.CDL.Logical.Validation.Latch` to the regression tests. This block tests a transition that is not tested in `swiCleInp`.

JayHuLBL:
@mwetter 
- [x] added more tests to the validation models to cover all scenarios
- [x] validated the `latch` and `toggle` implementation by comparing their outputs with the outputs from Eikon logic simulation. However, since EikonLogic simulation cannot export results, the validation was done by observing the behavior.
- [x] updated the info section and changed the example plots

PR #1407, opened 2019-04-10
Issue612 ffd wtian

xuhan425:
The following changes are made:
1. allocate memory for cosim->ffd->msg in ffd codes instead of cfdStartCosimulation.c
2. free memory for all pointers with allocated memory when the simulation finishes 
3. use "double " instead of "REAL" as a type conversion
4. use "Sleep()" consistently
5. remove "singleInstance = true" to avoid warnings


Buildings/Resources/C-Sources/cfdExchangeData.c
```diff
@@ -33,7 +33,7 @@ int cfdExchangeData(double t0, double dt, double *u, size_t nU, size_t nY,
 
 	/*check if current modelica time equals to last time*/
 	/*if yes, it means cfdExchangeData() was called multiple times at one synchronization point, then directly return*/
-	if(abs(cosim->modelica->lt - (REAL) t0) < 1E-6){
+	if(abs(cosim->modelica->lt - (double) t0) < 1E-6){
```
mwetter: @xuhan425 : Is there a reason why you have all these type conversions `(double)`? The variable `t0` (as the others) are already declared as double, so there is no reason I see why we need to cast them again to double.

xuhan425: @mwetter Yes, all these can be deleted. 
Our original purpose of the type conversions is to ensure the precision of all variables from modelica is consistent with that in ffd solver. For example, if we use double in ffd (and if modelica does not support double), we may `#define REAL double` and convert precision of variables from modelica to double by using `(REAL)`. 
Since modelica does support double, no need to cast them again. Will update the codes.

xuhan425: @mwetter Revised.

Issue #1188, opened 2018-06-13
add loop breaker


mwetter:
This issue is to test a component that avoids nonlinear system of equations by making the mass flow rate a state variable. This is motivated by the approach in https://github.com/ibpsa/modelica-ibpsa/issues/946 but allows
- to place the state in selected parts of the system, and 
- retaining the usual fluid modeling approach.

PR #1430, opened 2019-05-08
Issue1298 linear actuators

### LINKED ISSUE #1298: Implement linearization algorithm for damper and valve models
Difficulties are encountered by users when it comes to properly sizing dampers and valves. This results in poor authorities, unstable control loops and eventually in various strategies to deal with the non linearity of the controlled system e.g. in [this post](https://groups.google.com/forum/#!topic/modelica-buildings/L5rBkXLWubY).
Another typical case might be the control of the secondary fluid outlet temperature of an heat exchanger by actuating the valve position on the primary side.

The development aims at providing a built-in option in the actuator models to linearize the controlled system in its whole. It only addresses static non linearities (deviation of the whole system characteristic from linear trend) as opposed to dynamic non linearities (related to actuators' dynamics).

AntoineGautier:
This partly closes #1298 (the valves components still need to be updated with the same logic).
(Corresponding issue in IBPSA library: [#1188](https://github.com/ibpsa/modelica-ibpsa/pull/1188))

- `VAVBoxExponential` is retired. The option for the fixed resistance has been integrated into `Exponential`. (A conversion script is attached.)
- The exposed parameter `dp_nominal` in `Exponential` and `PressureIndependent` has been replaced by `dpExp_nominal` and `dpDam_nominal` (modification also automated by the above script). The motivation is to restrict the use of `dp_nominal` to the nominal pressure drop _at the model boundaries_ (which is used by `PartialDamperExponential` in the homotopy initialization cf. `dp/dp_nominal_pos`). In the master branch, it is indeed the case for `Exponential`, it might be the case for `VAVBoxExponential` if `dp_nominalIncludesDamper==True` and it is never the case for `PressureIndependent`. Introducing the new parameters ensures consistency (and appears to be the best way to merge `VAVBoxExponential` into `Exponential` properly).
- The option to linearize the characteristics of the damper plus fixed resistance has been added.
- New limiting values for parameters `k0` and `k1` are used in `PartialDamperExponential` motivated by [https://antoinegautier.github.io/plots/Actuators.html#/6](https://antoinegautier.github.io/plots/Actuators.html#/6).
- `PressureIndependent`: 
    * the algorithm implemented to deal with small derivatives of `m_flow` / `dp` while ensuring a bounded error (approximately) is illustrated here [https://antoinegautier.github.io/plots/Actuators.html#/4](https://antoinegautier.github.io/plots/Actuators.html#/4). It relies on  1) an intermediary `dp_lim` variable limiting the linear domain where `m_flow = y_actual * m_flow_nominal + cregul *...`, 2) an extended domain for the upper spline interpolation and  3) a leakage flow supposed to be mainly turbulent (to limit the domain where the 2 flow functions are not inverse from each other). The resulting flow characteristics is also part of the updated example see `Buildings/Fluid/Actuators/Dampers/Examples/Damper.mo` case `preIndCha`. The impact on the test case described in #1298 is presented in [https://antoinegautier.github.io/plots/Actuators.html#/2/1](https://antoinegautier.github.io/plots/Actuators.html#/2/1).
    * the steep transition of the opening from 1 to 0 when `y=0` and `dp` transitions from negative to positive values is handled through a threshold value (2%) of the input signal below which the opening is forced to zero.

Note:

- Some models of any library using `PressureIndependent` might fail to simulate without changing some parameters values to be consistent with the exponential damper characteristics that is now used by the model to compute the opening (typically models in `Buildings/Examples/VAVReheat` had a too high nominal pressure drop with a too low nominal velocity: those parameters have been changed manually, the script mentioned above does not handle that modification).
- The reference results have been updated for the 3 models hereafter which use `Exponential` or `MixingBox` damper models. This is unexpected but needed due to small numerical effects yielded by the modification of the model structure only (as one can see in the modified translation statistics): the governing equations are unchanged and it has been verified that all the parameters (including the protected ones) have the same value between the master and the development branch (except for `dp_nominal` as explained before but that should not impact the simulation after initialization).
    * `Buildings_Examples_DualFanDualDuct_ClosedLoop.txt`
    * `Buildings_Examples_HydronicHeating_TwoRoomsWithStorage.txt`
    * `Buildings_Applications_DataCenters_DXCooled_Examples_DXCooledAirsideEconomizer.txt`
- `Buildings.Examples.DualFanDualDuct.ClosedLoop` fails to simulate with JModelica for the same reason. I will notify Modelon (Agnes). In the meantime simulation is excluded from JModelica regression tests.

Conversion script:

[VAVBoxExponentialRefactor.py.txt](https://github.com/lbl-srg/modelica-buildings/files/3437794/VAVBoxExponentialRefactor.py.txt)


AntoineGautier:
@kuzha Thanks for proposing implementing the required updates on `Buildings/Examples/VAVReheat/ASHRAE2006.mo` and `Buildings/Examples/VAVReheat/Guideline36.mo`. The last commit already integrates the correct parameters values in `Buildings.Examples.VAVReheat.ThermalZones.VAVBranch`.
What still needs to be done:

- [ ] Correct the control blocks `Buildings/Examples/VAVReheat/Controls/RoomVAV.mo` and `Buildings/Examples/VAVReheat/Controls/DuctStaticPressureSetpoint.mo` as needed.
- [ ] Connect `y_actual` to the control signal input for `Buildings/Examples/VAVReheat/Controls/DuctStaticPressureSetpoint.mo` and `Buildings/Controls/OBC/ASHRAE/G36_PR1/TerminalUnits/Reheat/SystemRequests.mo`.

AntoineGautier:
Close as it is now addressed in [issue1188_dampersRefactor](https://github.com/ibpsa/modelica-ibpsa/tree/issue1188_dampersRefactor) of IBPSA.

PR #1438, opened 2019-05-23
Issue1434 matrix min max

### LINKED ISSUE #1434: Matrix min and max
Implement CDL basic blocks to output:
- [ ] row/column wise max vector of the input 2d matrix
- [ ] row/column wise min vector of the input 2d matrix

milicag:
Closes #1434 

If anyone can suggest a way to autosize a matrix sized input, please let me know. `connectorSizing` seems to work only with a scalar input size (input is a vector).


mwetter:
I don't think you can autosize a matrix input connector. You will need to propagate its dimension.

milicag:
I wonder why it is not possible to use two parameters with `connectorSizing` (simply one for each dimension).

mwetter:
@milicag :
I revised the model. The main change, other than renaming parameters, is that I changed the min attribute from `min=0` to `final min=1` as I don't think you want a matrix with 0 column (or rows).

Can you please address the following:
- [ ] The two validation cases are not good. I think you used the wrong sign on the input as the result is zero for all outputs. Please flip the sign of the input.
- [ ] Update the release notes in the top-level `package.mo` file as we want to inform users about changes in the library.

milicag:
@mwetter Thanks for the revisions. I did all zero outputs on purpose - it is the expected result given the inputs. What did you have in mind?

mwetter:
@milicag : If you were to switch the sign on all inputs, then the example would also test that the right minimum value (or maximum value) is in the correct element of the output signal. This is currently not tested.

milicag:
@mwetter I'm pretty sure the test covered the position as well, but I can revise it so it does not confuse the users.

milicag:
@mwetter Could you check out the revised tests and let me know if those look any better.

PR #1457, opened 2019-06-21
Discrete moving mean

kuzha:
Add the discrete moving mean block to the CDL package.


Buildings/Controls/OBC/CDL/Discrete/DiscreteMovingMean.mo
```diff
@@ -0,0 +1,104 @@
...
+        Line(points={{42,22},{62,22},{62,-22}}, color={28,108,200})}),
+Documentation(info="<html>
+<p>
+This block calculates the discrete moving mean values of an input signal.
+The continuous input signal is first sampled; then at each sampling rate, the 
+block outputs the average value of the past n samples including the current 
+sample. 
+</p>
+<p>
+During initialization, the initial value given by the user is the value for the
```
mwetter: @kuzha : This should be changed. At the first sample, it should produce the input. At the next sample, it should produce the average of the past two samples, then the past three samples and so on up to `n` samples. Then, the block is indeed a discretized version of `CDL.Continuous.MovingMean`.

Buildings/Controls/OBC/CDL/Discrete/DiscreteMovingMean.mo
```diff
@@ -0,0 +1,104 @@
+within Buildings.Controls.OBC.CDL.Discrete;
+block DiscreteMovingMean "Discrete moving mean of a sampled input signal"
+  parameter Integer nSam(min=2) "Number of samples to be averaged";
```
mwetter: I prefer to simply use `n` (which is shorter and customary for the size of samples).

Buildings/Controls/OBC/CDL/Discrete/DiscreteMovingMean.mo
```diff
@@ -0,0 +1,104 @@
...
+  parameter Modelica.SIunits.Time samplePeriod(min=1E-3)
+      "Sampling period of component";
+  parameter Real y_start = 0
+      "Initial value for the first nSam-1 samples";
+  parameter Modelica.SIunits.Time startTime = 0
```
mwetter: This should be removed. See `FirstOrderHold` for how we deal with the simulation start time.

Buildings/Controls/OBC/CDL/Discrete/DiscreteMovingMean.mo
```diff
@@ -0,0 +1,104 @@
...
+      "Simulation start time, can be negative";
+
+  Interfaces.RealInput u "Continuous input signal"
+    annotation (Placement(transformation(extent={{-140,-20},{-100,20}})));
+  Interfaces.RealOutput y "Discrete averaged signal"
+    annotation (Placement(transformation(extent={{100,-10},{120,10}})));
+  Sampler sampler(samplePeriod=samplePeriod) "Sampled input signal as a reference"
```
mwetter: This should be `final samplePeriod=samplePeriod`

kuzha:
@mwetter 

- I changed the name to MovingMean;
- The current implementation is based on equations;
- The initial n-1 samples is calculated as you indicated previously;
- I was trying to reduce the many time events but has not succeeded.

Issue #1429, opened 2019-05-08
update old annotations


mwetter:
Some models use old annotations that need to be corrected. See for example below.
```
Buildings/Examples/VAVCO2/VAVSystemCTControl.mo
...
       Text(
        extent=[-140,234; -96,192],
        string="PAtm"),
```
The correct syntax is
```
Text(
          extent={{-44,48},{-26,36}},
          textString="aaa")
```
This can be detected when exporting the html documentation from dymola.

If the models are also in the IBPSA library than they should be corrected there too and can be merged with a script such as below.
```python
import buildingspy.development.merger as m
import os
import shutil
from git import Repo
from os.path import expanduser
home = expanduser("~")
ibpsa_dir = os.path.join(home, "proj/ldrd/bie/modeling/github/ibpsa/modelica-ibpsa/IBPSA")
dest_dir = os.path.join(home, "proj/ldrd/bie/modeling/github/lbl-srg/modelica-buildings/Buildings")
mer = m.IBPSA(ibpsa_dir, dest_dir)
mer.merge()

os.remove(os.path.join(dest_dir, "Resources", "Images", "IBPSA-logo-text.png"))
os.remove(os.path.join(dest_dir, "Resources", "Images", "IBPSA-logo.png"))
```

mwetter:
@LisaRivalin : Can you please have a look at this issue and make a pull request. The problem of old annotations keeps coming up occasionally in Dymola warnings.

PR #1491, opened 2019-07-15
updated annotations (for #1429)

LisaRivalin:
Some models used old annotations causing graphical discrepancies (#1429 ). This has been addressed as follow:

- Changed "extent" coordinate convention from extent=[x1,x2;y1,y2] to extent={{x1,x2},{y1,y2}}
- When inside "annotation", changed string to textString


LisaRivalin:
Able to reproduce : 
Screen Shot 2019-07-17 at 9 44 33 AM

with Dymola FD2019 and Buildings Library (Version: 6.0.0, 2019-07-15, Uses: Modelica (version="3.2.2") )

mwetter:

PR #1492, opened 2019-07-16
Fix wording

tbeu:
I noticed when reading the release notes from https://github.com/lbl-srg/modelica-buildings/releases/tag/v6.0.0


mwetter:
@tbeu Thanks for catching and fixing this typo.

PR #1498, opened 2019-07-17
Issue1375 connector names size

### LINKED ISSUE #1375: add connector name to inputs and outputs, and make size consistent
The connectors in `CDL.Interfaces` do not show the name of the connection. This issue is to add the connector name.
Moreover, the size is different between the icon and diagram layer (which is also the case for the connectors from the MSL). This issue will also make the connector the same size so that they don't have to be resized if there are many inputs or outputs to a component.

JayHuLBL:
This closes #1375.

- [x] resized the connector


JayHuLBL:
@mwetter 
The connector sizes in icon and diagram layer (`Controls.OBC.CDL.Interfaces`), input and output, are same now:
- [x] classes in `Controls.OBC.CDL`. 
- [x] classes in `Controls.OBC.ASHRAE.G36_PR1`. 

When instantiating input and output connectors, they will have same size. 

However, the placement of connector would still be different between icon and diagram layer. To make sure same placement between layers, one needs to go to (in Dymola) `Edit -> Options... -> Graphical Editor` and enable `Apply movement of connectors to both layers`.

JayHuLBL:
@mwetter CI tests have passed and it's ready to merge.

PR #1503, opened 2019-07-22
Issue1265 sz controller

kuzha:
@dhblum 
The updates are what we have discussed today; you can start from this update


dhblum:
Thanks @kuzha.  One question I have is what is the reason for https://github.com/lbl-srg/modelica-buildings/pull/1503/commits/bee55b58a24ddbc48e58c2c402dfdf0d5404bfc3?

kuzha:
@dhblum 
If you check the block Economizers.Subsequences.Modulation, you will see uMin is set to be equal 0.1. This parameter should be propagated to the Economizers.Controller; otherwise the default value 0 will cause issues for the Modulation.HeaCoi (which is the case for the validation model Economizers.Validation.Controller_Mod_DamLim)

dhblum:
Ah ok, I see.  Thanks!

PR #1510, opened 2019-07-31
Issue1265 sz controller

kuzha:
@dhblum:
I have updated the vav controller, economizer controller and the block Enable to allow fixed + differential dry bulb temperature cutoff as well as associated validation models.

Validation example of VAV controller will be added soon; in the meantime, you can use the updated blocks for your system model


dhblum:
Great, thanks @kuzha !

Issue #1265, opened 2018-10-24
Single Zone VAV Controller


dhblum:
This issue is to implement a controller for a single zone VAV system, similar to `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller`.

PR #1513, opened 2019-08-03
Issue1265 sz controller

dhblum:
Implementation of Single Zone VAV Guideline 36 controller for https://github.com/lbl-srg/modelica-buildings/issues/1265.  FYI @kuzha for me creating this PR.

Controller can be found at ``Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.VAV.Controller``
Example system implementation can be found at ``Buildings.Air.Systems.SingleZone.VAV.Examples.Guideline36`` 

The following still needs to be done:

- [x] Add reference unit test results for new unit tests
- [x] Update validation tests for controller
- [x] Double check all documentation
- [x] Sizing of cooling/heating coils for G36 example needs to be bigger than existing example to meet zone setpoint.
- [x] Looks like there are some conflicts with the master to resolve
- [x] Review by @mwetter, @milicag , and others as directed (@AntoineGautier @LisaRivalin @JayHuLBL).

One issue to resolve also is moving forward with if/how to report an issue with Figure 5.18.5.2 in the guideline.  It seems the Figure as published in the guideline assumes the SATsp control signal to go from -1 (max heat - in a normalized case) to 0 (no heat) from left to right on the x-axis.  However, conventional control notation (e.g. in [Control Systems Engineering](https://www.amazon.com/Control-Systems-Engineering-Norman-Nise/dp/0471794759), 5th Edition by Nise) has the error signal of a controller as uSet - uMea.  Therefore, the control error signaling more heat would actually go from 0 (no heat) to +1 (max heat - in a normalized case).  This convention would require the Figure graphics to be flipped along a center line on the x-axis, so that as SATsp control signal gets larger as indicated to the right, more heat (closing OA damper, opening RA damper, and turning on heating coil) is used.


mwetter:
@dhblum : I think it is better to consistently use `uSet - uMea` for the error and hence flip the axis. It would also help to indicate the range (e.g., 0 to 1) for the x-axis.

dhblum:
@kuzha Added a task about looking into the sizing of the heating/cooling coil for the conventional and G36 system examples.

dhblum:
@mwetter This is ready for you to review.

FYI @kuzha.

mwetter:
@dhblum : I reviewed all files and made minor changes to formatting and documentation. Can you and @kuzha please address the following:
- [ ] In `EconEnableDisableStateMachineChart.png`, don't use notation such as `TOutCut / TOut`. Better use `TOutCut (or TOut)` to avoid using a division sign in this expression.
- [ ] In `Resources/Images/Controls/OBC/ASHRAE/G36_PR1/AHUs/SingleZone` not all png files have an svg file. Please add these to the repository.
- [ ] It would be easier to use the controllers if signal input names and output names were added to the icon layer. Can you please do so, and also add some graphics where possible so one can recognize the controllers easier.
- [ ] The block `SingleZone.VAV.Controller` uses the instance `zonSta`. This block has inequality on real signals, and these signals are the output of the integrators of the PI controller.  This can cause chattering. Please add hysteresis to these tests. Moreover, because of the dynamics of the two PI controllers, one cannot rule out that `uHea > 0` _and_ `uCoo > 0`, in which case the output of this block would be wrong. Better would be to use a cascading implementation that first checks if we are in heating mode, and if not, it checks if we are in cooling mode..
- [ ] Update the release notes in `Buildings/package.mo` to describe the new added functionality.

dhblum:
@kuzha The latest merge is passing.  Can you just check you've addressed everything on @mwetter to-do list?

dhblum:
@kuzha has confirmed the to-do list is complete.

Buildings/package.mo
```diff
@@ -196,6 +196,33 @@ its class name ends with the string <code>Beta</code>.
     <b style=\"color:blue\">non-backward compatible</b> way:
     </p>
     <table class=\"releaseTable\" summary=\"summary\" border=\"1\" cellspacing=0 cellpadding=2 style=\"border-collapse:collapse;\">
+     <tr><td colspan=\"2\"><b>Buildings.Controls.OBC.ASHRAE.G36_PR1</b>
+        </td>
+    </tr>
+    <tr><td valign=\"top\">Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.VAV.Controller
+        </td>
+        <td valign=\"top\">Improved the controller implementation. <br/>
```
mwetter: Be more specific than simply "Improved the controller implementation." Write what you changed it; that it is improved is clear, otherwise there would not be a change to the model.

Also, remove the trailing space after the period.

Buildings/package.mo
```diff
@@ -196,6 +196,33 @@ its class name ends with the string <code>Beta</code>.
...
+        </td>
+    </tr>
+    <tr><td valign=\"top\">Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.VAV.SetPoints.Supply
+        </td>
+        <td valign=\"top\">Added a switch for fan control.
+        </td>
+    </tr>
+    <tr><td valign=\"top\">Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.VAV.ZoneState
+        </td>
+        <td valign=\"top\">Improved the implementation.
```
mwetter: Be more specific than simply "Improved the controller implementation."

mwetter:

PR #1527, opened 2019-08-22
Issue1265 sz controller

kuzha:
@dhblum 
I have the following updates for the SZ VAV G36 model (the baseline case remains untouched)
- I changed the maximum fan speed for heating from 1 to 0.2; this is the minimum system airflow ratio in the baseline case; in this case, the maximum heating airflow is reduced from 0.75 kg/s to 0.15 kg/s, the same as the baseline (for instance, looking at the second day)
- I changed the minimum outdoor airflow and design outdoor airflow. My observation for the moment is the quadruple damper opening parameters are more critical for the outdoor airflow. For the second day, the damper opening is around 0.4 (I reversed to the default value because I am not sure what values are reasonable); while the baseline case is constantly 0.2
- I reduced the capacity of heating and cooling coils to be consistent with the baseline; the PI control for cooling is modified
- I changed the room temperature setpoints to be consistent with the baseline;

With the above changes, the room temperature setpoints can be reached without problems. And the annual energy of SZ VAV G36 is around 16% less. 

Still a few points to note:

- The outdoor airflow is still higher in the G36 case; but it varies in different conditions; and quite a few parameters as mentioned above impact the minimum outdoor airflow, so it takes time to tune them to make the minimum consistent with the baseline case; if the outdoor airflow is reduced, the heating energy can be further reduced for the G36 case
- The maximum supply air temperature in the G36 model is set to 40C while the baseline case can reach over 50C
- The G36 room temperature and airflow can vary more than the base case and we also need to tune the PI parameters etc. to make the two cases more comparable.

There may be other things I have not paid attention but given the current status, I think it is good enough to make a PR.

@mwetter 

I spent quite some time to resolve tons of conflicts manually after merging the latest master branch. This process is error prone and I have to run unit test every time to make sure I have chosen the right merge. This has been done several times and it really takes time, so please merge this branch before you update the master branch again, otherwise I have to resolve the merging conflicts again. Or if you have better ways to manage conflicts, please do let me know.

Hopefully the Travis CI test will pass.


dhblum:
I will merge this to the origin branch and then review.

PR #1532, opened 2019-08-28
updated T&R implementation, #1530

### LINKED ISSUE #1530: TrimAndRespond uses wrong limit if resAmo is negative
A user reports:

> I am writing to report what I believe is a minor error in the implementation of the AHU supply air temperature trim and respond in the VAVReheat example. I believe the minInp micrblock in the TrimAndRespond instance in the VAVSupplyTemperature should be a maximum microblock instead of being a minimum microblock.
>
> The minInp makes sure the respond amount does not exceed the maximum respond. In the static pressure T&R a minimum microblock would implement the correct logic. Assume, respond amount = 0.15" and maximum respond = 0.12", the respond should be the minimum, accordingly, equals 0.12". The following example shows why I think it should be a maximum for a supply air temperature. Assume, respond amount = -0.8 C and maximum respond = -0.6 C, the respond should be the maximum, accordingly, equals -0.6 C.

See https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/modelica-buildings/r3KyWdjVyIo/xfvgJNLiDwAJ

To allow `resAmo` to be either negative or positive, the following changes should be made to `Buildings.Controls.OBC.ASHRAE.G36_PR1.Generic.SetPoints.TrimAndRespond`:
1. Add an assertion to test if `triAmo * resAmo >= 0` (because they must have opposite signs).
2. Add an assertion to test if `resAmo * maxRes <= 0` (because they must have the same sign).
3. Change the implementation so that it uses the `max` if `resAmo < 0`, and the `min` otherwise. (Easiest may be to convert to positive numbers and then convert the result back if `resAmo < 0`)

JayHuLBL:
This closes #1530.

- [x] added assertions to check signs of trim amount, response amount and maximum response amount.
- [x] corrected implementation for the case when response amount is negative.
- [x] updated validation model to include the case when response amount and maximum amount are negative.
- [x] update reference results.
- [x] updated release note.


JayHuLBL:

PR #1534, opened 2019-09-07
Issue1379 water source heat pump

mwetter:
This merges the latest developments [ci skip]


mwetter:
This is replaced with https://github.com/lbl-srg/modelica-buildings/pull/1536

PR #1541, opened 2019-09-11
Converted parameters to protected final constants

### LINKED ISSUE #1539: ChillerPlant.Staging.Subsequences.CapacityRequirement use constants for water properties
This model uses
```modelica
  parameter Modelica.SIunits.Density watDen = 1000 "Water density";

  parameter Modelica.SIunits.SpecificHeatCapacity watSpeHea = 4184
  "Specific heat capacity of water";
```
As there is no need for users to change these properties for water, they should be implemented as protected constants.

milicag:
Closes #1539.


milicag:
@mwetter this small edit is ready for review.

mwetter:

PR #1544, opened 2019-09-12
Issue1537 tuning refactor

### LINKED ISSUE #1537: Refactor PrimarySystem.ChillerPlant.Economizer.Subsequences.Tunin
`Buildings.Controls.OBC.ASHRAE.PrimarySystem.ChillerPlant.Economizer.Subsequences.Tuning` on branch `issue1378_staging_primarySequences` implements a 1 second sampling time. This will lead to excessive simulation times. 
Can this be implemented with a continuous time implementation?

Also, the info section does not really say what the control intent is, and it is not clear why the `pre` blocks are needed.

milicag:
Closes #1537 

- Implements `onCounter` blocks instead of algebraic loops.
- Resets the counters when they reach a very large value
- Revises info sections in all blocks in the `Economizer` package to provide more clear control intents.


milicag:
This still needs something similar to an integrator anti-windup.

mwetter:
> @mwetter Related to my previous and this PR - Is there a reason we use `final` for protected parameters but not for protected constants?

If a parameter is protected, there is not much benefit I see to also make it `final`.

milicag:
@mwetter I also added anti-windup. One thing that I most likely won't have time for before my vacation is to figure out how to prevent the counters and the anti-windup to reach large values while not interrupting the output values. Maybe the reset can happen annually.

milicag:
@mwetter this is ready for review.

milicag:

PR #1580, opened 2019-10-07
Corrected bug that caused reported P of PV panel to not include DCAC conversion

### LINKED ISSUE #1577: PVSimple bug when using DCAC efficiency
Within the information section of [_Buildings.Electrical.AC.OnePhase.Sources.PVSimple_](https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Electrical/AC/OnePhase/Sources/PVSimple.mo) the power is defined as
_P = A * f_act * eta * G * eta_DCAC_
but the connector P is connected wrong if it should follow this formula. The DCAC conversion efficiency is not part of it.
Either the info section should be adopted or the connection between the real expression _solarPower_ and the connector _P_ should be corrected.

mwetter:
This closes #1577


mwetter:

Issue #1221, opened 2018-07-23
Update CDL.Logical.Timer to reset to zero output


JayHuLBL:
In current setting when parameter `reset = false`, the timer will output accumulated time whenever the input is `true`. However, it should be able to reset to zero when it is necessary.
The issue intends to update the block so to have one additional boolean input `u0` to reset output.

mwetter:
@JayHuLBL : Should this be closed or do you plan to make a pull request? Last update was in October 2019.

JayHuLBL:
This can be closed as the change has been merged to the master.

PR #1583, opened 2019-10-07
Issue1221 reset timer

JayHuLBL:
@mwetter @milicag 
This is for #1221. 

- [x] updated the timer so that the accumulated time can be reset by Boolean input.


milicag:
@JayHuLBL Thanks, will review as soon as tests pass.

Buildings/Controls/OBC/CDL/Logical/Timer.mo
```diff
@@ -93,6 +110,11 @@ the timer will not fully stopped but hold the accumulated true input time.
 </html>", revisions="<html>
 <ul>
 <li>
+July 23, 2018, by Jianjun Hu:<br/>
+Added conditional boolean input for resetting output.  This is for
```
milicag: maybe add `for cumulative time measuring.` Also there is an extra space after the sentence.

JayHuLBL:
@milicag @mwetter 
It will be ready for review after CI test passed.

Buildings/Controls/OBC/CDL/Logical/Timer.mo
```diff
@@ -3,42 +3,59 @@ block Timer
...
   parameter Boolean reset = true
-    "Set to true for reseting timer to zero when input becomes false";
+    "Set to true for reseting timer to zero when input becomes false, set to false when need to find accumulated time which can be reset by another boolean input";
 
   Interfaces.BooleanInput u "Connector of Boolean input signal"
     annotation (Placement(transformation(extent={{-140,-20},{-100,20}})));
+  Interfaces.BooleanInput u0 if not reset
+    "Connector of Boolean for resetting output to zero"
+    annotation (Placement(transformation(extent={{-140,-90},{-100,-50}}),
+      iconTransformation(extent={{-140,-100},{-100,-60}})));
```
mwetter: The design with the parameter `reset` and an input `u0` is not intuitive. Why not calling the parameter `accumulate` (because it determines whether the timer accumulates time) and calling the input `reset` because it resets the timer when it becomes `true`.
Calling an input `u0` and having the user figure out what the difference between `u` and `u0` is is not user friendly.

Also, why not change the icon to indicate the value of `accumulate`. This would further ease graphical modeling.

Buildings/Controls/OBC/CDL/Logical/Validation/Timer.mo
```diff
@@ -1,21 +1,52 @@
...
+    annotation (Placement(transformation(extent={{-40,20},{-20,40}})));
+  Buildings.Controls.OBC.CDL.Continuous.GreaterEqualThreshold greEquThr(
+    final threshold=1.5)
+    annotation (Placement(transformation(extent={{-80,-60},{-60,-40}})));
+  Buildings.Controls.OBC.CDL.Logical.Pre pre
+    annotation (Placement(transformation(extent={{-40,-60},{-20,-40}})));
+  Buildings.Controls.OBC.CDL.Logical.Sources.Pulse booPul(
+    final width=0.5,
+    final period=2) "Block that outputs cyclic on and off"
+    annotation (Placement(transformation(extent={{-40,60},{-20,80}})));
```
mwetter: There must also be a test with `reset=true` and `u0` becoming `true` while `u` is `true`.
The expected behavior is that `y` is reset to `0` and then keeps increasing.

Buildings/Controls/OBC/CDL/Logical/Timer.mo
milicag: I like this approach as well.

JayHuLBL:
@mwetter Thanks for your comments. It will be ready for your review when the CI tests passed. Following updates were done:
- [x] changed the parameter `reset` to `accumulate` and the input `u0` to `reset`.
- [x] updated icon layer so it can display `accumulate` value.
- [x] refactored the implementation so that the output `y` will be reset to zero at the moment when `reset` has a rising edge.
- [x] added validation of the case when `reset` and `u` become true at the same time.

Buildings/Controls/OBC/CDL/Logical/Timer.mo
```diff
@@ -2,32 +2,45 @@ within Buildings.Controls.OBC.CDL.Logical;
...
+  parameter Boolean accumulate = false
+    "Set to true when need to find accumulated time which can be reset by another boolean input";
 
   Interfaces.BooleanInput u "Connector of Boolean input signal"
     annotation (Placement(transformation(extent={{-140,-20},{-100,20}})));
+  Interfaces.BooleanInput reset if accumulate
+    "Connector of Boolean for resetting output to zero"
+    annotation (Placement(transformation(extent={{-140,-90},{-100,-50}}),
+      iconTransformation(extent={{-140,-100},{-100,-60}})));
   Interfaces.RealOutput y "Connector of Real output signal"
```
milicag: @JayHuLBL Should we add `final quantity="Time", final unit="s")`

JayHuLBL:
@milicag addressed your comments
@mwetter ready for your review after CI tests passed.

Buildings/package.mo
```diff
@@ -576,8 +576,10 @@ its class name ends with the string <code>Beta</code>.
     </tr>
     <tr><td valign=\"top\">Buildings.Controls.OBC.CDL.Logical.Timer
         </td>
-        <td valign=\"top\">Updated implementation to output accumulated time when input is <code>true</code>.<br/>
-                         This is for <a href=\"https://github.com/lbl-srg/modelica-buildings/issues/1212\">issue 1212</a>.
+        <td valign=\"top\">Updated implementation to output cumulative time during which the input was <code>true</code>.<br/>
```
mwetter: @JayHuLBL : You added this to the release notes of 6.0.0, but this needs to be in 7.0.0

mwetter:
@JayHuLBL : See inline notes for misplaced release notes.

- [ ] I think the timer is wrong, see output
image

Wouldn't you expect `y` to hold its value until `u` becomes `true`. E.g., for t=1...2 and t=3..4, y should be 1

JayHuLBL:
@mwetter Thanks for the review.  
About the plot, it is what I expected. The parameter `accumulate` of timer `resetTimer` is `false`, which makes it becoming the user timer as `Modelica.Blocks.Logical.Timer`. When the input `u` is `false`, the output will be 0.

Buildings/package.mo
JayHuLBL: @mwetter 
The note was in the notes of 6.0.0, as the timer that can accumulate true time is in version 6.0.0. 
I will revert the change of the release notes of 6.0.0 but add new notes in 7.0.0 to record the changes made this time.

mwetter:
@JayHuLBL : Thanks for the feedback https://github.com/lbl-srg/modelica-buildings/pull/1583#issuecomment-552035816
Let's leave your implementation as it is consistent with the timer of the Modelica Standard Library.

PR #1591, opened 2019-10-16
Merged IBPSA

mwetter:
This will merge the updated RC model with moisture balance.


mwetter:
@AntoineGautier and @nllong : This implements the RC network model that can also model latent heat gain.
See https://github.com/lbl-srg/modelica-buildings/pull/1591/files#diff-e7171bc0d7c082c117892da96dfd7fce for instructions.

PR #1625, opened 2019-11-07
update idf/osm/script/data file for IdenticalZones

YanfeiNREL:
I updated the files for idf/osm/python-script/reformatted data. 
The idf file has 0 warnings. The timestep=10min.


mwetter:
@YanfeiNREL : Can you please make the pull request against `lbl-srg:issue1575_spawn_eplus_twozones` rather than `lbl-srg:master`.
Thanks.
You should then only see the files you changed rather than the 164 files shown here.

mwetter:
I will close it as it should be replaced with a pull request against `lbl-srg:issue1575_spawn_eplus_twozones`

PR #1638, opened 2019-11-12
Issue1636 solar

### LINKED ISSUE #1636: prescribedHeatFlowRate=false in solarcollector if energyDynamics=SteadyState
When `energyDynamics=SteadyState`, the solar collector should have `prescribedHeatFlowRate=false` since the mixing volume temperature is computed from an algebraic loop. The heat flow rate is the iteration variable and the following equation becomes a residual equation:
```Modelica
port_b.h_outflow = inStream(port_a.h_outflow) + Q_flow * m_flowInv
```

since `Q_flow` is multiplied by zero for zero flow rate, the residual equation is satisfied for any value of Q, which leads to premature convergence of the solution and:
```
Error: The following error was detected at time: 0
  In EN12975NPanels.solCol.vol[1].steBal:
   The heat flow rate equals -7.28076 W and the mass flow rate equals 0 kg/s,
   which results in a temperature difference 167841 K > dTMax=200 K.
   This may indicate that energy is not conserved for small mass flow rates.
   The implementation may require prescribedHeatFlowRate = false.
  Failed condition: noEvent(abs(solCol.vol[1].heatPort.Q_flow) < 0.008675791775999999)
```

I propose to set `prescribedHeatFlowRate= not energyDynamics==SteadyState` such that the residual equation becomes

```Modelica
0 = solCol.heaLos.QLos[10]+solCol.vol[10].dynBal.ports_H_flow[1]+
      solCol.vol[10].dynBal.ports_H_flow[2]+solCol.solGai.QSol_flow[10];
```

mwetter:
This closes #1636 
This closes https://github.com/lbl-srg/modelica-buildings/pull/1637

Todo:
- [x] See https://github.com/lbl-srg/modelica-buildings/issues/1636#issuecomment-553084901

### DIFF
--- a/Buildings/Fluid/SolarCollectors/BaseClasses/PartialSolarCollector.mo
+++ b/Buildings/Fluid/SolarCollectors/BaseClasses/PartialSolarCollector.mo
@@ -98,7 +98,7 @@ model PartialSolarCollector "Partial model for solar collectors"
     each final C_nominal=C_nominal,
     each final mSenFac=mSenFac,
     each final allowFlowReversal=allowFlowReversal,
-    each final prescribedHeatFlowRate= not energyDynamics == Modelica.Fluid.Types.Dynamics.SteadyState)
+    each final prescribedHeatFlowRate=false)
     "Volume of fluid in one segment of the solar collector"
     annotation (Placement(transformation(
         extent={{10,-10},{-10,10}},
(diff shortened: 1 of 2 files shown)


mwetter:
This is ready to merge when the tests pass.

PR #1685, opened 2019-12-17
Issue1660

### LINKED ISSUE #1660: Unexpected cfdSendStopCommand function behavior on Linux environment
Dear all,
we found an unexpected behavior of the "cfdSendStopCommand" function when running the Modelica-FFD cosimulation examples on Linux environment. Although the Modelica model always finishes successfully, the FFD does not, generating the error message "Cannot stop the FFD simulation in required time".
After investigating the issue, a possible problem could be represented by the definition of the "Sleep" macro in the "cfdCosimulation.h" file for the Linux environment, that is:
`#define Sleep(x) sleep(x/1000)`.
The problem could be that the "sleep()" function takes an "unsigned int" argument. So, when in the "cfdSendStopCommand" function the argument "10" is passed to the "Sleep" function, the value passed to the "sleep()" function should be "10/1000", which is rounded to "0". In this way the "Sleep()" function would not work as expected, that is, it would not wait for 100 seconds (10/1000 *imax).
We have found a workaround by modifying the code in this way:
`#define Sleep(x) sleep(ceil(x/1000.0))` in the "cfdCosimulation.h" file;
`#if defined(_MSC_VER) || defined(__WIN32__) /* Windows */
  size_t imax = 10000;
#else /* Linux*/
  size_t imax = 100;
#endif` in the "cfdSendStopCommand.c" file.
The values may vary depending on the desired waiting time; the used values are tested for waiting 100 s.

Could anybody help us about this issue, please?

Thanks a lot.

mwetter:
This closes #1660 

@JayHuLBL : 
- [x] Can you please regenerate the binaries for Linux 32 bit, Windows 32 bit and Windows 64 bit.


JayHuLBL:
@mwetter 
On windows OS (64 bit):
- [x] added win64 binaries, which were not in the `Resources/Library/win64/` folder previously,
- [x] updated win32 binaries,
- [x] successfully tested the FFD models with Dymola.

On Linux (Ubuntu 18.04, 64 bit):
- [x] updated linux32 binaries,
- [x] regenerated linux64 binaries but it seems no change,
- [x] successfully run unit test of the package `Buildings.ThermalZones.Detailed.Examples.FFD`, both with Dymola and JModelica.

mwetter:

PR #1696, opened 2020-01-03
Issue1556 ntu cool tower

### LINKED ISSUE #1556: add epsilon-NTU cooling tower model
Add a Merkel-theory based cooling tower model to support chiller plant modeling.

@mwetter Dr. Wetter, can you make a new branch for this issue?

mwetter:
This closes #1556


mwetter:
@khinkelman : FYI, please see
https://github.com/lbl-srg/modelica-buildings/blob/b63f09c337b0e44ccbc26038135c9cfb16099d31/Buildings/Fluid/HeatExchangers/CoolingTowers/BaseClasses/Functions/equivalentHeatCapacity.mo
for a new formulation which is differentiable. This avoid the need for a numerical Jacobian.

mwetter:
@khinkelman : Thanks for the pull request. The code looks very good. I did some refactoring and revisions. Could you please address the following below before it can be merged to the master:

- [ ] Add validation for `Functions/equivalentHeatCapacity.mo`, as we want a small validation for every function. This is much easier to catch errors as opposed to when the function is only tested as part of a bigger model.
- [ ] Expand the info section for `Buildings.Fluid.HeatExchangers.CoolingTowers.Merkel` so that users understand at a high level what the main model equations are without having to check the EnergyPlus documentation.
- [ ] I don't quite understand the use case for `Buildings.Fluid.HeatExchangers.CoolingTowers.Data.UAMerkel`. If these need to be changed by the user, then their info section needs to be expanded so that users know how these coefficients are used. If the user will never need to change these values, then they could be directly entered in `Buildings.Fluid.HeatExchangers.CoolingTowers.Merkel`. Currently, it is not clear how `UAFunDifWB` etc. are used, and whether they need to be updated based on the size and type of cooling tower.

Please use the latest version from this branch so you get all the changes.

I can do the following
- [ ] Add the missing info sections at `PartialStaticTwoPortCoolingTower` and at `PartialStaticTwoPortCoolingTowerWetBulb`
- [ ] Add the missing info section at `Correlations.Examples.YorkCalc`

mwetter:
The following also need to be addressed in `Merkel.mo`:

1) The code contains
```modelica
  final parameter  Modelica.SIunits.Temperature T_a1_nominal = TAirInWB_nominal
    "Nominal temperature at port a1"
  final parameter Modelica.SIunits.Temperature T_a2_nominal = TWatIn_nominal
    "Nominal temperature at port a2"
  final parameter Modelica.SIunits.Temperature T_b1_nominal = TAirOutWB_nominal
    "Nominal temperature at port b1"
  final parameter Modelica.SIunits.Temperature T_b2_nominal = TWatOut_nominal
    "Nominal temperature at port b2"
```
I see no reason why to introduce another parameter name for the same quantity. This is confusing.

Also, it appears that the model has too many parameters: All flow rates and all temperatures at nominal condition are known. From these, it should be possible to compute `Q_flow_nominal` and `UA_nominal`, but these two are also parameters that the user needs to specify. The model would be easiest to use in my opinion if temperatures (range and approach) and water flow rate are specified as parameters, and from these, the other quantities are computed. Then, the model is scalable. Maybe the same quantities can be used for approach and range temperature as are used for `YorkCalc`, which has the advantage that by changing `m_flow_nominal`, the size of the tower can be scaled up and down, whereas `Merkel` has various parameters that must be adjusted simultaneously.

khinkelman:
Thanks @mwetter for your helpful comments and additions, and I'll be addressing your requests this week. For your comment on too many parameters: I agree the current formulation is not ideal and should be revised. Originally, we were calculating `UA_nominal` from `Q_flow_nominal` and other parameters based on the epsilon-NTU method. However, this method calculated `UA_nominal` differently than EnergyPlus (and EnergyPlus does not provide its calculation method in the documentation). Hence, we decided to have the user specify `UA_nominal` to avoid this discrepancy. I will look into the method employed for `YorkCalc` to see if it solves the above discrepancy, which I have not yet done. If the error persists, we can discuss preferred solutions.

mwetter:
@khinkelman : That sounds good. For the EnergyPlus method, you may need to look into the source code at https://github.com/NREL/EnergyPlus/tree/develop/src/EnergyPlus if the documentation is not sufficient.

khinkelman:
Hi @mwetter, I have addressed all of your comments to the best of my knowledge. Please see the revisions on my forked branch [here](https://github.com/khinkelman/modelica-buildings/tree/issue1556_ntu_coolTower). I was able to find the problem in our epsilon-NTU method, and so the merkel model is using this approach (similar to the EnergyPlus equivalent model). Please let me know if you need anything else.

mwetter:
@khinkelman : Thanks for your revision which were very helpful. I did another review/revision of the model. The main changes are
- renamed of variable to use `Air` and `Wat` rather than 1 and 2
- removed of `TApp_nominal` which was nowhere used
- introduction of `TWatOut_nominal` from which now `Q_flow_nominal` is computed.
- introduction of `raPFan_nominal` to provide a default value for `PFan_nominal` using a rule of thumb.
- changed sign of `Q_flow_nominal` which is now negative because the cooling tower cools the water. This definition is now consistent with cooling coils.

Can you please address the following:
Todo:
- [ ] Update documentation of `Buildings.Fluid.HeatExchangers.CoolingTowers.Merkel`, in particular it is not clear how to coefficients in `UACor` are used. Best would be to state the polynomial in the info section.
- [ ] The model has a parameter `mAir_flow_nominal`. Can you please check the ASHRAE Handbook to see what rule of thumb could be used. I would think it should depend on `mWat_flow_nominal`. Most users probably would not know what a good default value is.

khinkelman:
Thanks @mwetter, I greatly appreciate your revisions as well and your guidance. I will address these to-do items over the weekend.

khinkelman:
Hi @mwetter, I have addressed the requested to-do items on my forked branch. Please let me know if you need any further revisions. Thanks!

khinkelman:
As a heads up, I made a correction to the documentation for the merkel model (to the list of input parameters). Please see the commit on my forked branch.

mwetter:

PR #1703, opened 2020-01-08
Optimal start

kuzha:
@mwetter 

Please review the files related to optimal start block and open-loop validation. I did not include closed-loop examples in this PR because the examples use the singleZoneFloor that I created based on the five-zone floor model. I will do a separate PR for the singleZoneFloor and another PR for the closed-loop examples.

I merged the latest master branch to my local optimalStart branch and this results in 258 files changes. To see the file changes only from me, please merge master to the remote branch issue1589_optimalStartUp.

Thank you,
Kun


mwetter:
@kuzha : I did a brief review of the PR. Can you please address the following:
- [ ] The parameters, inputs and outputs and their comments should be self-explanatory without the user necessarily having to read the info section (and in particular not getting into the algorithm of the info section for example to understand what `uLow` and `uHig` means). Similar, the meaning of `start` is not known. Easier is to write something like "if true, system should run" [or be started, but started is only done once, while the output remains true for a quite a while]
- [ ] The output `start` need to be included in the regression tests
- [ ] The parameters `heating_only` and `cooling_only` are not good as they would allow users to select `heating_only = true` and `cooling_only = true`. Better would be something like `computeHeating` and `computeCooling`.
- [ ] if `n` were renamed to `nDay` then it would be clearer from the variable name what it is.
- [ ] Please go through the implementation and make protected what the user is not likely to be interested in. For example, blocks called `hys`, `temSlo1, `swi1` need not be public. Check what results are truly of interest to the typical user, and make the others protected. (The can still be investigated if needed by asking Dymola to output protected variables).
- [ ] Add graphics to the icon so the blocks are recognizable.
- [ ] Gains in `dtHea` should be final, no?
- [ ] `StaCal` and `StoCal` are inputs and hence should start with lower case.
- [ ] In the `info` section of `Utilities.OptimalStart`, adding an algorithm description is good, but please first explain at a high level what the main parameters, inputs, outputs and the underlying methods are so users are not forced to read the whole algorithm to understand how to use the block.
- [ ] The `info` section of the validation model need to explain the intention of the test. It currently only states that it validates the block, which is clear as it is a validation model. But what do you test? What are you trying to test by having a temperature input with a period that is not one day?
- [ ] Work with Jianjun to make sure the output of the block can integrate with the the sequences of G36.
- [ ] We need to see some closed loop tests before merging to make sure the calculations are correct.

kuzha:
@mwetter : I have addressed your comments, please see the bolded font for my short replies:
 * [x]  The parameters, inputs and outputs and their comments should be self-explanatory...        
_**Changed**_
 * [x]  The output `start` need to be included in the regression tests  
_**Added**_
 * [x]  The parameters `heating_only` and `cooling_only`...  
**_The user cannot select `heating_only = true` and `cooling_only = true` at the same time: when one parameter is true, the other is forced to be false. This setting is enforced in the annotation of these two parameters_** 
 * [x]  if `n` were renamed to `nDay` then it would be clearer from the variable name what it is.
_**Changed**_
 * [x]  Please go through the implementation and make protected what the user is not likely to be interested in. For example, blocks called `hys`, `temSlo1, `swi1` need not be public. Check what results are truly of interest to the typical user, and make the others protected. (The can still be investigated if needed by asking Dymola to output protected variables).
_**Changed**_
 * [x]  Add graphics to the icon so the blocks are recognizable.
_**Added**_
 * [x]  Gains in `dtHea` should be final, no?
_**Added**_
 * [x]  `StaCal` and `StoCal` are inputs and hence should start with lower case.
_**Changed**_
 * [x]  In the `info` section of `Utilities.OptimalStart`, adding an algorithm description is good, but please first explain at a high level what the main parameters, inputs, outputs and the underlying methods are so users are not forced to read the whole algorithm to understand how to use the block.
_**Added the descriptions as suggested and relocated the algorithm to the last section of the `info`**_
 * [x]  The `info` section of the validation model need to explain the intention of the test. It currently only states that it validates the block, which is clear as it is a validation model. But what do you test? What are you trying to test by having a temperature input with a period that is not one day?
_**Added details in the `info` section**_
 * [x]  Work with Jianjun to make sure the output of the block can integrate with the the sequences of G36.
_**Checked**_
 * [ ]  We need to see some closed loop tests before merging to make sure the calculations are correct.
_**I will add closed-loop tests in this branch later**_

kuzha:
@mwetter 
I also pushed my updates for the SingleZoneFloor model in this branch. This model is used for the closed-loop tests for optimal start. Please let me know if pushing to the same branch is okay; otherwise I can revert my PR.

Note that I also found a tiny error in the model Buildings.Examples.VAVReheat.ThermalZones.Floor and corrected in the updates. This may impact slightly the results of Buildings.Examples.VAVReheat.ASHRAE2006 and Guideline 36, but I think you should review the changes first.

kuzha:
> @mwetter
> I also pushed my updates for the SingleZoneFloor model in this branch. This model is used for the closed-loop tests for optimal start. Please let me know if pushing to the same branch is okay; otherwise I can revert my PR.
> 
> Note that I also found a tiny error in the model Buildings.Examples.VAVReheat.ThermalZones.Floor and corrected in the updates. This may impact slightly the results of Buildings.Examples.VAVReheat.ASHRAE2006 and Guideline 36, but I think you should review the changes first.

The SingleZoneFloor model will be pushed in a new PR after modifying the validation model

PR #1722, opened 2020-01-25
Issue1699 absorption indirect hot water and steam with switchable records

Helarga:


Helarga:
@mwetter The indirect absorption chiller- hot water or steam as the generator heat source and switchable records for both models.
 It includes modified model for the indirect absorption chiller
Still lacks the EP idf for the hot water.

PR #1730, opened 2020-01-31
Issue1589 optimal start up

### LINKED ISSUE #1589: add optimal startup block
Add a block for optimal start-up to the `OBC` package.

mwetter:
This closes #1589


mwetter:
Todo:
- [ ] add validation case that has `computeHeating = computeCooling = false`
- [ ] change validation case with heating and cooling to have 2 sequences of optimal heating and optimal cooling calculation
- [ ] Make a test with `nDay = 6` and make sure the initial few days for `tOptOn` are independent of `nDay`
- [ ] Change name and/or description of `tOptIni` which is used `tOptDef`

mwetter:
@kuzha : Can you please address the items below and make a new PR. Please update your branch first.

Todo:
- [ ] Replace in the examples the `TimeTable` from the Modelica Standard Library with `CDL.Continuous.Sources.TimeTable` (I did this already for `OptimalStartHeating`)

- [ ] Correct the implementation, as the following simulation produces wrong results, see trajectory `optStaHea.TZon`:
```
simulateModel("Buildings.Controls.OBC.Utilities.Validation.OptimalStartHeating", startTime=-660000, stopTime=0, method="Cvode", tolerance=1e-06, resultFile="OptimalStartHeating");
```
and add a validation test for a case where the simulation starts at a negative time such as above.

mwetter:
@kuzha : the test model `OptimalStartHeating` still produces results that I cannot explain. Please run
```
simulateModel("Buildings.Controls.OBC.Utilities.Validation.OptimalStartHeating", startTime=-1296000, stopTime=432000, method="Cvode", tolerance=1e-06, resultFile="OptimalStartHeating");
```
The value `tOpt` changes near t=0. This indicates a bug in the implementation as there is nothing that should be changing at this time. Moving the start time does not change where this change happens.

image

Please address this issue.

kuzha:
@mwetter 

I checked the model and think it may relate to the start time of the sine signal.

Could you please also change TOut.startTime=-1296000s in your test model so that it is consistent to the simulation start time?

My own test shows that tOpt does not change near t=0 once I made this change.

mwetter:
@kuzha : Yes, it is because of `TOut.startTime`.

PR #1800, opened 2020-03-01
Issue1298 dampers refactor

AntoineGautier:
This PR includes the modifications from IBPSA [PR#1267](https://github.com/ibpsa/modelica-ibpsa/pull/1267) and [PR#1300](https://github.com/ibpsa/modelica-ibpsa/pull/1300).
Additionally the example models from Buildings have been updated with the conversion script that is included in this PR. Please note:

- In most cases this leads to `dpDamper_nominal = (0.45)*1.2*(1)^2/2 = 0.27`. This has not been changed yet but we might consider setting more representative values, e.g. 10 Pa.
- Simulating `Buildings.Examples.VAVCO2.VAVSystemCTControl` with Dymola gives the following error.
```
Previous problem occured when evaluating crossing function, reducing step-size
cvode: CVODE cvRcheck3 At t = 38346.4, the rootfinding routine failed in an unrecoverable manner.
Unexpected value: -12 at time 38016
Unexpected CVode value
```
However:

- The model simulates with Dassl: this is the new setting proposed in the simulation script.
- The model simulates with CVode when "Output selection/Equidistant time grid" is disabled!
- The model simulates with CVode using Optimica or JModelica.

Here are some observations after investigating what could be the origin of the error.

The states dominating the error are the following:
error_dom

When disabling the VAV damper input filter the error persists. It thus seems related to the fan control loop. 

Here is the integrator step size with Radau.
step_size
Surprisingly, in the time frame where CVode error occurs (~ 38000 s), Radau relaxes slightly the time step and rejects less steps than between 36000 and 38000 s.
The state dominating the error has the following error statistics.
state_error
But again the highest contribution to the error estimate happens before 38000 s and close to 40000 s, not when CVode throws the error.
Eventually, to provide an order of magnitude of the impact on the simulation results, I put here some typical variable trajectories comparing the development branch to the master, as well as an error statistic on a large set of parameters and variables. 
ramLim y
roo roo45 vav dp
mixBox damRec dp
bar


AntoineGautier:
@mwetter This is ready to merge when all tests pass.
(This does not close #1298 yet.)

PR #1835, opened 2020-03-13
Corrected implementation for negative start times

### LINKED ISSUE #1834: CDL.Continuous.Sources.TimeTable assumes startTime=0
If the simulation is started at a negative number, this model produces a constant output because it sets `startTime=0` in the data table. 

The model needs to be changed so that it works for any simulation start time.

mwetter:
This closes #1834


JayHuLBL:
@mwetter It's ready to merge.

PR #1836, opened 2020-03-14
Merged IBPSA, issue1206_media_steam

mwetter:
This merges `IBPSA`, ` issue1206_media_steam`.
See todo list at https://github.com/ibpsa/modelica-ibpsa/pull/1320


mwetter:
I will merge it, and once we built a steam system model, we will be looking into the size of the nonlinear equations of this model vs a media that uses `d=dStp * p/pStp`.

PR #1857, opened 2020-03-27
Update TRad boundary check

kbenne:
A error is issued based on value limits that were based on degC, however
the units are now K. Updated the check accordingly.


mwetter:
This has been integrated in `issue1863_spawn_eplus_9_3`

mwetter:
This will be fixed in af11f1a55

PR #1859, opened 2020-03-28
Issue1853 lim partype

### LINKED ISSUE #1853: Limit parameter types in CDL sequences
In `Buildings.Controls.OBC.ASHRAE` package, there are some parameter declarations like
```
parameter Modelica.SIunits.Time samplePeriod=120;
```
This issue is to correct these parameter declarations to avoid the type `Modelica.SIunits.xxx`, and to limit the types to `Real`, `Integer`, `Boolean`, `String`, and `enumeration`.

JayHuLBL:
- [x] corrected parameter declarations to avoid use type like `Modelica.SIunits.xxx`

This closes #1853.


JayHuLBL:
- [x] in package `Buildings.Controls.OBC.ASHRAE.G36_PR1`, eliminated usage of `Evaluate=true` annotation in parameter declaration.

@mwetter It's ready for your review. After this PR being merged, I will then address potential issue in PR #1862.

PR #1874, opened 2020-04-04
the order of public elements in the package.mo needs to be the same a…

adrpo:
…s the order in package.order

Spec says:
Classes and constants that are stored in package.mo are also present in package.order but their relative order should be identical to the one in package.mo (this ensures that the relative order between classes and constants stored in different ways is preserved).

### DIFF
--- a/Buildings/Types/Azimuth/package.order
+++ b/Buildings/Types/Azimuth/package.order
@@ -1,8 +1,8 @@
-E
-N
-NE
-NW
 S
-SE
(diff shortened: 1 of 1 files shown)


mwetter:
@adrpo : Thanks for reporting. I will correct this first in BuildingsPy which writes `package.order` (and tests this file as part of the CI testing). See https://github.com/lbl-srg/BuildingsPy/pull/342

PR #1903, opened 2020-04-21
Ibpsa sync issue1340 weather data

mwetter:
This merges the revised weather data reader from https://github.com/ibpsa/modelica-ibpsa/issues/1340


mwetter:

PR #1918, opened 2020-04-29
Issue1873 pre ind dam

JayHuLBL:


JayHuLBL:
@AntoineGautier It will be ready for your review after tests passed.

AntoineGautier:
@JayHuLBL `Buildings.Examples.VAVReheat.Controls.DuctStaticPressureSetpoint` has an incorrect setting for the PI action. It should be reverse, not direct: when the maximum opening exceeds the set-point, the tracking error is negative, and we want to increase the duct pressure (so that the dampers close) i.e. a positive output.
Do you concur?
Could you please address my inline comments and correct this?

@kuzha Did you also notice this?

Buildings/Controls/OBC/ASHRAE/G36_PR1/TerminalUnits/Reheat/DamperValves.mo
```diff
@@ -350,19 +354,24 @@ protected
...
     "Nominal volume flow rate"
-    annotation (Placement(transformation(extent={{200,110},{220,130}})));
+    annotation (Placement(transformation(extent={{200,200},{220,220}})));
   Buildings.Controls.OBC.CDL.Continuous.Division VDisSet_flowNor
     "Normalized setpoint for discharge volume flow rate"
-    annotation (Placement(transformation(extent={{240,180},{260,200}})));
+    annotation (Placement(transformation(extent={{240,220},{260,240}})));
+  Buildings.Controls.OBC.CDL.Continuous.Gain gai(
+    final k=1) if have_pressureIndependentDamper
+    "Block that can be disabled so remove the connection"
```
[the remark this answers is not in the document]
JayHuLBL: Thanks.

AntoineGautier:
Here is the impact analysis on the computational performance. Three branches are compared.

1. `master_ashrae2006` is identical to master except that the simulation period for `Examples.VAVReheat.ASHRAE2006` has been changed from `stopTime=172800` to `startTime=16934400, stopTime=17748000`. The reason for this change is twofold. 

    - Firstly the initial simulation period has a low outdoor temperature for which the duct pressure reset is disabled. 
    - Secondly the building model has a rather heavy construction so it takes several days (even in summer conditions) to condition the model and hit the room cooling set-point.

    We should perhaps consider changing also the simulation period in master, perhaps not for such a long period but at least to a period where the pressure reset is enabled. @mwetter What do you think?

2. `issue1873_preIndDam_noopening` takes into account the refactoring of the VAV models from the current feature branch but with the pressure independent damper model from master, i.e. without the computation of the damper opening. (And the same simulation period as `master_ashrae2006` for `Examples.VAVReheat.ASHRAE2006`.) 

3. `issue1873_preIndDam_opening` takes into account the refactoring of the VAV models from the current feature branch, including the refactored pressure independent damper model, i.e. with the computation of the damper opening. (And the same simulation period as `master_ashrae2006` for `Examples.VAVReheat.ASHRAE2006`.) 

The impact is assessed only for the package `Buildings.Examples` which is the only one that contains models where the damper opening is used in closed loop control.
Either with OCT or Dymola the impact on CPU time appears marginal so I would greenlight merging the current PR.

Screen Shot 2020-05-01 at 3 47 57 PM

Screen Shot 2020-05-01 at 3 48 07 PM

Buildings/Resources/Scripts/Dymola/ConvertBuildings_from_6_to_7.0.0.mos
```diff
@@ -80,6 +80,14 @@ convertElement(
...
     "alphaWinOut",
     "hConWinOut");
+convertElement(
+    "Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Reheat.SystemRequests",
+    "uDam",
+    "yDam_actual");
+convertElement(
+    "Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller",
+    "uDam",
+    "yDam_actual");
```
mwetter, line 90: This non-backward compatible change must also be documented in the release notes of the top-level `package.mo` file.

Buildings/Fluid/BaseClasses/FlowModels/basicFlowFunction_dp_m_flow.mo
```diff
@@ -0,0 +1,43 @@
...
+protected
+  Modelica.SIunits.PressureDifference dpSqrt
+    "Regularized square root value of pressure drop";
+  Modelica.SIunits.MassFlowRate mPos_flow
+    "Regularized absolute value of mass flow rate";
+algorithm
+  dpSqrt := Buildings.Utilities.Math.Functions.regNonZeroPower(
+    dp, 0.5, dp_small);
+  mPos_flow := Buildings.Utilities.Math.Functions.regNonZeroPower(
+    m_flow, 1, m_flow_small);
```
mwetter, line 25: Is there a particular reason for using `regNonZeroPower` with exponent `1` rather than using `smoothMax(m_flow, -m_flow, m_flow_small)`? If not, then `smoothMax` would be preferred as it is more efficient.

AntoineGautier: @mwetter `regNonZeroPower` is used originally to guard against a zero value since we need to take the inverse of the flow coefficient when calling
```
Buildings.Fluid.Actuators.BaseClasses.exponentialDamper_inv(
      kThetaSqRt=sqrt(2*rho)*A/kDam, kSupSpl=kSupSpl, ySupSpl=ySupSpl, invSplDer=invSplDer)
```
However, thinking again about it, it would be more efficient to use the actual value of the loss coefficient `kTheta` in the above equation and to compute directly the square inverse of the flow coefficient in `basicFlowFunction_dp_m_flow` as

```
  output Real kSquInv
    "Square inverse of flow coefficient";
protected 
  Modelica.SIunits.PressureDifference dpPos=
    Buildings.Utilities.Math.Functions.smoothMax(dp, -dp, dp_small)
    "Regularized absolute value of pressure drop";
  Real mSqu_flow = Buildings.Utilities.Math.Functions.smoothMax(
    m_flow^2, m_flow_small^2, m_flow_small^2)
    "Regularized non zero square value of mass flow rate";
algorithm 
  kSquInv := dpPos / mSqu_flow;
```

The drawback is that the input variables of `exponentialDamper_inv` and `basicFlowFunction_dp_m_flow` are no more consistent with the variables of the original functions `exponentialDamper` and `basicFlowFunction_dp`. But I think this is acceptable as the inverse functions are unlikely to be used outside of the pressure independent model.

Please tell me if you think the changes in commit [5837774](https://github.com/lbl-srg/modelica-buildings/pull/1918/commits/58377741dc08e3d9b072c055a8bdfc3670981914) are then worth considering.

mwetter: I think this is fine; but if they are no longer consistent and only used in that model, you could put the function directly in `protected` section of the model. Then no other model can use it, and consistency is less important.

AntoineGautier: Here is the updated performance comparison with Dymola
Screen Shot 2020-05-04 at 3 14 56 PM
and Optimica
Screen Shot 2020-05-04 at 3 14 41 PM
So no noticeable impact.

Buildings/Resources/Scripts/Dymola/ConvertBuildings_from_6_to_7.0.0.mos
JayHuLBL: resolved

AntoineGautier:
On my side this is ready to merge when all tests pass.

PR #1966, opened 2020-05-22
Issue1936 fanger Added a validation file

kim1077:
Added a validation file which compares the updated PMV model and CBE Thermal Comfort Tool.
There is **one more issue.** 
The results from the latest Fanger.mo (Michale revised version) was different from that of Daniele's original modelica model (see the result comparison from the PMV_tester.mo). 
Michael, would you please double check the changes you made?


mwetter:
@kim1077 : Please use Daniele's version if this agrees to the equations from CBE, clean it up as needed (also to minimize diffs so it is clear what changed), and address the following:
- [ ] Add in `Fanger.mo` in the revision notes, and in `Buildings/package.mo` in the release notes, a short description of what changed. We want to inform users what changed for what reason.
- [ ] Correct the entry in `Examples.PMV_tester` that points to the .mos script so it points to the correct script.
- [ ] Add in `Examples.PMV_tester` an info section that describes the validation case, and revision notes. The current info section is not informative as it is clear that the test model tests a model. Think about what a user needs to know about the model.
- [ ] Change the validation model to conform to the convention for how instances are names, see https://simulationresearch.lbl.gov/modelica/releases/latest/help/Buildings_UsersGuide.html#Buildings.UsersGuide.Conventions (three letters, camel case, starting with lower case).
- [ ] Remove commented out code if they are no longer needed.
- [ ] Make sure the new code is well defined, for example, the expression `(TClo - TAir)^0.25` does not work for `TClo<TAir`.
- [ ] Move `Examples.PMV_tester` to something like `Validation.FangerCBE`. The suffix `_tester` is not good as all examples are testing models.
- [ ] Update the reference results by running the unit tests for `Buildings.Utilities.Comfort`.
- [ ] Explain, or correct, the large differences in the validation at `t=1500...1700`
image

PR #1987, opened 2020-06-03
Issue1986 restructure dhc

### LINKED ISSUE #1986: Restructure Applications.DHC package
This issue is to update the package structure based on @khinkelman proposition.

AntoineGautier:
This closes #1986.


JayHuLBL:
@AntoineGautier 
- If the `CentralPlants` can be quite generalized so that they can be used for district system and single building system (which would be very nice), then keeping `CentralPlants` outside of the DHC package would be better.
- It would be better to have the `BaseClasses` package in `Buildings.Applications.DHC.CentralPlants.Combined`, `.Cooling` and `.Heating`, rather than have `Buildings.Applications.DHC.CentralPlants.BaseClasses`.

JingWang-CUB:
@AntoineGautier Thanks for restructuring the package! I have a few questions:

1. What kind of model is going to be put in `Buildings.Applications.DHC.EnergyTransferStations.Cooling.Subsystems`?  Currently, from cooling and 1st generation heating side, we don't have the need for subsystem models. Same for the `Buildings.Applications.DHC.EnergyTransferStations.Cooling.Controls`, the control blocks have been embedded inside ETS models.

2. We actually have subpackages such as `Buildings.Applications.DHC.Examples.Cooling.BaseClasses` where the coupled building load time series and ETS model is located. This is needed as a component for our network modeling. Do you have better ideas about where to put those models?

In response to @JayHuLBL 's comments:

1. From our side, the plant models are generalized and could be applied to districts and single buildings.

Thank you!

AntoineGautier:
@JayHuLBL Regarding the location of `BaseClasses` I anticipate that both locations (directly under `CentralPlants` or under the subpackages of `CentralPlants`) might be needed. For instance, we may need generic junction and pump models that will be in the top `BaseClasses` package. I propose that each developer further add a `BaseClasses` package in the subpackages as needed. 
If that is fine with you please let me know if the branch can be merged.

PR #1991, opened 2020-06-04
Issue1925 staging boiler plant sequences

karthikeyad-pnnl:
@milicag: Can you please approve this pull request, so that the branch is up-to-date with previously approved and merged commits in the staging branch?

Thanks,
Karthik


milicag:
@karthikeyad-pnnl How come you need this PR? As this is not adding new content and if you'd just like to update the 1953 branch with the fresh 1925, I can do that without a PR and then you can update your fork. Would that work do you think?

PR #1997, opened 2020-06-08
Ibpsa sync issue1365

mwetter:
This merges IBPSA issue 1365 except for the steam models.

It also changes `reverseAction` to `reverseActing`, and removed the assignment of `SimpleController.PI` which is now the default.


mwetter:
@dhblum : Can you please review this, in particular whether the explanation of `reverseActing` is clear in the four PID controllers in `Buildings.Controls.Continuous` and in `Buildings.Controls.OBC.CDL.Continuous`.

dhblum:
@mwetter The explanation looks clear to me and implementation in accordance with the description given in Montgomery and McDowall 2008.

PR #2014, opened 2020-06-22
Rearranged icons, removed non-used block for occupant density

mwetter:
This is needed for the OBC final report


dhblum:
Will review this morning.

PR #2017, opened 2020-06-25
Issue1893 operation mode

### LINKED ISSUE #1893: Warm up and cool down modes are not activated in the operation mode
The block Buildings.Controls.OBC.ASHRAE.G36_PR1.Generic.SetPoints.OperationMode has 7 operation modes. 

Two of the seven modes, warm up and cool down modes, are never activated.

Another issue is: TZonMax and TZonMin are used as inputs for these two modes; but using TZon is enough according to the G36 description. This can slightly simplify the implementation.

- [ ] Needs to set the warm up and cool down time to 1800s in the VAV reheat G36 model so that it is consistent to that in the ASHRAE 2006 model

JayHuLBL:
This closes #1893. 
@kuzha Could you please review following changes?
- removed `Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.ModeAndSetPoints`. As for multizone VAV system, we cannot compose the `Buildings.Controls.OBC.ASHRAE.G36_PR1.Generic.SetPoints.OperationMode` and `Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.SetPoints.ZoneTemperatures` to one class, since we'd need to avoid vector-related calculation. Check `Buildings.Examples.VAVReheat.Guideline36` to see how to find the group operating mode and zones setpoint.
- added `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.VAV.SetPoints.ModeAndSetPoints` to be used for single zone AHU controller.
- updated `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.VAV.Controller`. 
- added `Buildings.Controls.OBC.ASHRAE.G36_PR1.Generic.SetPoints.GroupStatus`.
  - check if the group should be occupied mode, calculate the time to next group occupied period
  - find maximum cooldown and warm-up time
  - check if the group should be in cooldown or warm-up mode by comparing the zone temperature and the occupied setpoints
  - find out total cold/hot zone by comparing the zone temperature with unoccupied setpoints
  - compare the average zone temperature with the average unoccupied setpoints, to indicate if the group should be in setback (or setup mode).
  - check if the setback/setup mode should end, if all zones indicate the modes could end.
  - find total number of zones with opening window
- updated `Buildings.Controls.OBC.ASHRAE.G36_PR1.Generic.SetPoints.OperationMode` according to latest G36
- updated `Buildings.Examples.VAVReheat.Guideline36`

Could you please also check if the changes affect the annual results?


Buildings/Controls/OBC/ASHRAE/G36_PR1/Generic/SetPoints/ZoneStatus.mo
```diff
@@ -1,380 +1,420 @@
...
-    annotation (Placement(transformation(extent={{-20,10},{0,30}})));
+    annotation (Placement(transformation(extent={{0,80},{20,100}})));
+  Buildings.Controls.OBC.CDL.Continuous.Add add1(
+    final k1=-1)
+    "Calculate difference between zone temperature and the occupied cooling setpoint"
+    annotation (Placement(transformation(extent={{-40,0},{-20,20}})));
   Buildings.Controls.OBC.CDL.Continuous.Hysteresis hys1(
-    final pre_y_start=false,
     final uLow=-0.1,
     final uHigh=0.1)
```
kuzha, line 139: - uLow and uHigh settings of hys and hys1 should be exposed as parameters for this block so that the user can change them, for example, when coupling with optimal start block. uLow of hys and hys1 can be set as one parameter, so does uHigh of hys and hys1.
- I suggest to use two parameters separately for uLow and uHigh, instead of using one parameter "bouLim" and setting uLow=-0.5*bouLim and uHigh=0.5*bouLim. The reason why I suggest this is because in the optimal start, I set the default uLow to be non-negative. uLow is used to judge when the zone temperature reaches the setpoint, and setting uLow to be zero helps to narrow down the deviation between zone temperature and setpoint (comparing with -0.5). Setting uLow to positive helps to mitigate corner cases that the zone temperature takes too long or never reaches the setpoint. Please correct me if I am wrong here.

Buildings/Controls/OBC/ASHRAE/G36_PR1/Generic/SetPoints/ZoneStatus.mo
```diff
@@ -1,380 +1,420 @@
...
-    "Hysteresis that outputs if the zone temperature is lower then setpoint"
-    annotation (Placement(transformation(extent={{80,-70},{100,-50}})));
-  Buildings.Controls.OBC.CDL.Continuous.Add add3(final k2=-1)
-    "Calculate the difference between minimum zone temperature and unoccupied heating setpoint"
-    annotation (Placement(transformation(extent={{40,-120},{60,-100}})));
-  Buildings.Controls.OBC.CDL.Continuous.Hysteresis hys3(
-    final uLow=-0.5*bouLim,
-    final uHigh=0.5*bouLim,
```
kuzha, line 152: uLow of hys2 and hys3 should remain the same so that the setback mode can be activated. So does uHigh of hys2 and hys3. For this one, you can set one parameter bouLim or set two parameters separately uLow and uHigh as I suggested for hys and hys1. This is up to you to decide.

Buildings/Controls/OBC/ASHRAE/G36_PR1/Generic/SetPoints/ZoneStatus.mo
```diff
@@ -1,380 +1,420 @@
...
     final uHigh=0.1)
-    "Hysteresis that outputs if the zone temperature is higher than setpoint"
-    annotation (Placement(transformation(extent={{80,-230},{100,-210}})));
-  Buildings.Controls.OBC.CDL.Continuous.Add add4(final k2=-1)
-    "Calculate the difference between maximum zone temperature and unoccupied cooling setpoint"
-    annotation (Placement(transformation(extent={{40,-180},{60,-160}})));
-  Buildings.Controls.OBC.CDL.Continuous.Hysteresis hys4(
-    final pre_y_start=false,
-    final uLow=-0.5*bouLim,
-    final uHigh=0.5*bouLim)
```
kuzha, line 171: The same rule applies to hys4 and hys5 as that of hys2 and hys3.

Buildings/Controls/OBC/ASHRAE/G36_PR1/AHUs/SingleZone/VAV/Controller.mo
```diff
@@ -281,6 +281,12 @@ block Controller
...
     "Physically fixed minimum position of the return air damper"
     annotation(Dialog(tab="Economizer", group="Commissioning"));
+  parameter Integer cooDemLimLevCon=Buildings.Controls.OBC.ASHRAE.G36_PR1.Types.DemandLimitLevels.cooling0
+    "Cooling demand limit level"
+    annotation (Dialog(tab="Advanced"));
+  parameter Integer heaDemLimLevCon=Buildings.Controls.OBC.ASHRAE.G36_PR1.Types.DemandLimitLevels.heating0
+    "Heating demand limit level"
+    annotation (Dialog(tab="Advanced"));
 
   Buildings.Controls.OBC.CDL.Interfaces.RealInput TOut(
```
kuzha, line 291: Please expose the warUp and cooDow parameters as inputs of the block and update the SZVAV G36 example model.

Buildings/Controls/OBC/ASHRAE/G36_PR1/Generic/SetPoints/ZoneStatus.mo
```diff
@@ -1,380 +1,420 @@
...
+    "Unoccupied cooling setpoint";
   parameter Real bouLim(
     final unit="K",
     final displayUnit="K",
     final quantity="TemperatureDifference",
-    final min=0.5) = 1.1
-    "Value limit to indicate the end of setback or setup mode";
+    final min=0.5) = 1
+    "Threshold of temperature difference for indicating the end of setback or setup mode";
   parameter Boolean have_winSen=false
```
kuzha, line 30: should we change this parameter to have_winSen = have_winSen so that it can be propagated up?

Buildings/Controls/OBC/ASHRAE/G36_PR1/Generic/SetPoints/ZoneStatus.mo
JayHuLBL: @kuzha 
The `hy3` and `hy4` are different from other hysteresis. Here we'd need to check if the temperature is higher/lower than setpoints by 1 degC. 
Other hysteresis check if temperatures is higher/lower than setpoints.

Buildings/Controls/OBC/ASHRAE/G36_PR1/AHUs/SingleZone/VAV/Controller.mo
JayHuLBL: Propagated.

Buildings/Controls/OBC/ASHRAE/G36_PR1/Generic/SetPoints/ZoneStatus.mo
JayHuLBL: `have_winSen` is already a parameter here. We don't need to propagate it here. However, we would need to propagate it in blocks using the `ZoneStatus` to compose. Like the block `ModeAndSetPoints` in single zone AHU package, it has been propagated.

JayHuLBL:
@kuzha 
I have addressed your comments and merged the changes to the issue1502 branch. 

Please double check if it is good.

JayHuLBL:
@kuzha It's ready for your review.

kuzha:
@JayHuLBL 

The issues related to operation modes that we discussed before have been addressed. 
This branch also has the same error as mentioned by @AntoineGautier before (see message below).
I think it should be addressed in the branches that Antoine is working on.

_ERROR: Failed to solve nonlinear system using Newton solver.
  Time: 29051744.94056464
  Tag: simulation.nonlinear[7]_

JayHuLBL:
@mwetter It would be ready for your review after CI tests passed.

JayHuLBL:

Issue #2019, opened 2020-06-26
Minimum outdoor air flow control in Examples.VAVReheat 


AntoineGautier:
Issues with a checked box below are addressed in the branch `issue2019_VAV_MinOA`.

- [x] One issue is related to the default values of the gain (0.05) and integral time (1200s) for the minimum outdoor air flow control in `Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller`, `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Controller`,
`Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Subsequences.Limits`.
This leads to losing the set point tracking as illustrated below (from simulation of `Buildings.Examples.VAVReheat.Guideline36`).
![Guideline36](https://user-images.githubusercontent.com/10965994/85868787-9f537d80-b77f-11ea-89a0-d3c8f7817a63.png)

- [x] Another issue is related to the values of `yMin` and `yMax` in the same set of controllers. According to G36-5.16.5.3 the output of the control loop should be between 0 and 1 whereas the defaults in `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Subsequences.Limits` are -1 and +1 and the parameters `yMinDamLim=0` and `yMaxDamLim=1` in `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller` are not propagated (these parameters are not used at all).

- [x] Another issue is related to the value of `VOut_flow_min` in `ASHRAE2006` which is set to 0.3*m_flow_nominal/1.2 = 1.7 m3/s. A proper sizing with ASHRAE 62.1 leads to the following values, with a total of 0.9 m3/s, so nearly 50% less.
![Screen Shot 2020-07-02 at 7 36 44 PM](https://user-images.githubusercontent.com/10965994/86392227-65143f80-bc9b-11ea-96e4-c01ad1a17b66.png)
This makes the comparison with  `Guideline36` unfair as illustrated below.  

![Guideline36_1](https://user-images.githubusercontent.com/10965994/85869288-56e88f80-b780-11ea-8777-631c3547585e.png)

[...]

Issue #1995, opened 2020-06-06
freeze protection issues in the VAVReheat example models


kuzha:
The two VAVReheat models ASHRAE2006 and Guideline36 in the package Buildings.Examples.VAVReheat have incorrect mixed air temperature TMix. When the AHU system is off and TMix is lower than the freeze protection setpoint at 4C, it invokes the freeze protection mode. 

**ASHRAE2006 model**

_Issue 1: incorrect outdoor air damper control for freeze protection_

![ASHRAE2006](https://user-images.githubusercontent.com/18061609/83931758-5dc94700-a76c-11ea-979b-056e205c1698.png)

The blue curves in the above figure show the results of the current implementation. TMix (shown in the upper plot) is as low as 4C when the AHU system is on. This is caused by the incorrect setpoint and measurement signals for the outdoor air damper control for freeze protection. More specifically, the setpoint and measurement signal for the block yOATFre in the economizer controller Buildings.Examples.VAVReheat.Controls.Economizer should be swapped. In the current implementation, yOATFre tries to cool down the mixed air temperature to the freeze protection temperature. After having swapped these two inputs, TMix increases to around 11C as shown by the red curve. This change also impacts the heating energy use as shown in the second plot.

_Issue 2: simultaneous heating and cooling_

The heating coil and cooling coil can be turned on simultaneously in the current implementation as the two coils use the same fan on signal from the block modeSelector. This is not obvious in winter in Chicago but has been detected using the weather of LA.

**Guideline36**

The mixed air temperature in the Guideline 36 model is incorrect when the system is off. TMix goes to negative in the evening (see the blue curve in the upper plot) and triggers freeze protection.
![G36](https://user-images.githubusercontent.com/18061609/83932495-e4802300-a770-11ea-9e89-0aef653ec90c.png)

This is caused by the exhaust air damper control setpoint which is 1 all the time. In the evening, the outdoor air damper is closed but some amount of outdoor air goes into the room through the exhaust damper. After changing the exhaust damper setpoint using yOutDamPos, TMix goes above freeze protection temperature (see the red curve in the upper plot). Since this happens when the system is off, there is minimal impact on the energy use.

AntoineGautier:
@kuzha I think the first issue you mention (incorrect connection of set-point and measurement signals for `yOATFre` in `Buildings.Examples.VAVReheat.Controls.Economizer`) is not a real one.

[...]

PR #2022, opened 2020-06-30
Issue1502 replaceable thermal zone oa

AntoineGautier:
This PR addresses

1. the issues related to the minimum outdoor air flow rate as reported in #2019 (significant impact expected on simulation),
2. the incorrect connection of set-point and measurement signals for `yOATFre` in `Buildings.Examples.VAVReheat.Controls.Economizer` as reported in #1995 (no impact on simulation).

The impact on annual energy use is being assessed. It will be documented in this PR before merging.

@JayHuLBL Could you please review the first point? And also verify the issue of the default value for `yMin` as described in #2019. I did not modify the models yet because there might be a reason I do not know for the current default.

@kuzha Could you please review the second point?


AntoineGautier:
The new settings for the gain and integral time in G36 AHU controller lead to non convergence issues such as

```
ERROR: Failed to solve nonlinear system using Newton solver.
  Time: 5463672.030250992
  Tag: simulation.nonlinear[5]

  For debugging help refer to the first error message of this type.

  Jacobian inverse norm estimate: 1.64691e+06
  Condition number estimate: 346875
  1-norm of the residual = 0.420988

  Last value of the solution:
    amb.ports[2].m_flow = 0.0950916
    dpRetDuc.port_a.m_flow = -0.0989388
    amb.ports[1].m_flow = -3.94233E-07
```

JayHuLBL:
@AntoineGautier I noticed that you changed the value of zone minimum airflow setpoint, like `conVAVCor.VDisSetMin_flow`, from its default value of `0.15*V_flow_nominal` to the value for ventilation purpose, `(ratOAFlo_P * ratP_A + ratOAFlo_A) * AFloCor / effZ`. Note that the minimum airflow setpoint `VDisSetMin_flow` should be a value provided by the system designer as specified in section 3.1.2.2 of the Guideline 36 official release. Not sure if this change will affect the final results much.

See below my suggestions responding #2019:
1. In block `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Controller`, 
   - the input `uTSup` is output from `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.SetPoints.SupplySignals` which generates the supply temperature loop signal, with heating `[-1, 0]` and cooling `[0, 1]`. The `uTSup` was then used in block `mod` for modulating return and outdoor air damper position.
   - Also note that in both `mod` and `damLim`, there are parameter `uRetDamMin`. But they are different -- one is based on the control of the outdoor airflow rate and the other one is based on the supply temperature control loop signal `uTSup`.
2. In block `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller`,
   - `yMinDamLim=0` and `yMaxDamLim=1` are not used. I believe it intents to be used by `eco.damLim` for the constant `yMin` and `yMax` (since the `yMinDamLim` and `yMaxDamLim` are in tab `economizer` and group `Damper limit`). These two constants `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Subsequences.Limits` originally might be `parameter` but at some point were changed to `constant`.

Suggest to do following:
1. Delete `yMinDamLim=0` and `yMaxDamLim=1` in block `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller`. 
2. In `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Subsequences.Limits`:
    - change the values of constant `yMin` and `yMax`  to be `0` and `1` and remove their annotation for `group` and `tab`.
    - keep the default value of parameter `uRetDamMin` to be 0.5.
3. In `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Controller`,
    - create a new parameter like below and assign it to `damLim.uRetDamLim`
    ```
        parameter Real uRetDamMinLim(
            final min=0,
            final max=1,
            final unit="1") = 0.5
            "Minimum control signal for the return air damper position limit, for specifying limits of the return and outdoor air damper position"
        annotation (Dialog(tab="Commissioning", group="Controller"));
    ```
4. The parameter `uRetDamMinLim` can then be propagated to the controller `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller` and put it the in the tab `Economizer` and group `Damper limits`.
5. The block `Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Reheat.SetPoints.ActiveAirFlow` is to calculate the min/max occupied flow rate, not the outdoor airflow rate. There is no requirement to consider the distribution effectiveness when specifying the min/max flow rate. In addition, the implementation is based on the Guideline36_PR1 section 5.E.3-5, which corresponded to section 5.6.4-5 of the G36 official release.
6. `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.SetPoints.OutdoorAirFlow.Zone` and `Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller` assumes if there is occupancy sensor, it will not just show whether or not the zone is occupied, but also could count how many people in the zone. It would be too advanced now. We could keep it for now as the case study assumes that there is no occupancy sensor so this input will not show anyway. We can refactor it in the process of upgrading to G36 official release version.

AntoineGautier:
@JayHuLBL 

About the zone minimum airflow setpoint: my mistake. I forgot to increase the zone minimum outdoor air flow rate as specified in ASHRAE 62.1-2019 §6.2.4.3.2. _"Zone Minimum Primary Airflow. For each zone, the minimum primary air- flow (Vpz-min) shall be determined in accordance with Equation 6-9. Vpz-min = Voz × 1.5"_. I suggest to adopt this default for Guideline 36. This leads to a percentage between 11% and 19% of the zone maximum primary air flow rate. This also satisfies the requirement from G36 §3.1.2.2 "select Vmin to prevent creating critical zones" as all zones then have the same Zpz value.

About your suggestions

1. Delete `yMinDamLim=0` and `yMaxDamLim=1` in block `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller`.

Agreed and implemented.

2. In `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Subsequences.Limits`:
    - change the values of constant `yMin` and `yMax`  to be `0` and `1` and remove their annotation for `group` and `tab`.
    - keep the default value of parameter `uRetDamMin` to be 0.5.

Agreed and implemented.

3. In `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Controller`,
    - create a new parameter like below and assign it to `damLim.uRetDamLim`
    ```
        parameter Real uRetDamMinLim(
            final min=0,
            final max=1,
            final unit="1") = 0.5
            "Minimum control signal for the return air damper position limit, for specifying limits of the return and outdoor air damper position"
        annotation (Dialog(tab="Commissioning", group="Controller"));
    ```

Agreed and implemented.
I further suggest a different wording and to split the group "Controller" into "Minimum outdoor air control" and "Supply air temperature control", see my commit.

4. The parameter `uRetDamMinLim` can then be propagated to the controller `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller` and put it the in the tab `Economizer` and group `Damper limits`.

I wonder if it is worth exposing this parameter though. From G36 Figure 5.16.6.3 I do not see any specification that this value should be adjustable.

5. The block `Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Reheat.SetPoints.ActiveAirFlow` is to calculate the min/max occupied flow rate, not the outdoor airflow rate. There is no requirement to consider the distribution effectiveness when specifying the min/max flow rate. In addition, the implementation is based on the Guideline36_PR1 section 5.E.3-5, which corresponded to section 5.6.4-5 of the G36 official release.

I agree, sorry for the confusion.

6. `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.SetPoints.OutdoorAirFlow.Zone` and `Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller` assumes if there is occupancy sensor, it will not just show whether or not the zone is occupied, but also could count how many people in the zone. It would be too advanced now. We could keep it for now as the case study assumes that there is no occupancy sensor so this input will not show anyway. We can refactor it in the process of upgrading to G36 official release version.

I cannot see any kind of occupancy sensor returning the number of people in the zone. Also consider G36 §4.2 VAV Terminal Unit with Reheat, where the optional occupancy sensor is listed as DI (not AI). We should open a specific issue I believe, although I agree that this is not a priority.

AntoineGautier:
With the updates from this PR the annual energy use is now
```
.. table:: Heating, cooling, fan and total site HVAC energy, and savings of guideline 36 case versus base case.

   ===================================== ===================================== ====================================== ====================================== =====
   :math:`E_{h} \quad [kWh/(m^2\,a)]`    :math:`E_{c} \quad [kWh/(m^2\,a)]`    :math:`E_{f} \quad [kWh/(m^2\,a)]`     :math:`E_{tot} \quad [kWh/(m^2\,a)]`     [%]
   ===================================== ===================================== ====================================== ====================================== =====
                                   9.577                                 15.46                                  2.093                                  27.13   
                                   7.565                                  10.8                                  1.154                                  19.52  28.0 
   ===================================== ===================================== ====================================== ====================================== =====
```

AntoineGautier:
TODO

- [ ] Add revision notes.

AntoineGautier:
This is closed as it will be addressed through `issue2019_VAV_MinOA` to be merged into the master first.

Issue #2027, opened 2020-07-07
Correct parameter assignment


JayHuLBL:
In `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Controller`, following parameters specification should be fixed:
<pre><code>
parameter Real iniSetSupTem(
    final unit="K",
    final displayUnit="degC",
    final quantity="ThermodynamicTemperature")=<b>supTemSetPoi.maxSet</b>
    "Initial setpoint for supply temperature control"
    annotation (Dialog(tab="Supply air temperature", group="Trim and respond for reseting TSup setpoint"));

  parameter Real maxSetSupTem(
    final unit="K",
    final displayUnit="degC",
    final quantity="ThermodynamicTemperature")=<b>supTemSetPoi.TSupSetMax</b>
    "Maximum setpoint for supply temperature control"
    annotation (Dialog(tab="Supply air temperature", group="Trim and respond for reseting TSup setpoint"));

  parameter Real minSetSupTem(
    final unit="K",
    final displayUnit="degC",
    final quantity="ThermodynamicTemperature")=<b>supTemSetPoi.TSupSetDes</b>
    "Minimum setpoint for supply temperature control"
    annotation (Dialog(tab="Supply air temperature", group="Trim and respond for reseting TSup setpoint"));
</code></pre>

JayHuLBL:
This issue has been addressed through PR #2031.

PR #2030, opened 2020-07-09
Issue2019 vav min oa

### LINKED ISSUE #2019: Minimum outdoor air flow control in Examples.VAVReheat

AntoineGautier:
This closes #2019 
@JayHuLBL As discussed, I included the changes from #2022 into this PR to be merged directly into the master.
I also did two more changes related to the last two items listed in #2019 
See the impacts of these changes at [https://github.com/lbl-srg/modelica-buildings/issues/2019#issuecomment-656115178](https://github.com/lbl-srg/modelica-buildings/issues/2019#issuecomment-656115178)


AntoineGautier:
@JayHuLBL This is ready for your review.
It includes changes in blocks from `Buildings/Controls/OBC/ASHRAE/G36_PR1/AHUs/MultiZone/VAV` (`yMin=0` and `uRetDamMinLim`) so I think this should part of the final sequence we send to Dave Robin.

Buildings/Examples/VAVReheat/Guideline36.mo
```diff
@@ -17,27 +19,37 @@ model Guideline36
   Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller conVAVCor(
     V_flow_nominal=mCor_flow_nominal/1.2,
     AFlo=AFloCor,
-    final samplePeriod=samplePeriod) "Controller for terminal unit corridor"
+    final samplePeriod=samplePeriod,
+    final VDisSetMin_flow=max(1.5*VCorOA_flow_nominal, 0.15*mCor_flow_nominal/1.2))
```
JayHuLBL, line 23: What is the benefit of having this `max()`, rather than the default `0.15*V_flow_nominal` (which is `0.15*mCor_flow_nominal/1.2)`?

JayHuLBL:
@AntoineGautier I made few small changes in commit 97ab605, which addressed issue #2027. Also, in `Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.Economizers.Subsequences.Limits`, I moved the `yMin` and `yMax` to protected section and made them to be parameters.
These changes are necessary for parsing it to json (which can generate list of the value propagation).

Buildings/Examples/VAVReheat/Guideline36.mo
AntoineGautier: This is per ASHRAE Standard 62.1-2019 Eq. (6-9): `VMin = 1.5 * Voz`. However, this equation is part of the simplified procedure (and can lead to values as low as `0.1*V_flow_nominal`) so I chose the compromise of the max with the original default setting which was `0.15*V_flow_nominal` (which offers a generic, not project specific, order of magnitude but can create critical zones).
The requirement from G36 is

> For ASHRAE Standard 62.1 ventilation, select Vmin to prevent creating critical zones.

Taking the max will not create critical zones. It can only lower the zone primary outdoor air fraction Zpz so it will not decrease the system ventilation efficiency.

AntoineGautier:
Closing this one as it was merged in PR #2031 which has now been reviewed.

PR #2039, opened 2020-07-21
Issue1571 wrong weather file

### LINKED ISSUE #1571: spawn hangs if wrong weather file is provided
For `FloorOpenLoop.mo`, use as the weather file in EnergyPlus the `.mos` version of the weather file. Then, EnergyPlus hangs forever. It should return with an error instead.

mwetter:
This closes #1571.

The destructor is now invoked through `ModelicaFormatError`.


mwetter:
The tests fails the same way as they fail on the branch `issue1633_schedulesActuators` due to some memory corruption. I will merge this and then further analysis the cause on `issue1633_schedulesActuators`.

PR #2058, opened 2020-07-29
issue1129_energyPlus_zone branch patched for Windows

DeadParrot:
C sources in Buildings/Resources/C-Sources/EnergyPlus patched to allow building FMUs on Windows, working around the unistd.h and execinfo.h includes.

The printBacktrace function will do nothing on Windows since I'm not aware of a Windows-compiler-portable backtrace approach and didn't take the time to build one.


mwetter:
@DeadParrot : Thanks for the changes.

PR #2021, opened 2020-06-29
Issue1502 replaceable thermal zone staging

AntoineGautier:
@kuzha @JayHuLBL This is our staging branch for the updates we just discussed.
It was created with
```
git checkout issue1502_replaceable_thermal_zone
git pull origin master
```
I opened a draft PR (against `issue1502_replaceable_thermal_zone`) to track the changes.

TODO:

- [x] @JayHuLBL Can you start with merging the mode selection refactoring?

PR #2063, opened 2020-07-30
Merge the corrections to the VAV models

mwetter:
This should be merged after https://github.com/lbl-srg/modelica-buildings/pull/2062 is merged.


mwetter:
Once merged, then #2021 can be closed and the branch `issue1502_replaceable_thermal_zone_staging` deleted

PR #2065, opened 2020-07-31
Issue1951 5 gdhc ets

### LINKED ISSUE #1951: Create template model for 5GDHC energy transfer station
This is addressed through `issue1951_5GDHC_ETS`.

AntoineGautier:
This closes #1951. 

@mwetter This is ready for a review. The developments are in `Applications.DHC.EnergyTransferStations.Combined.Generation5`. Two options for the supervisory controller are proposed.

- `Supervisory1` is the first option which was developed based on previous work (Sidewalk mainly). Basically, heat rejection (resp. cold rejection) is controlled based on the state of the hot tank (resp. cold tank). Overheating of the hot tank happens "spontaneously" when the evaporator load exceeds the condenser load. However, undercooling of the cold tank does not happen simply like that since the chiller maintains the CHWST set point. To trigger cold rejection, a reset of the CHWST is used when the hot tank is in heating demand, to control the HHWST. Note that this reset is anyway needed in winter conditions for instance when the district water is near the CHWST set point. However, the CHWST reset tends to "stick" to low temperatures, as only a decrease of the heating load enables resetting up. So it is anticipated that this approach is detrimental to the system efficiency.

- `Supervisory` is a more straightforward approach where (nearly) everything is controlled from the hot tank side. When the tank overheats, the heat rejection demand increases. When the tank is in heating demand, the cold rejection demand increases (to false load the chiller) and ultimately the CHWST is reset down. On the cold tank side, the CHWST is controlled and limits the cold rejection demand if it compromises the set point tracking (but the CHWST reset signal is not limited for the reason mentioned earlier in winter conditions). It is anticipated that this approach is beneficial to the system efficiency. However, the simulation of the example models show that the CHWST set point tracking is less precise.

I want to further benchmark these two approaches and I am fine with making the two available for now.


JayHuLBL:
@mwetter Following comments have been addressed:
- [x] merged master branch to staging branch `issue1437_district_heating_cooling`, as well as to this being merged branch. So the greater than comparisons now have hysteresis. 
- [x] updated the PID controller after merging master
- [x] In `Buildings.Applications.DHC.EnergyTransferStations.Combined.Generation5.Validation.ChillerOnly`, used `CombiTimeTable` to output loads. And changed simulation time to two days.
- [x] corrected revision note in `Buildings.Applications.DHC.EnergyTransferStations.BaseClasses.CollectorDistributor`
- [x] changed Temperature sensors to enthalpy sensor in class `Buildings.Applications.DHC.Networks.BaseClasses.PartialConnection2Pipe` and `PartialConnection1Pipe`, so the calculation of `Q_flow` becomes more efficient.
- [x] create enumeration type `EnergyTransferStations.Combined.Generation5.Types.HeatExchangerConfiguration` for indicating if the district heat exchanger has two-way valve or pump on the district side.

Buildings/Applications/DHC/EnergyTransferStations/Combined/Generation5/Controls/SideCold.mo
```diff
@@ -67,19 +61,26 @@ model SideCold "Control block for cold side"
...
   Buildings.Controls.OBC.CDL.Continuous.AddParameter addPar1(p=-nSouAmb, k=1)
     "Control signal minus nSouAmb"
     annotation (Placement(transformation(extent={{-150,-110},{-130,-90}})));
   Buildings.Controls.OBC.CDL.Continuous.Min min1
     "Ambient source control signal"
     annotation (Placement(transformation(extent={{-90,-130},{-70,-110}})));
+  Buildings.Controls.OBC.CDL.Continuous.MultiMax mulMax(
+    final nin=nSouAmb)
+    "Maximum value"
+    annotation (Placement(transformation(extent={{90,-130},{110,-110}})));
```
AntoineGautier, line 75: This new block is not needed. The maximum of the elements of the output array from `mapFun[nSouAmb]` will always be at index 1. Hence the previous connect clause `connect(mapFun[1].y, greThr.u)`. Did you make the change because you thought it was incorrect or because you thought it was clearer for the user? In the latter case I will go with your modification.

AntoineGautier:
@mwetter @JayHuLBL There was no hysteresis implemented for the comparison blocks because I used either the built-in wait time parameter from the state graph transition block or a specific true/false hold block. I tend to favor using wait times over hysteresis when possible because the intent is to avoid short cycles. Using a hysteresis instead raises the question of its sizing and the time constant of the short cycle is not obvious in the end and will depend on the system dynamics. So if you agree with this approach I will get rid of the hysteresis where a wait time is already used.

AntoineGautier:
@mwetter I have added a sensor for the difference between two enthalpy flow rates with `tau = 0` by default. 
It will probably be used a lot so your review might be useful.
It is located in `Buildings.Applications.DHC.Networks.BaseClasses.DifferenceEnthalpyFlowRate` for now.
On a side note, I noticed that the enthalpy flow rate sensor has `tau = 1` by default but `Buildings.Fluid.Sensors.UsersGuide` states that 
>  Sensors for quantities that are the product of mass flow rate times a measured fluid property: Such quantities include volumentric flow rate or enthalpy flow rate. For these quantities, sensors are by default configured as steady-state sensor.

Let me know if I should correct the user guide.

mwetter:
@AntoineGautier : Using a wait timer instead of the hysteresis is fine.
Regarding the sensors, I think we should propose to the IBPSA library to change the models to use as default `tau=0` for all "Sensors for quantities that are the product of mass flow rate times a measured fluid property". This seems to make more sense, and then is consistent with the recommendation in the user guide. Can you please open an issue and address this.

AntoineGautier:
TODO

- [ ] Remove hysteresis if a wait timer is present.
- [ ] Convert mass flow rate controlled pumps into speed-controlled

AntoineGautier:
Closed after moving to `Experimental`: a new PR will be open against the master branch directly.
The branch `issue1951_5GDHC_ETS` SHALL NOT be deleted.

PR #2073, opened 2020-08-01
Updated spawn files for use with cmake

### LINKED ISSUE #2072: Change build system for Spawn to using CMake
This issue is to change the build system for the libraries used by Spawn to using `CMake` rather than compiling it from the Modelica simulation environment.
This is in preparation of multi-platform support.

mwetter:
This closes #2072


CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+elseif(CMAKE_SYSTEM_NAME MATCHES Darwin)
+    set(MACOSX 1)
+    message("Mac OS X system detected.")
+else()
+    message(WARNING "OS unknown!")
+endif()
+
+# Set target directories.
+# NOTE: always add quotes to protect spaces in path when setting new variables.
+if (WINDOWS)
```
kbenne, line 45: Consider...

```
if(WIN32)
...
elseif(APPLE)
...
elseif(UNIX)
...
else()
  message(WARNING...)
endif()
```

These are all defined natively by cmake and you can avoid the system name tests above.
https://cmake.org/cmake/help/v3.0/variable/WIN32.html
https://cmake.org/cmake/help/v3.0/variable/UNIX.html
https://cmake.org/cmake/help/v3.0/variable/APPLE.html

A common idiom to test for linux is `if(UNIX AND NOT APPLE)` but here the logic makes that unnecessary.

CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+    if(CMAKE_SYSTEM_PROCESSOR MATCHES ".*64$")
```
kbenne, line 46: I think you might want to use CMAKE_SIZEOF_VOID_P https://cmake.org/cmake/help/v3.12/variable/CMAKE_SIZEOF_VOID_P.html#variable:CMAKE_SIZEOF_VOID_P because this will more accurately tell you the architecture of the target you are making as opposed to just the type of machine you are compiling on. Also I believe CMAKE_SIZEOF_VOID_P is more conventional. Relevant discussion is here http://cmake.3232098.n2.nabble.com/Detecting-32-64-bit-systems-td6019142.html

CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+##################################################################
+cmake_policy(SET CMP0048 NEW)
+
+project( Modelica-Buildings VERSION 5.1.0 )
+cmake_minimum_required(VERSION 3.8)
+
+# find_package(Java REQUIRED)
+# include(UseJava)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -ldl")
```
kbenne, line 20: it is unconventional to hard code `-g`. This should be specified as a config option, `CMAKE_BUILD_TYPE`, at configure time for make based generators, or as a build option at build time for multi config generators such as visual studio.

Linking to dl would be handled in a more cross platform way by linking to ${CMAKE_DL_LIBS} like this https://stackoverflow.com/questions/33678965/need-to-link-cmake-project-to-dl-library

mwetter: Thanks for the feedback, addressed through 40a3f55

PR #2077, opened 2020-08-06
Issue2076 cdl hysteresis

### LINKED ISSUE #2076: add hysteresis to CDL.Greater
This issue is to add optional hysteresis to `CDL.Continuous.Greater*` and `CDL.Continuous.Less*`.

We should also move `CDL.Continuous.GreaterEqual`, `CDL.Continuous.GreaterEqualThreshold` and `CDL.Continuous.LessEqual`, `CDL.Continuous.LessEqualThreshold` to the `Obsolete` package. This will reduce the number of CDL blocks without loosing functionality. The reason is that these work on real-valued signals, and hence due to numerical error or noise, there is no point to distinguish the ">" and ">=".

For comparison, EIKON only has the equivalent of `Greater` and `Less`, and has hysteresis with a default value of 0.

mwetter:
This closes #2076


mwetter:
The tests currently fail due to the initialization of the start value of `y`, which for the block without hysteresis, should not be set, but it must be set for the block with hysteresis.

The solution is likely to use conditionally removable blocks such as in https://github.com/lbl-srg/modelica-buildings/pull/2077/commits/bbe4f580932745665896eb5c7f28017dfdbf43e1

This will need to be integrated for `GreaterThreshold`, `Less` and `LessThreshold`.
This also seems to be a more efficienty implementation as the removable block ensures that only those zero crossing functions appear in the code that are indeed needed based on the value of `h`.

mwetter:
@JayHuLBL : Can you please review this PR. This is ready to merge when all tests pass.

Buildings/Controls/OBC/CDL/Continuous/LessThreshold.mo
```diff
@@ -23,32 +117,64 @@ equation
...
                else {235,235,235}),
           fillColor=DynamicSelect({235,235,235}, if y then {0,255,0}
                else {235,235,235}),
           fillPattern=FillPattern.Solid),
-        Text(
-          extent={{-150,-140},{150,-110}},
-          lineColor={0,0,0},
-          textString="%threshold"),
+        Line(
+          points={{-16,10},{2,2},{-16,-8}},
```
JayHuLBL, line 127: It should be `<`, not `>`.

Buildings/Controls/OBC/CDL/Continuous/Greater.mo
```diff
@@ -1,22 +1,117 @@
...
     annotation (Placement(transformation(extent={{-140,-100},{-100,-60}})));
 
-  Interfaces.BooleanOutput y "Connector of Boolean output signal"
+  Interfaces.BooleanOutput y "Output y"
     annotation (Placement(transformation(extent={{100,-20},{140,20}})));
 
-equation
-  y = u1 > u2;
+protected
+  final parameter Boolean haveHysteresis = h >= 1E-10
```
JayHuLBL, line 20: follow our convention `have_Hysteresis`?

Buildings/Fluid/CHPs/BaseClasses/Validation/CoolDown.mo
```diff
@@ -49,9 +49,11 @@ protected
 protected
   Buildings.Controls.OBC.CDL.Logical.Timer timer
     annotation (Placement(transformation(extent={{0,0},{20,20}})));
-  Buildings.Controls.OBC.CDL.Continuous.GreaterEqualThreshold timeDel(
-    final threshold=per.timeDelayCool)
-    "Check if the time of  plant in cool-down mode has been longer than the specified delay time"
+  Buildings.Controls.OBC.CDL.Continuous.GreaterThreshold timeDel(
+    final t=per.timeDelayCool,
+    final h=0)
+    "Check if the time of  plant in cool-down mode has been longer than the
```
JayHuLBL, line 55: don't change line

mwetter:
@milicag : I think you have been waiting for this PR to be merged before you merge the master to your chiller plant control implementation.

PR #2104, opened 2020-08-27
Issue2101 timer

### LINKED ISSUE #2101: separate CDL.Logical.Timer
This issue is to update the timer:
- [ ] separate `CDL.Logical.Timer` to two timer: `Timer` and `TimerAccumulating`
- [ ] add boolean output to check if the time is greater than threshold

JayHuLBL:
- [x] created `TimerAccumulating`
- [x] added parameter `t` as the threshold time in the timers
- [x] created boolean output `pasThr` to indicate if the time is greater than threshold time
- [x] moved obsolete timer to `Obsolete` package
- [x] updated the classes that instantiate the obsolete timer

This closes #2101.


mwetter:
@JayHuLBL : The model behaves incorrect if the simulation does not start at zero. It also has state events, which can be avoided if the model is formulated differently. We need to make sure we don't have state events if they can be avoided.

I fixed this in https://github.com/lbl-srg/modelica-buildings/pull/2104/commits/ef6b969e03f2904052536bb5d5b230d81212615a

I also added the value of the output as is done for other blocks, this is done in https://github.com/lbl-srg/modelica-buildings/pull/2104/commits/c6950cf551cabee8e5f37bf044d129d888a65762

I also corrected the comments for the accumulating timer, it looks like the top level comment and the comment for `y` was copied from the other timer or left from the old implementation, but these are not correct for the accumulating timer.

The regression tests in `CDL.Logical` all passed, but I committed minor changes for the optimal startup.

Can you please add regression tests for the case where the simulation does not start at zero, say at -10, and inspect carefully all results of the different versions of the block that are being tested.

Buildings/package.mo
```diff
@@ -253,7 +253,13 @@ to <b style=\"color:blue\">existing</b> libraries:
                        This implementation allows to reset the output of the controller
                        to a parameter value. (Resetting it to an input was never used and is now removed for simplicity.)<br/>
                        This is for <a href=\"https://github.com/lbl-srg/modelica-buildings/issues/2056\">issue 2056</a>.</td>
-</tr>                       
+</tr>
+<tr><td valign=\"top\">Buildings.Controls.OBC.CDL.Logical.TimerAccumulating
+    </td>
+    <td valign=\"top\">New timer that accumulates time. The output will be reset to zero when the reset input
+                       becomes true.<br/>
```
mwetter, line 260: Use "<code>true</code>"

PR #2115, opened 2020-08-31
Boolean pulse - fixed special initiation condition behavior #2110

### LINKED ISSUE #2110: Correct BooleanPulse block to avoid wrong result
Block `Modelica.Blocks.Sources.BooleanPulse` and `Buildings.Controls.OBC.CDL.Logical.Sources.Pulse` would give wrong results. 

For example, by setting its parameters as `width=0.5, period=4, startTime=-11`, 

1. run simulation with settings as: `experiment(StartTime=-3, StopTime=5, Tolerance=1e-06, __Dymola_Algorithm="Dassl")`, gives output:
image

2. run simulation with settings as: `experiment(StartTime=-2, StopTime=5, Tolerance=1e-06, __Dymola_Algorithm="Dassl")`, gives output:
image

The simulation results should not shift along with the different simulation start times, in which, the second one is not correct.

milicag:
Closes #2110.


milicag:
@mwetter Sure! I also want to look at the continuous pulse as already had a nearly identical condition implemented to see if we've covered a good range of behaviors with both now.

milicag:
@mwetter Please see the edits as requested. I have two other comments:
* Should we notify MSL of this, also as their continuous pulse does not have the omission that I've just corrected in our lib? I would call the issue resolved here (#2110) a bug
* I see that the continuous pulse has a parameter to specify the number of periods. Just something to consider for our boolean implementation as well, if there is ever a need.

mwetter (approved):

mwetter:
@milicag : can you please report this to the MSL implementation if it has the same bug.

PR #2127, opened 2020-09-07
Optimal start examples

kuzha:
@mwetter 

Please review the example models for the block OptimalStart for a single-zone building and VAV system.
Two models are created: one with a conventional controller from the Buildings.Air package, the other with a Guideline 36 controller.

The optimal start time is sensitive to the PI parameters of the VAV controllers because those parameters determine how fast the zone reaches or passes the zone set point temperature. This is especially true for the conventional controller. The G36 controller in general has smoother transient behavior when the optimal start switches the system on. In result, the G36 controller has smoother optimal start time than the conventional controller.

I also corrected a tiny bug in the model Buildings.Air.Systems.SingleZone.VAV.BaseClasses.ControllerEconomizer.mo.


Buildings/Air/Systems/SingleZone/VAV/Examples/OptimalStart/BaseClasses/package.mo
```diff
@@ -0,0 +1,10 @@
+within Buildings.Air.Systems.SingleZone.VAV.Examples.OptimalStart;
+package BaseClasses "Base classes of models in the package OptimalStart"
+extends Modelica.Icons.BasesPackage;
+
+  annotation (preferredView="info", Documentation(info="<html>
+<p>
+This package contains base classes used by the example models of OptimalStart.
```
mwetter, line 7: Add the full link in the form ``<a href=\"modelica://Buildings....``. Then a hyperlink will be added, and if we rename the class, it will get updated too.

PR #2136, opened 2020-09-14
Issue2125 additional sources

### LINKED ISSUE #2125: add blocks to CDL.*.Sources
This issue is to make the package `CDL.*.Sources` more uniform. Blocks that are applicable for all data types should be available for all these data types, and have the same parameters.

Specifically, the following blocks should be present (x means they are already present)
```
Continuous         Logical   Integer   Block
   x                 x           x     Constant
   x                 x                 Pulse (with a parameter nPeriod,
 rename nperiod                         renamed from nperiod to use camel case)
   x                                  TimeTable
```
Note that for `Logical.Pulse` the parameter `nPeriod` should be added.

milicag:
Closes #2125


Buildings/package.mo
```diff
@@ -290,6 +290,11 @@ have been <b style=\"color:blue\">improved</b> in a
     <td valign=\"top\">Added option to specify a hysteresis, which by default is set to <i>0</i>.
```
mwetter, line 290: @milicag : Please also add the list of new blocks to the release notes.

mwetter (changes requested):
Please add a list with the new blocks to the release notes.

milicag:
@mwetter Sure. Please let me know if you have a clue about what might be failing the tests, local tests are passing and all new validation tests should generate a results file with a unique name based on the `mos` scripts.

JayHuLBL:
@milicag 
- For this error (https://travis-ci.com/github/lbl-srg/modelica-buildings/jobs/388940359): it's because that the .mos scripts `CDL.Integers.Sources.Validation.TimeTable.mos` and `CDL.Integers.Sources.Validation.TimeTableNegativeStartTime.mos` runs the same class:
```
simulateModel("Buildings.Controls.OBC.CDL.Integers.Sources.Validation.TimeTable",
```
This is also why in the package `/Resources/Scripts/OpenModelica/compareVars`, you can see `Buildings.Controls.OBC.CDL.Integers.Sources.Validation.TimeTable.mos`, but cannot see `Buildings.Controls.OBC.CDL.Integers.Sources.Validation.TimeTableNegativeStartTime.mos`


- For this error (https://travis-ci.com/github/lbl-srg/modelica-buildings/jobs/388940357): you'd need to update the reference results after running the unit tests (with dymola) for the package `Buildings.Controls.OBC.CDL`. As in my local machine, when run the unit test for the package, I did see following:
image

milicag:
@JayHuLBL Thanks, much appreciated. I wonder why my local tests for the `CDL` package are passing then.

JayHuLBL:
@milicag Best would be to check your dymola version and may also update your BuildingsPy. My local dymola version is dymola 2021, which I think it's the same as in travis.

milicag:
@JayHuLBL Thank you for your help on debugging.
@mwetter This is ready for your review.

Buildings/Controls/OBC/CDL/Integers/Sources/TimeTable.mo
```diff
@@ -0,0 +1,207 @@
+within Buildings.Controls.OBC.CDL.Integers.Sources;
+block TimeTable "Table look-up with respect to time with constant segments"
+
+  parameter Real table[:,:]
+    "Table matrix with time as a first table column (in seconds, unless timeScale is not 1) and Integers in all other columns";
+  parameter CDL.Types.Extrapolation extrapolation=CDL.Types.Extrapolation.HoldLastPoint
+    "Extrapolation of data outside the definition range";
```
mwetter, line 7: Remove extrapolation option, see comment.

Buildings/Controls/OBC/CDL/Integers/Sources/TimeTable.mo
```diff
@@ -0,0 +1,207 @@
...
+  parameter Integer offset[:]=fill(0, nout) "Offsets of output signals";
+  parameter Modelica.SIunits.Time timeScale=1
+    "Time scale of first table column. Set to 3600 if time in table is in hours";
```
mwetter, line 10: `timeScale` must have no units. The first table column has seconds, so scale is just a non-dimensional factor.

Buildings/Controls/OBC/CDL/Integers/Sources/TimeTable.mo
```diff
@@ -0,0 +1,207 @@
...
+    "Value comparisson"
+    annotation (Placement(transformation(extent={{-40,-80},{-20,-60}})));
+  Utilities.Assert assMes(message="Scheduled values are not all of type Integer")
+    "Assert all scheduled values are integer values"
+    annotation (Placement(transformation(extent={{60,-80},{80,-60}})));
+  Logical.MultiOr mulOr(
+    final nu=nout) "Multi or"
+    annotation (Placement(transformation(extent={{0,-80},{20,-60}})));
+  Logical.Not not1 "Logical not"
+    annotation (Placement(transformation(extent={{30,-80},{50,-60}})));
```
mwetter, line 44: Simply make a loop in the initial equation section, this will be more efficient.

Buildings/Controls/OBC/CDL/Integers/Sources/TimeTable.mo
```diff
@@ -0,0 +1,207 @@
...
+If <code>extrapolation === CDL.Types.Periodic</code>, then the above example
+would give a schedule with periodicity of one day. The simulation can start at any time,
+whether it is a multiple of a day or not, and positive or negative.
+</p>
+
+<p>
+The value of the parameter <code>offset</code> is added to the tabulated
+values.
+The parameters <code>timeScale</code> is used to scale the first column
+of the table. For example, set <code>timeScale = 3600</code> if the first
```
mwetter, line 157: Time scale has been explained above already. Make sure you only explain it once.

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
```diff
@@ -0,0 +1,235 @@
...
+    final nu=nout) "Multi and"
+    annotation (Placement(transformation(extent={{50,60},{70,80}})));
+  MultiOr mulOr1(
+    final nu=nout) "Multi or"
+    annotation (Placement(transformation(extent={{20,-100},{40,-80}})));
+  And and2
+    "Logical and"
+    annotation (Placement(transformation(extent={{60,20},{80,40}})));
+  Buildings.Controls.OBC.CDL.Conversions.RealToInteger realToInteger[nout]
+    "Converts scheduled values to integer"
```
mwetter, line 44: Move the test for the data type to `initial equation` because this test only depends on parameter values.

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
```diff
@@ -0,0 +1,235 @@
...
+    annotation (Line(points={{2,-90},{18,-90},{18,-90}}, color={255,0,255}));
+  connect(mulOr1.y, not1.u)
+    annotation (Line(points={{42,-90},{58,-90}}, color={255,0,255}));
+  connect(not1.y, and2.u2) annotation (Line(points={{82,-90},{90,-90},{90,-20},{
+          50,-20},{50,22},{58,22}}, color={255,0,255}));
+  connect(mulAnd.y, and2.u1) annotation (Line(points={{72,70},{80,70},{80,50},{50,
+          50},{50,30},{58,30}}, color={255,0,255}));
+  connect(and2.y, assMes1.u) annotation (Line(points={{82,30},{90,30},{90,70},{98,
+          70}}, color={255,0,255}));
+  annotation (
```
mwetter, line 126: That looks very complicated. Why not just use the output of the table with a `Less` block to convert to a boolean?

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
```diff
@@ -0,0 +1,235 @@
+within Buildings.Controls.OBC.CDL.Logical.Sources;
+block TimeTable
+  "Table look-up with respect to time with constant segments"
+
+  parameter Real table[:,:]
+  "Table matrix with time as a first column (in seconds, unless timeScale is not 1) and Booleans in all other columns";
+  parameter CDL.Types.Extrapolation extrapolation=CDL.Types.Extrapolation.HoldLastPoint
+    "Extrapolation of data outside the definition range";
+  parameter Modelica.SIunits.Time timeScale=1
+    "Time scale of first table column. Set to 3600 if time in table is in hours";
```
mwetter, line 10: `timeScale` must be unitless.

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
```diff
@@ -0,0 +1,235 @@
...
+  Not not1
+    "Logical not"
+    annotation (Placement(transformation(extent={{60,-100},{80,-80}})));
+
+initial equation
+  t0=time;
+
+equation
+  assert(n > 0, "No table values defined.");
+  assert(extrapolation <> CDL.Types.Extrapolation.LastTwoPoints, "Unsuitable extrapolation setting.");
```
mwetter, line 86: Move to initial equation section

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
```diff
@@ -0,0 +1,235 @@
...
+Any number of columns can be specified.
+</p>
+<p>
+The values in all columns apart from the first column must equal either <code>0</code> or <code>1</code>, 
+to represent <code>false</code> or <code>true</code>, respectively, otherwise a warning is issued.
+</p>
+<p>
+The parameter <code>smoothness</code> determines how the table values
+are interpolated. The following setting is implemented:
+</p>
```
mwetter, line 158: There is no smoothness parameter (and it makes not sense to use smoothness for a boolean signal).

milicag: I can protect it, yet we need to define how the values between the defined points in time are determined.

mwetter (changes requested):
For the tables with Integer and Boolean, we should simply set them to be always periodic. Otherwise we need to introduce yet another `CDL.Types` because the current types allow extrapolation based on the derivative which does not make sense for these data types. Without such an extrapolation parameter, it would not be clear if the value after `nPeriods` would be the value the output had before `startTime`, or wether a hold is used. Hence, easiest to understand the blocks is to simply set them to be periodic.

For the `TimeTable` for `Integer` and `Boolean`, please make the assertion for the data type in the `initial equation` section by looping over the table. This will be more efficient than doing the type conversion and checks during the time stepping.

Please carefully review the info section again. For example, in `Buildings.Controls.OBC.CDL.Integers.Sources.TimeTable` it is explained what `smoothness` means, but there is, correctly, no such parameter.

Please see inline comments for more changes.

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
milicag: Would that be `Less` than 0.5 that you are suggesting? How does one make sure that the user has not entered values below zero, very large values, values between 0 and 1?

Buildings/Controls/OBC/CDL/Integers/Sources/TimeTable.mo
milicag: Which comment are you referring to? If I remove extrapolation, how should the values outside of the time range be handled.

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
mwetter: You could set it directly in the instance declaration or as a protected parameter; but the info section should not mention it. Typical users may just read the info section, so explaining smoothness will be confusing if there is no such option (and there should not be such an option for the user).

Buildings/Controls/OBC/CDL/Integers/Sources/TimeTable.mo
milicag: I see, you want it to be periodic. That's probably good for now, yet, there is a way to implement it without looking a the derivative, you'd say equal to first point when time less than firs point in time, and equivalent for after the last point.

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
milicag: That sounds good to me.

milicag:
@mwetter It is still not clear to me why we would not also allow first and last point hold as extrapolation. Maybe I'm missing something here.

mwetter:
@milicag Regarding https://github.com/lbl-srg/modelica-buildings/pull/2136#issuecomment-701159065 I don't see a strong use case. If there is one, we could add this feature through a boolean parameter, but not through the enumeration that is currently used because the enumeration allows to set
`LastTwoPoints  "Extrapolate by using the derivative at the first/last table points outside of the table scope"`
which is not applicable for integers and booleans.

milicag:
@mwetter We have an assertion in the equation section for the unsuitable extrapolation `CDL.Types.Extrapolation.LastTwoPoints`. This is based on the MSL implementation. Also the assertion might better be placed in the initial equation section. 

To me asserting the extrapolation setting is suitable is similar to asserting that the table scheduled entries, that is all values in column other than the first one, are of type `Integer` or of type `Boolean`. We discussed previously that the user is expected to in fact provide integers, rather than to have us perform rounding, in case of integers, and value comparisons, in case of booleans. How is the extrapolation type limitation different?

mwetter:
@milicag : The extrapolation type is different in that we ask the user to explicitly, through the enumeration, whether he/she wants to use "Extrapolate by using the derivative at the first/last table points outside of the table scope", just to then tell through an assertion that this option is not valid. That makes no sense. Also, derivatives are not defined for integers or booleans, so we could not even explain what we mean by this option.  Let's make sure the blocks are simple and easy to understand and use for the user.

milicag:
@mwetter If you take a closer look, my implementation, as well as MSL implementation, just says that if you are not using `Periodic`, whichever other option you choose the extrapolation will be set to hold the last points. Also, we are making it clear in our documentation. The only issue is the fall down menu accessible to the user to set the parameter, since it will list an option `LastTwoPoints`, that if selected the code will convert to `HoldLastPoint`.

Is the development of a new `Type` really the only option to offer only a subset of types to a user through that fall down menu? One would think there is a way in Modelica to impose that limitation using the existing `Type`. I agree we could just convert to a boolean flag and say `is_periodic = True or False`, if False then `HoldLastPoint`.

mwetter:
@milicag Try to make it as simple as possible for the user, and don't add complexity for a use case that may not exist.
There is no way to constrain an enumeration in Modelica or CDL. Providing an option that always triggers an assertion when selected is bad design as it confuses the user and creates a situation where we have no rational other than "it was easier to program" which is not of interest to the user. I would not know how to explain why we selected that option to be available. Let's keep it simple.

milicag:
@mwetter ok, then I'll have the periodic be the only option for now and we can add as needed. Thanks!

Buildings/Controls/OBC/CDL/Logical/Sources/TimeTable.mo
milicag: Sure, and I removed the second line as we now only use periodic. This applies to both integer and boolean flavors of the block.

Buildings/Controls/OBC/CDL/Integers/Sources/TimeTable.mo
milicag: I updated for the continuous time table block as well.

milicag:
> For the tables with Integer and Boolean, we should simply set them to be always periodic. Otherwise we need to introduce yet another `CDL.Types` because the current types allow extrapolation based on the derivative which does not make sense for these data types. Without such an extrapolation parameter, it would not be clear if the value after `nPeriods` would be the value the output had before `startTime`, or wether a hold is used. Hence, easiest to understand the blocks is to simply set them to be periodic.
> 
> For the `TimeTable` for `Integer` and `Boolean`, please make the assertion for the data type in the `initial equation` section by looping over the table. This will be more efficient than doing the type conversion and checks during the time stepping.
> 
> Please carefully review the info section again. For example, in `Buildings.Controls.OBC.CDL.Integers.Sources.TimeTable` it is explained what `smoothness` means, but there is, correctly, no such parameter.
> 
> Please see inline comments for more changes.

@mwetter To me it seems like we could omit mentioning in the documentation both interpolation and extrapolation parameters and just in plain text say that the latest value is held until the new point in the table and that the table is repeated periodically. Does that sound good to you?

mwetter:
@milicag Please propose in the Modelica file wording that you think is clearest to the user and then I can review it.

milicag:

mwetter:
Todo
- [ ] For `Logical.Sources.Validation.TimeTable` check if the 9 zero crossing functions can be avoided by simply sampling the model (use of the continuous table and convert back to Boolean seems too complicated. Maybe we can simply use something like
```modelica
   when {sampleTrigger(startTime[k], timeSpan) for k in 1:nCol-1} then // the last column is just used for computing timeSpan
     y[i] = ...
   end when;
```
- [ ] Review the models other than `TimeTable`

PR #2150, opened 2020-09-21
Refactored TimeSeries model in staging branch issue1437 to allow heating-only or cooling-only modeling

### LINKED ISSUE #2147: BuildingTimeSeries model in staging branch issue1437 will fail to run if no heating or no cooling
When instancing the class `Buildings.Applications.DHC.Loads.Examples.BaseClasses.BuildingTimeSeries` and setting `have_watHea = false` or `have_watCoo`, it will fail to check with following error:
```
Check of Buildings.Applications.DHC.Loads.Examples.CouplingTimeSeries:

The problem is structurally singular.

Attempting to further localize singularity.

The model Buildings.Applications.DHC.Loads.Examples.CouplingTimeSeries component bui.disFloHea is structurally singular.
when assuming inputs to be known
and the most generic outside couplings to all the flow variables of its connectors.

Some connectors are not connected from the outside:
bui.disFloHea.mReqTot_flow
bui.disFloHea.port_a
bui.disFloHea.port_b
bui.disFloHea.QActTot_flow

There was an attempt to connect to port_a of bui.disFloHea.senTSup connector, as an outside connector - but the other part was disabled. That is a likely cause of a singularity

ERRORS have been issued.
```

JayHuLBL:
This is for issue #2147. 

- [x] corrected instantiation of conditional blocks so when `have_watHea` or `have_watCoo` becomes false, disabling corresponded blocks.


AntoineGautier (changes requested):
I agree with the modifications made but I would like to further discuss the modeling intent before merging, see https://github.com/lbl-srg/modelica-buildings/issues/2147#issuecomment-705192554

AntoineGautier (approved):

AntoineGautier:
@JayHuLBL I think we can merge this: does that seem good to you? But I would not close the issue #2147 until we discuss the shortcomings described in https://github.com/lbl-srg/modelica-buildings/issues/2147#issuecomment-705192554 with the rest of the team.

JayHuLBL:
@AntoineGautier I agree with the changes and keep the issue open.

Issue #2082, opened 2020-08-14
FMU_EP_VERBOSITY does not work correctly


kbenne:

mwetter:
Please make a PR against `issue2082_ep_verbosity`

mwetter:
I will close this as the logging has been completely redesign and no longer uses `FMU_EP_VERBOSITY`. The new code is on the master.

Issue #2132, opened 2020-09-11
add Windows binaries for Spawn


mwetter:
This issue is to add Windows binaries to Spawn of EnergyPlus.

mwetter:
@kbenne : 
I get an error when unloading the dll on Windows. The code on this branch works on Linux with
```
jm_ipython.sh jmodelica.py  Buildings.ThermalZones.EnergyPlus.Validation.ThermalZone.OneZone
```
and it also works in Dymola on Linux.

When I run this model in Dymola on Windows, I get
```
Unkown error
fmi2Terminate returned with status Error for building OneZone.building.
```
The place where the error is triggered is
https://github.com/lbl-srg/modelica-buildings/blob/issue2132_spawn_windows/Buildings/Resources/src/ThermalZones/EnergyPlus/C-Sources/EnergyPlusFMU.c#L327
which returns a non OK status.

I committed the code, except for the Windows E+ binaries. You can install them by running (on any OS)
```
cd Buildings/Resources/src/ThermalZones/EnergyPlus
python install.sh
```
Can you please check why I don't get OK when calling `fmi2_import_terminate`

kbenne:
Do you know where the message "Unkown error" originates? The typo should make it easy to find unless it is in third party code. I grepped for it in my code with no result.

Given that terminate normally happens at the end of the simulation can you tell if the simulation was able to startup and begin an EnergyPlus simulation? You should have output files even if they are incomplete. And before that, an EnergyPlus FMU should have been generated.

mwetter:
@kbenne : The error message is "Unknown error" (my typo in this issue only). I do not have code that has this message, and neither does the fmi-library, so it either comes from your code or from Windows.

The FMU is generated, but I see no output files of EnergyPlus or no sign that it started.

mwetter:
@kbenne : The line that I added in https://github.com/lbl-srg/modelica-buildings/commit/f73cd4e8e0ee8117bcfa282eb9cd3e5540b7aa1c is not reached on Windows. The call above seems to issue "Unknown error". Afterwards, Dymola calls the destructor function, e.g., I see "Entered FMUBuildingFree" in `dslog.txt`.

Does this work in your CI testing on Windows?

kbenne:
@mwetter  https://github.com/NREL/Spawn/commit/b630852af9ec7338c47dfbedcd25aaad32c87469 yields successful simulation on windows. you will need a new epfmi.dll

[...]

PR #2153, opened 2020-09-22
Fix bug related to FMU_EP_VERBOSITY

kbenne:
ref #lbl-srg/modelica-buildings#2082


mwetter:
@kbenne : Thanks. This code has actually been completely redesigned in https://github.com/lbl-srg/modelica-buildings/issues/2132. I will close this PR without merging to avoid a conflict with https://github.com/lbl-srg/modelica-buildings/tree/issue2132_spawn_windows

Can you please look into https://github.com/lbl-srg/modelica-buildings/issues/2132#issuecomment-693673707

I plan to get back to https://github.com/lbl-srg/modelica-buildings/issues/2132 next week when the conferences and workshop are over.

PR #2155, opened 2020-09-29
Update CDL.Psychrometrics package

### LINKED ISSUE #2139: Refactor Buildings.Controls.OBC.CDL.Psychrometrics
The blocks
`Buildings.Controls.OBC.CDL.Psychrometrics.TDewPoi_TDryBulPhi`
`Buildings.Controls.OBC.CDL.Psychrometrics.TWetBul_TDryBulPhi`
take as an input `p`, which is not used. This can be removed.

The block
`Buildings.Controls.OBC.CDL.Psychrometrics.h_TDryBulPhi`
takes as an input p, this should be removed and instead an parameter `pAtm = 101325` should be used, with annotation `tab=Advanced`.

To distinguish them from the blocks with the same name in `Buildings.Utilities`, we may want to move them to `Obsolete` and rename them to `DewPoint_TDryBulPhi`, `WetBulb_TDryBulPhi` and `SpecificEnthalpy_TDryBulPhi`.

mwetter:
This closes #2139

@JayHuLBL : Please verify 
https://github.com/lbl-srg/modelica-buildings/blob/issue2139_CDL_Psychrometrics/Buildings/Controls/OBC/CDL/Psychrometrics/SpecificEnthalpy_TDryBulPhi.mo#L36
which is from your implementation.

I think this should be
```modelica
h = (1-XiDryBul)*1006*TDryBul_degC + XiDryBul*(2501014.5+1860*TDryBul_degC);
```
Please correct this line if you agree as part of the review.


JayHuLBL:
@mwetter In 2009 ASHRAE handbook fundamental (page 1.8 and 1.9):

image

and
 
image

where the humidity ratio W is defined as `the ratio of the mass of water vapor to the mass of dry air in the sample`, with unit `kg_w/kg_da`.

I think the existing calculation is correct:
```
XiDryBul = 0.6219647130774989*p_w/(p-p_w);
h = 1006*TDryBul_degC + XiDryBul*(2501014.5+1860*TDryBul_degC);
```

mwetter:
@JayHuLBL : I see, the equation is correct because `XiDryBul = 0.6219647130774989*p_w/(p-p_w);` is in kg/kg dry air. Because in Modelica, Xi (in the connectors and the media definitions) is in kg/kg total air, I think it would be clearer to change
```modelica
  Modelica.SIunits.MassFraction XiDryBul(nominal=0.01)
    "Water vapor mass fraction at dry bulb state";
```
to 
```modelica
  Real w(final unit="1", nominal=0.01)
    "Water vapor mass fraction in kg per kg dry air";
```

PR #2173, opened 2020-10-09
Removed default value of maximum/minimum airflow setpoints of the terminal box control

### LINKED ISSUE #2172: Propagate maximum and minimum airflow setpoints of the VAV terminal unit controller
In class `Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller`, the parameter `V_flow_nominal` is only used for set default values of the design airflow setpoints `VDisCooSetMax_flow`, `VDisSetMin_flow`, `VDisHeaSetMax_flow` and `VDisConMin_flow`. These setpoints should not have default values and the parameter `V_flow_nominal` is not needed.

This issue is to
- [ ] delete `V_flow_nominal`,
- [ ] remove default values of the airflow setpoints,
- [ ] update parameter values of terminal controller in class `Buildings.Examples.VAVReheat.Guideline36`, so the `VDisCooSetMax_flow` is zone nominal flow, `VDisHeaSetMax_flow` will be 0.3 of zone nominal flow.

JayHuLBL:
This closes #2172.


Buildings/Controls/OBC/ASHRAE/G36_PR1/TerminalUnits/Controller.mo
```diff
@@ -5,10 +5,7 @@ block Controller "Controller for room VAV box"
     final unit="s",
     final quantity="Time")
     "Sample period of component, set to the same value as the trim and respond that process yPreSetReq";
-  parameter Real V_flow_nominal(
```
mwetter, line 8: Removing the parameter `V_flow_nominal` is a non-backward compatible change. This needs to be listed in the release notes of the top-level `package.mo` file under the non-backward compatible changes, and an entry must be made in the conversion script.

Buildings/Controls/OBC/ASHRAE/G36_PR1/TerminalUnits/Controller.mo
```diff
@@ -156,10 +153,7 @@ block Controller "Controller for room VAV box"
     final quantity="TemperatureDifference")=11
     "Zone maximum discharge air temperature above heating setpoint"
     annotation (Dialog(tab="Damper and valve", group="Parameters"));
-  parameter Real TDisMin(
-    final unit="K",
-    final displayUnit="degC",
-    final quantity="ThermodynamicTemperature")=283.15
+  parameter Real TDisMin=283.15
```
mwetter, line 156: Is there a reason for deleting the units?

mwetter (changes requested):
See inline comments.

JayHuLBL:
@AntoineGautier It will be ready for your review after the tests passed.

JayHuLBL:
@AntoineGautier It's ready for your review.

Buildings/Resources/Scripts/Dymola/ConvertBuildings_from_7_to_8.0.0.mos
```diff
@@ -74,3 +74,8 @@ convertClass(
 convertClass(
     "Buildings.Controls.OBC.CDL.Types.Reset",
     "Buildings.Obsolete.Controls.OBC.CDL.Types.Reset");
+
+// Conversion for https://github.com/lbl-srg/modelica-buildings/issues/2172
+// remove parameter V_flow_nominal
+convertModifiers("Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller",
+                {"V_flow_nominal"}, fill("",0), true);
```
AntoineGautier, line 84: This command will delete the parameter binding for `V_flow_nominal`. However, the converted model will still fail to simulate as the air flow set point parameters are bound by default to `V_flow_nominal` in the implementation on the master. Additionally, the user will lose the information about the original value that was assigned to `V_flow_nominal`, which is more problematic.
A call like the following would remove `V_flow_nominal` and assign the current default values to the other set points.
```
convertModifiers("Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller",
    {"V_flow_nominal", "VDisConMin_flow"},
    {"VDisConMin_flow=0.1*%V_flow_nominal%"},
    true);
```
However, I cannot see a call to `convertModifiers` that would preserve a possible overriding from the user, for instance if he changed the default to `VDisConMin_flow = 0.2 * V_flow_nominal` or even a literal value.

Buildings/package.mo
```diff
@@ -367,6 +374,16 @@ have been <b style=\"color:blue\">improved</b> in a
...
 </tr>
+<tr><td colspan=\"2\"><b>Buildings.Controls.OBC.ASHRAE</b>
+    </td>
+</tr>
+<tr><td valign=\"top\">Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller
+    </td>
+    <td valign=\"top\">Deleted <code>V_flow_nominal</code> and removed default values of maximum heating, maximum cooling, minimum
+                       cooling and controllable minimum discharge airflow setpoints.<br/>
+                       This is for <a href=\"https://github.com/lbl-srg/modelica-buildings/issues/2172\">issue 2172</a>.<br/>
+                       For Dymola, a conversion script makes this change.</td>
```
AntoineGautier, line 410: See my comment above. We should precise the limitations of the script.

AntoineGautier (changes requested):
I agree with the changes made to the models. However, the conversion script and/or the description of what the conversion script exactly does should be changed I believe.

Buildings/Controls/OBC/ASHRAE/G36_PR1/TerminalUnits/Controller.mo
```diff
@@ -122,22 +119,22 @@ block Controller "Controller for room VAV box"
...
     final unit="m3/s",
-    final quantity="VolumeFlowRate")=V_flow_nominal
+    final quantity="VolumeFlowRate")
     "Zone maximum heating airflow setpoint"
     annotation (Dialog(tab="Airflow setpoint", group="Nominal conditions"));
   parameter Real VDisConMin_flow(
     final unit="m3/s",
-    final quantity="VolumeFlowRate")=0.1*V_flow_nominal
+    final quantity="VolumeFlowRate")
     "VAV box controllable minimum"
```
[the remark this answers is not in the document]
JayHuLBL: The sequence was implemented according to the G36 public release 1. As see in the table below from G36_PR1 (section 5.E.5), the required design inputs are maximum cooling, maximum heating and the occupied min flow. In order to find the occupied minimum airflow setpoint (V_min^*), which was implemented (section 5.E.4 of G36 PR1) in block `Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Reheat.SetPoints.ActiveAirFlow`, it requires zone minimum airflow and controllable minimum flow, no heating minimum required.

image

It's been updated in branch `issue1913_g36_final`, block [`Buildings.Controls.OBC.ASHRAE.G36.TerminalUnits.Reheat.Controller`](https://github.com/lbl-srg/modelica-buildings/blob/issue1913_g36_final/Buildings/Controls/OBC/ASHRAE/G36/TerminalUnits/Reheat/Controller.mo). Note that the sequence is in-progress and may still needs to be validated.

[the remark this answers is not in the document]
AntoineGautier: You are right. There was a confusion on my end between the required design inputs and the computed set points (depending on the operating mode).

Buildings/Resources/Scripts/Dymola/ConvertBuildings_from_7_to_8.0.0.mos
JayHuLBL: I agree to add the script as you suggested, to update the default values of the set points. So far, I also cannot see how to solve the issue when the user made the change to the default values.
I suggest that we just update the default values.

JayHuLBL:
@AntoineGautier I reverted the changes as we discussed earlier. It will be ready for your second review when the tests passed.

AntoineGautier (approved):

JayHuLBL:
@mwetter It could be merged if it is good to you.

PR #2190, opened 2020-10-20
Added vendor annotation showing default value of the conditional removable connectors

### LINKED ISSUE #1854: Add default value to conditional connectors of CDL sequences
This is to add default value to conditional connectors of CDL sequences.
```
CDL.Interfaces.IntegerInput nOcc if have_occSen
  "Number of occupants"
    annotation (__cdl(default = 0));
```

JayHuLBL:
This closes #1854. 

- [x] added default values to conditional removable connectors as following:
- For RealInput, the default values are:
  - If `unit=K`, the default is 293.15 K.
  - If `quantity="Pressure"`, the default is 101325 Pa.
  - If `quantity="PressureDifference"`, the default is 0 Pa.
  - For all other units, the default value is 0.
- For `IntegerInput`, the default value is `0`.
- For `BooleanInput`, the default value is `false`.
- For `DayTypeInput`, the default value is `WorkingDay`.

### DIFF
--- a/Buildings/Controls/OBC/ASHRAE/G36_PR1/AHUs/MultiZone/VAV/Controller.mo
+++ b/Buildings/Controls/OBC/ASHRAE/G36_PR1/AHUs/MultiZone/VAV/Controller.mo
@@ -1000,7 +1000,7 @@ Buildings.Controls.OBC.ASHRAE.G36_PR1.AHUs.MultiZone.VAV.SetPoints.SupplySignals
 revisions="<html>
 <ul>
 <li>
-October 20, 2020, by jianjun Hu:<br/>
+October 20, 2020, by Jianjun Hu:<br/>
 Added vendor annotation to show the default value of the conditional removable connectors.<br/>
 This is for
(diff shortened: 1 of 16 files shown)


mwetter (changes requested):
@JayHuLBL : I think what *default* values should be added should be part of the CDL specification: This can be automated during the translation, and we should automate what can be automated. The `__cdl(default...)` annotation should only be used where we differ from these defaults:
For RealInput, the default values are:

    If unit=K, the default is 293.15 K.
    If quantity="Pressure", the default is 101325 Pa.
    If quantity="PressureDifference", the default is 0 Pa.
    For all other units, the default value is 0.

For IntegerInput, the default value is 0.
For BooleanInput, the default value is false.
For DayTypeInput, the default value is WorkingDay.

JayHuLBL:
@mwetter Thanks. Then I see there is no need to add the vendor annotation to the existing sequences. The default value will be added by m2j through [issue143](https://github.com/lbl-srg/modelica-json/issues/143). This PR should be closed.

PR #2239, opened 2020-11-11
Refactored for #2229

### LINKED ISSUE #2229: Invalid model because of lookup in partial class
Please check [Buildings.Fluid.HeatExchangers.DXCoils.BaseClasses.Examples.NominalCondition](https://libraries.openmodelica.org/branches/newInst/Buildings_7.0.0/files/Buildings_7.0.0_Buildings.Fluid.HeatExchangers.DXCoils.BaseClasses.Examples.NominalCondition.err). OpenModelica reports that
```
[Buildings 7.0.0/Fluid/HeatExchangers/DXCoils/BaseClasses/NominalCondition.mo:36:3-40:32:writable] 
Error: Medium is partial, name lookup is not allowed in partial classes.
```
The package `Medium` is declared to be a replaceable partial class, but is not subsequently redeclared to a non-partial one before being used. This is forbidden by the [Modelica Specification, Section 5.3.2](https://specification.modelica.org/v3.4/Ch5.html#composite-name-lookup): "The class we look inside may not be partial in a simulation model".

mwetter:
This closes #2229.

This also removes the `BaseClasses.NominalCondition` record which is only used in this model through an extend statement.


mwetter:

PR #2241, opened 2020-11-11
Enabled omc tests

### LINKED ISSUE #2233: OpenModelica fails to install on travis
OpenModelica started to fail to install during the CI tests, see for example
https://travis-ci.com/github/lbl-srg/modelica-buildings/jobs/432365392

This issue is to temporarily remove it from the tests, and then enable it through a separate PR so that the other tests can continue.

mwetter:
This closes #2233


AntoineGautier:
@mwetter I noticed a glitch here https://travis-ci.com/github/lbl-srg/modelica-buildings/jobs/434048674 where the apt addons installation was launched (and failed) despite `TEST_ARG="make test-omc"` being commented out. I wonder if the conditional assertion should not be rather `if: env(TEST_ARG) =~ /*test-omc*/` per https://docs.travis-ci.com/user/conditions-v1#specification

mwetter:
@AntoineGautier : Let me remove the condition first to make sure it installs. I think the problem is that it does not find the omc package.

mwetter:
@AntoineGautier : It looks like conditions are not supported in `addons`, see https://travis-ci.com/github/lbl-srg/modelica-buildings/jobs/444033561/config  I will remove the condition.

mwetter:

Issue #2202, opened 2020-10-29
UrbanOpt_TimeSeries terminal Unit: heating and chilled water deltaT is very low comparing to the nominal value


Helarga:
For the time series model at Buildings.Applications.DHC.Loads.Examples.CouplingTimeSeries
which lives in the staging branch of issue1437_district_heating_cooling

the heating and chilled water temperature difference across the terminal unit heat exchanger is very low "0.5degC in case of heating and 0.2degC in case of cooling" comparing to the nominal deltaT of 5 degC.

for heating terminal unit:

![image](https://user-images.githubusercontent.com/47648835/97606829-d8be2300-19d5-11eb-8718-88284fbc2c42.png)

for cooling terminal unit:

![image](https://user-images.githubusercontent.com/47648835/97606762-c512bc80-19d5-11eb-8d19-8cee0c63eeb0.png)

AntoineGautier:
Nice catch, thanks.
That is indeed a bug. The nominal HHW and CHW mass flow rate in the FCU components of `Buildings.Applications.DHC.Loads.Examples.BaseClasses.BuildingTimeSeries` should be scaled by `1 / facSca` as it is done in `Buildings.Applications.DHC.Loads.Validation.TerminalUnitScaling`.

AntoineGautier:
This will be addressed through https://github.com/lbl-srg/modelica-buildings/issues/2302.

Issue #2201, opened 2020-10-28
UrbanOpt_BuildingTimeSeries model fails if time series load file changes


Helarga:
For the model BuildingTimeSeries located in  
Buildings.Applications.DHC.Loads.Examples.BaseClasses.BuildingTimeSeries
and lives at the staging branch: issue1437_distrcit_heating_cooling.

The  <code>facScaCoo</code> and <code>facScaHea</code> are two parameters implemented to scale up/down the terminal units capacity and equal to 40 and 10 respectively. These values have been evaluated using trial and error approach and accordingly they are customized for a single time series file of "Buildings/Applications/DHC/Loads/Examples/Resources/SwissResidential_20190916.mos"

Two time series files generated from Open studio were examined 
MediumOffice-90.1-2010-5A.mos
primaryschool-3c-2013.mos

When the user changes the time series  loads file, this error message will show up 
![image](https://user-images.githubusercontent.com/47648835/97495970-ecae4a00-192d-11eb-943a-7b52b50c641a.png)

![image](https://user-images.githubusercontent.com/47648835/97496383-78c07180-192e-11eb-9f01-d625fb1a5e16.png)

AntoineGautier:
Please see also the limitations described here: https://github.com/lbl-srg/modelica-buildings/issues/2147#issuecomment-705192554 and the important mention that this model was not developed for generic use but as an example illustrating how to extend `Buildings.Applications.DHC.Loads.BaseClasses.PartialBuilding` and `Buildings.Applications.DHC.Loads.BaseClasses.PartialTerminalUnit` to address various modeling needs.

Helarga:
thanks for your reply, yep i have read it yesterday actually. I still need to keep the current ticket because i have hyperlinked it under the documentation i.e. revisions section of the model.

Helarga:
[...]

PR #2249, opened 2020-11-14
Issue1951 5 gdhc ets

AntoineGautier:
This closes #1951
This is an update of #2065 (closed without merge) which includes

1. the modifications after @mwetter's review, see https://github.com/lbl-srg/modelica-buildings/pull/2065#issuecomment-686783032,
2. the transition to the new package location under `Experimental.DHC`,
3. the addition of a block computing the difference in enthalpy flow rate between two fluid streams (and optionally its time integral), see `Buildings.Experimental.DHC.Networks.BaseClasses.DifferenceEnthalpyFlowRate`,
4. the modification of the district HX control, see `Controls.HeatExchanger`: previously the primary pump speed was modulated to maintain a constant ΔT on the secondary side. This was not satisfactory in case of cold rejection at low district water temperature. Before resetting down the CHWST, the secondary inlet T can indeed be higher than the nominal, resulting in a ΔT with the primary inlet lower than nominal so a lower reheating capacity at nominal mass flow rate. This forced the primary and secondary pumps to be operated at maximum speed (even at low load) before the CHWST getting reset, 
5. the modification of the hot side controller, see `Controls.SideHot`: the logic is now simpler, with a better control in the end and no more short cycles between different modes (also only the temperature at the top of the HHW tank and at the bottom of the CHW tank are now needed),
6. the suppression of the state graphs that are no more needed with the previous modification,
5. the deletion of the alternative supervisory controller (formerly `Generation5.Controls.Supervisory1`) which turned out too complex with no clear benefit.

@mwetter @JayHuLBL Which of you would be willing to have a final look at it? Hopefully in the next couple weeks because Katy is waiting for the `DifferenceEnthalpyFlowRate` to finalize a PR on her side.

PR #2269, opened 2020-11-23
Issue2201 loads time series

Helarga:
This PR addresses issue #2201 and issue #2202  where 

the modifications of time series model allow for uploading different time series loads files. The <code>facScaHea</code> and <code>facScaCoo</code> values are assigned to {1} hence no trial and error approach is needed.

and it includes

four identical examples " with four different time series files"  just to highlight that the model serves its purpose. The four types of buildings are : Hospital, primary school, medium office and residential.


Helarga:
@JayHuLBL, why there is a conflict with the MediumOffice-90.1-2010-5A.mos load file?  I think this load file was not available at the MBL, We implemented it first at the GMT as i remember.

JayHuLBL:
@Helarga It was merged to master at PR #2249, which was to address issue #1951.

Helarga:
@JayHuLBL ok how we may proceed?

JayHuLBL:
> @JayHuLBL ok how we may proceed?

Pull the latest mater to your development branch (in your repo). I believe you will see the conflict again as seen here. To solve it, please replace the `MediumOffice-90.1-2010-5A.mos` in your branch with the one in the master branch. Then rerun the simulations which uses this data. Push the change to your own repo. I will pull the changes in your branch (Helarga:issue2201_loadsTimeSeries) to the same branch here (lbl-srg:issue2201_loadsTimeSeries).

Helarga:
@JayHuLBL kindly pull my branch from 
https://github.com/Helarga/modelica-buildings/tree/issue2201_loadsTimeSeries

Buildings/Experimental/DHC/Loads/Examples/CouplingTimeSeries_Hospital.mo
```diff
@@ -0,0 +1,158 @@
...
+  annotation (
+    Diagram(
+      coordinateSystem(
+        preserveAspectRatio=false,
+        extent={{-160,-140},{160,140}})),
+    __Dymola_Commands(
+      file="modelica://Buildings/Resources/Scripts/Dymola/Experimental/DHC/Loads/Examples/CouplingTimeSeries_Hospital.mos" "Simulate and plot"),
+    experiment(
+      Tolerance=1e-6,
+      StopTime=86400),
```
JayHuLBL, line 142: This stop time (86400 seconds) is not the same as the one used in its `.mos` script (which is 604800 seconds).

Helarga: @JayHuLBL fixed it , thanks!
https://github.com/Helarga/modelica-buildings/tree/issue2201_loadsTimeSeries

Helarga:
@JayHuLBL I am not sure why the CI build failed.

JayHuLBL:
> @JayHuLBL I am not sure why the CI build failed.

The class order of package `Buildings/Experimental/DHC/Loads/Examples` is not right. To fix that, run below
```
jianjun@jianjun-virtual-machine:~/GitFolder/modelica-buildings/Buildings$ ipython
Python 3.6.9 (default, Jul 17 2020, 12:50:27) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.11.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import buildingspy.development.refactor as r                                                                                                                             

In [2]: r.write_package_order(".", recursive=True)                                                                                                                               

In [3]: exit() 
```

Helarga:
@JayHuLBL thanks , fixed it.

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
```diff
@@ -2,19 +2,21 @@ within Buildings.Experimental.DHC.Loads.Examples.BaseClasses;
...
-    redeclare package Medium=Buildings.Media.Water,
+    redeclare package Medium=MediumW,
     final have_fan=false,
     final have_pum=true,
     final have_eleHea=false,
     final have_eleCoo=false,
     final have_weaBus=false);
-  package Medium2=Buildings.Media.Air
-    "Load side medium";
+  package MediumA=Buildings.Media.Air
```
JayHuLBL, line 11: @Helarga @AntoineGautier Remind me about the discussion on the (`Medium1`, `Medium2`) vs (`MediumW`, `MediumA`), did we decide to use (`MediumW` and `MediumA`) in `Example` and `Validation` package, but use (`Medium1` and `Medium2`) in models?

Helarga: @JayHuLBL yes we did, Medium1 "Source", Medium2"Load" in models

Helarga: @JayHuLBL  https://github.com/lbl-srg/modelica-buildings/pull/2296

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
```diff
@@ -2,19 +2,21 @@ within Buildings.Experimental.DHC.Loads.Examples.BaseClasses;
...
+    "Air medium";
+  package MediumW=Buildings.Media.Water
+    "Water medium";
   parameter String filNam
     "File name with thermal loads as time series";
-  parameter Real facScaHea=10
+  parameter Real facScaHea=1
     "Heating terminal unit scaling factor";
```
JayHuLBL, line 18: The comment of the scaling factor is not clear to me. It seems like that the original scaling factor is to break down the load to individual terminal unit. This is why I see originally in the modification of `QCoo_flow_nominal`, `min = Buildings.Experimental.DHC.Loads.BaseClasses.getPeakLoad()` without multuplying`facScaCoo`.
But the new change intends to use the scaling factor to scale up from one terminal unit to whole building?

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
```diff
@@ -84,7 +90,7 @@ model BuildingTimeSeries
       each unit="W"),
     offset={0,0,0},
     columns={2,3,4},
-    smoothness=Modelica.Blocks.Types.Smoothness.MonotoneContinuousDerivative1)
+    smoothness=Modelica.Blocks.Types.Smoothness.LinearSegments)
```
JayHuLBL, line 93: `MonotoneContinuousDerivative1` would output more smooth time series which attributes to better numerical performance.

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
```diff
@@ -120,37 +126,34 @@ model BuildingTimeSeries
...
-  DHC.Loads.Validation.BaseClasses.FanCoil2PipeCooling terUniCoo(
-    redeclare final package Medium1=Medium,
-    redeclare final package Medium2=Medium2,
+  Buildings.Experimental.DHC.Loads.Validation.BaseClasses.FanCoil2PipeCooling terUniCoo(
+    redeclare final package Medium1=MediumW,
+    redeclare final package Medium2=MediumA,
     final facSca=facScaCoo,
-    final QHea_flow_nominal=QHea_flow_nominal/facScaHea,
-    final QCoo_flow_nominal=QCoo_flow_nominal/facScaCoo,
+    final QCoo_flow_nominal=QCoo_flow_nominal,
```
JayHuLBL, line 151: In original implementation, the `QHea_flow_nominal` is specified for each unit, and the `facSca` will scale up the settings to multiple unit.

But in the new change, the `QHea_flow_nominal` is already specified for multiple units and the `facSca` is still there? Again, it seems to related the comment I made previously, what does the scaling factor actually mean?

AntoineGautier: @JayHuLBL See here for the documentation of the scaling factor (if it helps).
https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Experimental/DHC/Loads/BaseClasses/PartialTerminalUnit.mo#L562-L592

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
Helarga: ok, i will modify it.

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
Helarga: @JayHuLBL do you still want me to elaborate on this item?

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
Helarga: the intention of the change is to avoid the trial and error approach used previously to customize the scaFacHea, scaFacCoo for the implemented time series loads file. If the user change the time series loads file, he/she will face an error like that

image
 i.e. error indicates a violation  of the 2nd law of thermodynamics at the terminal unit HX. 

In addition , Only one heating terminal unit is serving the building heating loads and only one cooling terminal unit is serving the buildings cooling loads, we dont have buildings zones for multiple units. After the current changes if the user "for some reason" wants to scale up or down the QHea/QCoo_nominal  i.e. design parameters of the heating / cooling terminal unit the model will still run without any error.. If not, {1} shall be the default value for the two parameters.

Buildings/Experimental/DHC/Loads/Examples/CouplingTimeSeries_Hospital.mo
```diff
@@ -0,0 +1,158 @@
+within Buildings.Experimental.DHC.Loads.Examples;
+model CouplingTimeSeries_Hospital
+  "Example illustrating the coupling of a hospital building model
```
JayHuLBL, line 3: make the comment in one line.

Buildings/Experimental/DHC/Loads/Examples/CouplingTimeSeries_MediumOffice.mo
```diff
@@ -0,0 +1,158 @@
+within Buildings.Experimental.DHC.Loads.Examples;
+model CouplingTimeSeries_MediumOffice
+  "Example illustrating the coupling of a simultaneous heating and cooling
+    office building loads model to heating and chilled water loops."
```
JayHuLBL, line 4: one line comment.

Buildings/Experimental/DHC/Loads/Examples/CouplingTimeSeries_PrimarySchool.mo
```diff
@@ -1,13 +1,20 @@
 within Buildings.Experimental.DHC.Loads.Examples;
-model CouplingTimeSeries
-  "Example illustrating the coupling of a building model to heating water and chilled water loops"
+model CouplingTimeSeries_PrimarySchool
+  "Example illustrating the coupling of a school building model to heating
```
JayHuLBL, line 3: one line.

Helarga: done!
https://github.com/Helarga/modelica-buildings/tree/issue2201_loadsTimeSeries

Buildings/Experimental/DHC/Loads/Examples/CouplingTimeSeries_MediumOffice.mo
Helarga: done

Buildings/Experimental/DHC/Loads/Examples/CouplingTimeSeries_Hospital.mo
Helarga: done

Buildings/Experimental/DHC/Loads/Examples/CouplingTimeSeries_Residential.mo
```diff
@@ -0,0 +1,178 @@
+within Buildings.Experimental.DHC.Loads.Examples;
+model CouplingTimeSeries_Residential
+  "Example illustrating the coupling of a residential building model to 
+    heating water and chilled water loops"
```
JayHuLBL, line 4: one line.

Helarga: yes already done here https://github.com/lbl-srg/modelica-buildings/pull/2297

JayHuLBL (approved):
@Helarga @AntoineGautier I agree with the changes. 

The CI test is running just for the `DHC` package. If you agree with the change as well, we can then enable CI test over all packages.

Helarga:
@JayHuLBL the CI test has stopped.

Buildings/Experimental/DHC/Loads/Examples/CouplingTimeSeries_Hospital.mo
```diff
@@ -0,0 +1,157 @@
...
+    __Dymola_Commands(
+      file="modelica://Buildings/Resources/Scripts/Dymola/Experimental/DHC/Loads/Examples/CouplingTimeSeries_Hospital.mos" "Simulate and plot"),
+    experiment(
+      Tolerance=1e-6,
+      StopTime=604800),
+    Documentation(
+      info="<html>
+<p>
+This example demonstrates the hydraulic connection between a hospital's heating and cooling
+loads provided as time series to infinite sources of district heating and cooling.
```
AntoineGautier, line 146: This sentence (repeated in all the new example models) is not clear to me.

AntoineGautier (changes requested):

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
Helarga: **The rational behind the changes is to make the model work at the GMT because currently it fails if the time series loads file changes.** 

The changes are to avoid using the trail and error approach used at the current model to evaluate two customized scaling factor parameters every time the time series file change because if we don't do that, the model will simply fail.

**Discussion**
We consider the time series loads model as a special type of UO building loads since we don't have thermal zones, AHU data , FCU data , whether a radiant floors are implemented or not etc,  all what we know is the aggregated time series thermal loads for the whole building's zones.

So we assume to have one ideal terminal unit to handle the heating time series loads and one ideal terminal unit to handle the cooling time series loads. Hence I believe that for this specific type of building loads, we have to consider “mentally” a larger capacity systems, and not only hover around the limited capacity of a FCU.

On the other hand, the commercial FCU nominal air flow rate could start from 300cfm to 3000cfm, hence there is no universal nominal air flow = 1kg/s that we should commit to, i mean what if the user decided to change the nominal flow rate to 0.5 kg/s, this change will need at least two weeks of trial and error  efforts “as per your experience in evaluating the 10 and 40 values” to find the correct combination of <code>scaFacHea and scaFacCoo </code>.

For the sake of brevity, depending on "trial and error evaluated parameters" is not an extensible approach for the GMT.

**Conclusion**
The <code>scaFac</code>10  and  40 both are evaluated arbitrarily i.e by trial and error, and if the time series file changes; the model will fail and we are trying to solve these problems because it will provoke the extensibility of the UrbanOpt_GeoJson Modelica Translator UO_GMT.

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
```diff
@@ -38,20 +40,24 @@ model BuildingTimeSeries
   parameter Modelica.SIunits.Temperature T_aLoaCoo_nominal=273.15+24
     "Load side inlet temperature at nominal conditions in cooling mode"
     annotation (Dialog(group="Nominal condition"));
-  parameter Modelica.SIunits.MassFlowRate mLoaHea_flow_nominal=1
+  parameter Modelica.SIunits.MassFlowRate mLoaHea_flow_nominal=QHea_flow_nominal/(cp_air*delTAirHea)
     "Load side mass flow rate at nominal conditions in heating mode"
     annotation (Dialog(group="Nominal condition"));
-  parameter Modelica.SIunits.MassFlowRate mLoaCoo_flow_nominal=1
+  parameter Modelica.SIunits.MassFlowRate mLoaCoo_flow_nominal=-QCoo_flow_nominal/(cp_air*delTAirCoo)
     "Load side mass flow rate at nominal conditions in cooling mode"
```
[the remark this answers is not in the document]
Helarga: So far we have "sensible loads only" FCU, so we used a sensible loads equation, and once we change the FCU to handle the latent loads we can change the equation to the difference between enthalpy values of the supply and room points "S&R on psychometric chart", and later we can consider the "M" point if we have a FA...etc.
This how we work at MBL models , we start by a model, change update and modify then add "revisions" under the revision section for each model.

Buildings/Experimental/DHC/Loads/Examples/BaseClasses/BuildingTimeSeries.mo
```diff
@@ -38,20 +40,24 @@ model BuildingTimeSeries
...
-    max=-Modelica.Constants.eps)=Buildings.Experimental.DHC.Loads.BaseClasses.getPeakLoad(
+    max=-Modelica.Constants.eps)=facScaCoo*Buildings.Experimental.DHC.Loads.BaseClasses.getPeakLoad(
     string="#Peak space cooling load",
     filNam=Modelica.Utilities.Files.loadResource(filNam))
     "Design cooling heat flow rate (<=0)"
     annotation (Dialog(group="Design parameter"));
   parameter Modelica.SIunits.HeatFlowRate QHea_flow_nominal(
-    min=Modelica.Constants.eps)=Buildings.Experimental.DHC.Loads.BaseClasses.getPeakLoad(
+    min=Modelica.Constants.eps)=facScaHea*Buildings.Experimental.DHC.Loads.BaseClasses.getPeakLoad(
     string="#Peak space heating load",
```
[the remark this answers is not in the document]
Helarga: We decided this point internally, so if the user wants to add a safety percentage “ during the conceptual design phase” 20% for eg. the heating peak loads,  so simply she/he will change the <code> scaFacHea to 1.2 </code> and the same approach could be applied for cooling purposes.

nllong:
@AntoineGautier are these technical implementation differences resolvable after our milestone? CU needs the updates ASAP. I propose that we merge and resolve later as needed. Let me know your thoughts.

AntoineGautier:
@nllong @Helarga The proposed changes introduce some ambiguity in `DHC.Loads.Examples.BaseClasses.BuildingTimeSeries`. I will try to clarify why this is an issue.
The scaling factor of the terminal unit can be used in two different ways depending on the value of `have_scaLoa`.

- If `have_scaLoa = true` (default) then it basically "breaks down the load" to represent multiple identical realistic terminal units (that can be parameterized based on manufacturer data). That was the original use of that parameter in the example model with the time series building.
- If `have_scaLoa = false` then it serves as a "load multiplier" to represent for instance multiple identical zones. It *seems* that the changes here try to use that approach. However
  - It does not work as it is implemented now, for two reasons. First, `have_scaLoa` is not set to `false` in the FCU components of `BuildingTimeSeries`. Second, there is an actual bug in `PartialTerminalUnit` that still needs to be corrected (`QReqHea_flow` should not be scaled is `have_scaLoa = false`). So for instance if you set `bui.facSacHea = 10` (instead of 1) in `CouplingTimeSeries_Hospital` the actual heating heat flow rate `bui.QHea_flow` or the required heating water mass flow rate `bui.terUniHea.mReqHeaWat_flow` will remain the same (only `bui.QHea_flow_nominal` will be increased).
  - It is not documented anywhere in `BuildingTimeSeries` so there is no way one can understand the proper use of the scaling factor exposed in that model and propagated to the instance of the FCU model.

Regarding the best course of action, in my opinion.

- In any case `BuildingTimeSeries` must be refactored. The simplest way would be to not expose the scaling factors since they are anyway not used properly. However, we should check with @khinkelman and @JingWang-CUB if they actually use those parameters in their current development.
- I propose that we change the target branch towards `issue2204_gmt_mbl` and not merge those changes in the master.
- I will open a new issue to refactor `PartialTerminalUnit` to use two distinct scaling parameters and propagate them more clearly in `BuildingTimeSeries`.

nllong:
@AntoineGautier perhaps targeting the `gmt_mbl` branch would be the fastest resolution. We need to get this merged so that the models the GMT creates support the CU case study (and OSU case study).

AntoineGautier:
I agree.
@khinkelman @JingWang-CUB please confirm that you do not assign different values than 1 (the default) to the parameters `facScaHea` and `facScaCoo` of `BuildingTimeSeries` in your current development. (There is a bug that is still not fixed by #2269 and the time series loads will not be scaled as expected. The fastest way to move forward is to remove those parameters from `BuildingTimeSeries` for now.)

JingWang-CUB:
@AntoineGautier In my current development, I do not assign different values other than 1.

khinkelman:
I do not change them from 1 either. Thanks @AntoineGautier for checking.

Helarga:
@AntoineGautier so this bug is here Buildings.Experimental.DHC.Loads.BaseClasses.PartialTerminalUnit,
yes i didn't noticed it, good catch,  I only noticed and fixed this bug #2202.

Ok, I will remove the scaFacHea and scaFacCoo and open a new PR against the staging branch.

Thanks for the review efforts and the proposal to refactor the model, we do need it.

AntoineGautier:
Thanks. I created a branch from `issue2204_gmt_mbl` to relocate the developments from the current feature branch, see
https://github.com/lbl-srg/modelica-buildings/tree/issue2204_2201_loadsTimeSeries

AntoineGautier:
This is closed without being merged.
The development was updated based on https://github.com/lbl-srg/modelica-buildings/pull/2269#issuecomment-741963606 and merged into `issue2204_gmt_mbl` through #2306.
A new issue is open to track what still needs to be addressed on the master branch, see  #2302

Issue #2170, opened 2020-10-09
For CDL blocks that sample, change blocks to sample at multiple period align with t=0


mwetter:
Currently, some CDL blocks have a parameter `startTime=0` (which is not clear what it means when translated to a BAS) because it could be local time, unix time, UTC time, or seconds since the controller was switched on.

This issue is to change the CDL blocks that sample based on some `period` (such as `Pulse`) or `samplePeriod` (such as `Discrete.Sampler`) so that they sample at t=0, and at `t_k = k *period` where `k` is any (negative or positive) integer.

The initial sample time can for example be set to
```modelica
initial equation 
  t0 = round(integer(time/period)*period, 6);
```
which is already implemented for `Sources.TimeTable` (through https://github.com/lbl-srg/modelica-buildings/pull/2136).

Todo:
- [x] Move
```modelica
  function round "Round function from Buildings.Controls.OBC.CDL.Continuous.Round"
    input Real x "Argument";
    input Real n "Digits";
    output Real y "Rounded argument";
  protected 
    Real fac = 10^n "Factor used for rounding";
  algorithm 
    y :=if (x > 0) then floor(x*fac + 0.5)/fac else ceil(x*fac - 0.5)/fac;
  end round;
```
to the `Utility.Math.Functions` package, and also provide a block that implements this function because all functions are also available as a `block` in the Buildings library (and propose to add to IBPSA).

- Refactor the blocks that sample:
  - [x] `{Continuous,Integers,Logical}.Sources.Pulse` and `Logical.Sources.SampleTrigger`: Remove parameter `startTime`.
  - [x] `Discrete.{FirstOrderHold,Sampler,UnitDelay,ZeroOrderHold}`: Refactore `t0` using `round()`.

PR #2283, opened 2020-12-03
Issue2282 sample trigger and pulse: rename delay to shift

### LINKED ISSUE #2282: CDL: delay parameter in SampleTrigger and Pulse
Rename `delay` parameter in `Buildings.Controls.OBC.CDL.Logical.Sources.SampleTrigger`,  `Buildings.Controls.OBC.CDL.Logical.Sources.Pulse`, `Buildings.Controls.OBC.CDL.Continuous.Sources.Pulse`, and `Buildings.Controls.OBC.CDL.Integers.Sources.Pulse` to `shift`.

milicag:
Closes #2282.


mwetter (changes requested):
@milicag : This is not the right fix. It essentially reverses the change done through https://github.com/lbl-srg/modelica-buildings/issues/2170

The problem in this new implementation is just with the mod( , ) function if arguments are equal but non-zero. Please correct this in a robust way.

milicag:
@mwetter Just to make sure I understand the control intent of the blosk, you would like the sampler to always sample at times where values mod(time, period) = 0, but not until the delay time has elapsed?

mwetter:
The delay is not a time that need to elapse, but rather a shift in time,
see figure in Buildings.Controls.OBC.CDL.Logical.Sources.SampleTrigger

milicag:
So why don't we call it `shift` parameter?

mwetter:
That may be more suitable. Can you please make the change (and update conversion scripts if we used this before) and also update the figures. The same applies to similar blocks, such as `Pulse`.

milicag:
@mwetter Should I remove release notes that are deprecated through revisions within the same release cycle?

mwetter:
@milicag : Release notes should only show what changed relative to previous releases, so if there are notes that we reverted, they should be removed. The same applies to the conversion script. Both are meant for official releases only.

milicag:
@mwetter As I assumed. I will go through both conversion script and release notes and remove redundancies. There was, for example, a version that specifies a number of periods for the pulse that is now removed.

mwetter (approved):

mwetter:
@milicag : What's the purpose of https://github.com/lbl-srg/modelica-buildings/pull/2283/commits/8bc9b022ddbb5fc09afaf0e9c0b9b198c56031ba?

milicag:
@mwetter I just merge the master in. It looks like git allowed me to push an empty commit. I thought there were a few changes to the master, as I had them locally. If the remote was identical I expected git to be clever about it. Sorry for delaying the test. I merged this already into the chiller plants branch.

PR #2298, opened 2020-12-08
re run the unit test locally to count for combiTable smoothness change

Helarga:


JayHuLBL:
@Helarga the reference results have been updated.

PR #2310, opened 2020-12-14
Issue2291 partial ets

### LINKED ISSUE #2291: Base classes for ETS, building with ETS, and plant
This issue addresses the following.

- [x] Refactor `Buildings.Experimental.DHC.EnergyTransferStations.BaseClasses.PartialETS` to support all DHC system types (an enumeration type reflecting the structure of the `DHC` package may be used to define a structural parameter),
- [x] Refactor `Buildings.Experimental.DHC.Networks.BaseClasses.DifferenceEnthalpyFlowRate` to support two different media
- [x] Develop a class `PartialBuildingWithPartialETS`
- [x] Develop a class `PartialPlant`
- [x] Change the parameter names from `have_watHea` and `have_watCoo` into `have_chiWat` and `have_heaWat` in `DHC.Loads.BaseClasses.PartialBuilding` and `DHC.Loads.BaseClasses.PartialTerminalUnit`, and change all default values to `false` instead of `true`
  - This is ***non-backward compatible*** for all developments with a dependency to classes that extend these base classes.

The feature branch is `issue2291_partialETS`. 

[EDIT] This branch also includes the following changes.

- In `PartialTerminalUnit`, change of the parameter set that pertains to the scaling mechanism in order to use two distinct scaling parameters (instead of one real scaling parameter which effect changes depending on another Boolean parameter)
- Implement a similar scaling mechanism for `PartialBuilding` and `PartialBuildingWithPartialETS`

AntoineGautier:
This closes #2291.


AntoineGautier:
Converted to draft until URBANopt team can provide some feedback (expected 1/4/21).

khinkelman:
Hi @AntoineGautier, thank you for your efforts in creating these base classes! Overall, they look good to me. Below are my comments. Please let me know if you have any questions.

- `DHC.CentralPlants.BaseClasses.PartialPlant`: The heating/cooling power is labeled as electric. Some systems will use gas (e.g. boilers). Should we have a second real output `PHeaGas` for this case (and add Boolean `have_gasHea`)? Or do you think it is not common enough to justify a dedicated output in this base class?
- The design with `TypDisSys` and the multiple medium options looks good to me. I wonder though if you want to set the default `TypeDisSys` to `CombinedGeneration2to4` (e.g. in `PartialPlant` and `PartialETS`). This would then be consistent with the default liquid water medium declarations in these models.
- General design question: Why are you declaring the medium models as `Buildings.Media.Water` constrained by `PartialMedium` rather than just declaring them as `Modelica.Media.Interfaces.PartialMedium`?
The reason why I ask is there are functions in the top level `Buildings.Media.Steam` that are unavailable in `PartialMedium`, which I have worked around to date, but I wonder if using `constrainedby` similar to what you've done here would be a better approach. 

Thanks!

AntoineGautier:
@khinkelman Thanks for your feedback.

> - Should we have a second real output `PHeaGas` for this case (and add Boolean `have_gasHea`)?

Yes. I included an array connector for the fuel rates (compatible with plants using both oil and gas). The corresponding structural parameter is the number of fuel types (0 meaning no combustion system).

> I wonder though if you want to set the default `TypeDisSys` to `CombinedGeneration2to4` (e.g. in `PartialPlant` and `PartialETS`).

I agree and used that as a default assignment.

> General design question: Why are you declaring the medium models as Buildings.Media.Water constrained by PartialMedium rather than just declaring them as Modelica.Media.Interfaces.PartialMedium?

See our discussion during the last meeting. This basically avoids aliasing the inlet medium and allows redeclaring the inlet medium without having to redeclare the outlet medium for steam systems.

AntoineGautier:
Converting to draft until the segmentation fault (with the three simulation tools!) is solved.

AntoineGautier:
@mwetter In the commit 6656a6e Travis gets a segmentation fault for the models

- `Buildings.Experimental.DHC.Loads.Examples.CouplingTimeSeriesSingleLoop`
- `Buildings.Experimental.DHC.Loads.Examples.CouplingTimeSeries`

with the three simulation tools!

On my system, the only one I get systematically is for the second model, and only with JModelica.

In my last commit cc76fce I conditionally removed the data reader for the load file and replaced it with a sinusoidal source, see parameter `have_dumLoa = true` in `Buildings.Experimental.DHC.Loads.Examples.BaseClasses.BuildingTimeSeries`. 
This seems to solve the issue.

Do you have any idea where this might come from?

mwetter:
@AntoineGautier : I don't see a reason for this segmentation fault. The data reader was updated in a recent version of the Modelica Standard Library, but I never encountered a problem when using it. Also the models don't use Spawn, so there is no suspicion that a Spawn coupling would cause it.

Buildings/Experimental/DHC/Loads/BaseClasses/Validation/BaseClasses/Building.mo
```diff
@@ -0,0 +1,95 @@
...
+      coordinateSystem(
+        preserveAspectRatio=false)),
+    Documentation(info="<html>
+<p> 
+This is a minimum example of a class extending 
+<a href=\"modelica://Buildings.Experimental.DHC.Loads.BaseClasses.PartialBuilding\">
+Buildings.Experimental.DHC.Loads.BaseClasses.PartialBuilding</a>
+developed for testing purposes only.
+</p>
+</html>"));
```
mwetter, line 94: Please add revision notes. Also some other models in this package miss the revision notes.

mwetter (approved):
This is ready to merge after you add the missing revision notes. Please see the inline comment.

AntoineGautier:
@mwetter This is now ready, could you please merge?

Issue #2059, opened 2020-07-30
Refactor Examples.VAVReheat.ThermalZones.VAVBranch and VAV models to expose the coil sizing parameters


AntoineGautier:
In `Buildings.Examples.VAVReheat.ThermalZones.VAVBranch` the sizing of the heating coil is incorrect as it uses the nominal air flow rate (cooling design conditions) instead of the air flow rate at heating design conditions. Also `T_a2_nominal` is set to 82.2 degC and is not an exposed parameter whereas `souTer` supplies water at 50 degC (not exposed) and `Guideline36` and `ASHRAE2006` model the AHU heating coil with a constant HHW temperature of 45 degC. Eventually, best would be to use ports `*1` for water and ports `*2` for air.

The AHU coils should use the same design temperatures as the terminal reheat coils.
Also the nominal air inlet temperature should correspond to the TMix limit of the freeze protection: 4 degC for AHSRAE2006, 6 degC for Guideline36, so 4 degC should be used. 
(Note that this discrepancy between the two models should be reconciled for a fair comparison of HVAC energy use: currently ASHRAE2006 allows higher outdoor air flow rate in cold conditions.)

AntoineGautier:
HVAC site energy use computed with OCT
```
branch=master
commit=895a8dc11c40ee36d8cfbf09bc5cbd9caca65962
```
```
.. table:: Heating, cooling, fan and total site HVAC energy, and savings of guideline 36 case versus base case.

   ===================================== ===================================== ====================================== ====================================== =====
   :math:`E_{h} \quad [kWh/(m^2\,a)]`    :math:`E_{c} \quad [kWh/(m^2\,a)]`    :math:`E_{f} \quad [kWh/(m^2\,a)]`     :math:`E_{tot} \quad [kWh/(m^2\,a)]`     [%]
   ===================================== ===================================== ====================================== ====================================== =====
                                   4.813                                  16.1                                  2.332                                  23.25   
                                    3.67                                 14.77                                  1.902                                  20.34  12.5 
   ===================================== ===================================== ====================================== ====================================== =====
  
  
CPUtime, annual_base 72'
CPUtime, annual_g36 133'
```

PR #2347, opened 2021-02-02
Issue2059 refactor vav reheat example

bravache:
Still to do before merging:

- [x] Update the expected unittest outputs

I am leaving the old output intacts so that the reviewer might see the impact of those changes.

Summary of the changes:

In `VAVBranch`:
- Class was renamed to `VAVReheatBox` to more explicitly describe its components and function
- Heat exchanger ports were switched so that `*1` is water and `*2` is air, per convention. 
- A new parameter `ratVFloHea` is defined to characterize the ratio of supply air flow rate in heating mode, used for sizing of the reheat coil
- Several constant parameter were defined, so that equations would be more explicit. _Question: Did I use the right type of variables by using `final parameter`. Also, is there a better way to define the water and air cp? I assume that accessing the value from the medium might create some problem by making the sizing variable in time._
- Water source and sink were moved out of the class and replaced by connector ports.

In `PartialOpenLoop`:
- Adding individual source to each VAVBox and connecting all hot water outlets to the same sink. _Question: Does it make sense to use different source for each coil? I wanted to avoid the extra complexity of modeling a full hot water loop with valve and controllers, but there might be another accessible solution that I am missing._


Buildings/Fluid/HeatExchangers/BaseClasses/PartialEffectivenessNTU.mo
```diff
@@ -103,8 +103,8 @@ initial equation
   Q_flow_nominal = m1_flow_nominal*cp1_nominal*(T_a1_nominal - T_b1_nominal);
   if use_Q_flow_nominal then
     Q_flow_nominal = -m2_flow_nominal*cp2_nominal*(T_a2_nominal - T_b2_nominal);
-    eps_nominal = abs(Q_flow_nominal/((T_a1_nominal - T_a2_nominal)*
-      CMin_flow_nominal));
+    eps_nominal = Q_flow_nominal/((T_a1_nominal - T_a2_nominal)*
```
mwetter, line 106: This file is from https://github.com/ibpsa/modelica-ibpsa
If you need changes, it needs to be fixed on the IBPSA library, please see https://github.com/ibpsa/modelica-ibpsa/wiki/Workflow-for-code-changes
This is then merged to Buildings using https://simulationresearch.lbl.gov/modelica/buildingspy/development.html#buildingspy.development.merger.IBPSA.merge

mwetter (commented):
Please see inline comment regarding upstream changes from IBPSA if needed.

Buildings/Examples/VAVReheat/BaseClasses/VAVBranch.mo
```diff
@@ -11,8 +11,20 @@ model VAVBranch "Supply branch of a VAV system"
...
+  parameter Modelica.SIunits.Temperature T_HW_nominal=355.35 "Reheat coil nominal inlet water temperature";
+
+  final parameter Modelica.SIunits.Temperature TAirInlNom = 281.65 "Inlet air nominal temperature";
+  final parameter Modelica.SIunits.Temperature TAirOutNom = 323.15 "Outlet air nominal temperature";
+  final parameter Modelica.SIunits.SpecificHeatCapacity CpAir = 1004 "Air specific heat capacity";
+  final parameter Modelica.SIunits.SpecificHeatCapacity CpWater = 4180 "Water specific heat capacity";
+  final parameter Modelica.SIunits.HeatFlowRate QFloNom = m_flow_nominal*ratVFloHea*CpAir*(TAirOutNom - TAirInlNom) "Nominal heat flow rate";
+  final parameter Modelica.SIunits.TemperatureDifference dTHWNom = 10 "Nominal hot water temperature delta";
+
+  final parameter Modelica.SIunits.MassFlowRate m_flow_HW_nominal = QFloNom/(CpWater * dTHWNom)  "Mass flow rate of hot water to reheat coil";
```
mwetter, line 26: Please use `*_flow_nominal` for heat/mass flow rates at design (nominal) condition.
Also, use lower case for `cpWater` (or 3 letter camel case as in `cpWat` where `c` is lower case because it is a variable name, and we use `cp` because that is a commonly used variable in the literature and hence commonly understood as opposed to `speHeaCap`)

Buildings/Examples/VAVReheat/BaseClasses/PartialOpenLoop.mo
```diff
@@ -265,36 +267,41 @@ partial model PartialOpenLoop
     redeclare package MediumW = MediumW,
     m_flow_nominal=mCor_flow_nominal,
     VRoo=VRooCor,
-    allowFlowReversal=allowFlowReversal)
+    allowFlowReversal=allowFlowReversal,
+    T_HW_nominal=THWSup)
```
mwetter, line 271: Better use something like `THotWat_nominal`

mwetter (commented):
Please also check why the number of nonlinear equations changed. You could probably avoid that by setting `from_dp` for the pressure drop elements (as I don't think structurally you changed the equations).
```
*** Warning: ASHRAE2006.mat: Translation statistics for initialization and results changed for nonlinear.
 Old = 5, 2, 5, 2, 5, 2, 5, 2, 2, 2, 0
 New = 5, 2, 5, 2, 5, 2, 5, 2, 3, 0, 1
```

bravache:
> Please also check why the number of nonlinear equations changed. You could probably avoid that by setting `from_dp` for the pressure drop elements (as I don't think structurally you changed the equations).
> 
> ```
> *** Warning: ASHRAE2006.mat: Translation statistics for initialization and results changed for nonlinear.
>  Old = 5, 2, 5, 2, 5, 2, 5, 2, 2, 2, 0
>  New = 5, 2, 5, 2, 5, 2, 5, 2, 3, 0, 1
> ```

After reading through the process of translation, I am still a bit unsure about what that statistic is representing, or more specifically how it has changed. I assume that equations have moved from one loop to another after I refactored the water source out of the `VAVBranch` class into the `PartialOpenLoop` class, but I would not be able to substantiate my argument.
Is it a problem that this stat has changed, given that the total number of non linear equations has remained unchanged? If so, can you elaborate on your solution, since I am not sure what setting the `from_dp` refers to.

bravache:
> Please also check why the number of nonlinear equations changed. You could probably avoid that by setting `from_dp` for the pressure drop elements (as I don't think structurally you changed the equations).
> 
> ```
> *** Warning: ASHRAE2006.mat: Translation statistics for initialization and results changed for nonlinear.
>  Old = 5, 2, 5, 2, 5, 2, 5, 2, 2, 2, 0
>  New = 5, 2, 5, 2, 5, 2, 5, 2, 3, 0, 1
> ```

@mwetter After reading through the process of translation (just FYI, I really liked [this article](https://www.claytex.com/tech-blog/model-translation-and-symbolic-manipulation/) if you ever have to point someone to it) , I am still a bit unsure about what that statistic is representing, or more specifically what caused it to change. I assume that equations have moved from one loop to another after I refactored the water source out of the `VAVBranch` class into the `PartialOpenLoop` class, but I would not be able to substantiate my argument.
Is it a problem that this stat has changed, given that the total number of non linear equations has remained unchanged? If so, can you elaborate on your solution, since I am not sure how the pressure drop elements affected this array.

mwetter:
@bravache : I generated for the new and the old model the dsmodel.mof file (Simulation -> Setup -> Translation). The new file has
```
    // Nonlinear system of equations
    // Tag: initialization.nonlinear[9]
      // It depends on the following parameters: 
      //   eco.damExh.m_flow_turbulent
      //   eco.damOut.m_flow_turbulent
      //   eco.damRet.m_flow_turbulent
      //   weaDat.pAtmSel.p
      // It depends on the following variables: 
      //   eco.damExh.k
      //   eco.damOut.k
      //   eco.damRet.k
      //   flo.portsSou[1].p
      //   heaCoi.port_b2.p
      // Start values for iteration variables of non-linear system of 3 equations: 
      //   amb.ports[2].m_flow(start = 0)
      //   dpRetDuc.dp(start = 0)
      //   amb.ports[1].m_flow(start = 0)
    algorithm // Torn part
```
whereas the old has
```
   // Nonlinear system of equations
    // Tag: initialization.nonlinear[9]
      // It depends on the following parameters: 
      //   cor.vav.coeff1
      //   cor.vav.kTotMax
      //   cor.vav.m_flow_turbulent
      //   eas.vav.coeff1
      //   eas.vav.kTotMax
      //   eas.vav.m_flow_turbulent
      //   fanSup.eff.pCur1.V_flow[1]
      //   fanSup.eff.pCur1.V_flow[2]
      //   fanSup.eff.pCur1.dp[1]
      //   fanSup.eff.pCur1.dp[2]
      //   fanSup.eff.pCur1.n
      //   fanSup.eff.preDer1[1]
      //   fanSup.eff.preDer1[2]
      //   nor.vav.coeff1
      //   nor.vav.kTotMax
      //   nor.vav.m_flow_turbulent
      //   sou.vav.coeff1
      //   sou.vav.kTotMax
      //   sou.vav.m_flow_turbulent
      //   wes.vav.coeff1
      //   wes.vav.kTotMax
      //   wes.vav.m_flow_turbulent
      // It depends on the following variables: 
      //   cor.vav.dp_min
      //   cor.vav.dp_x1
      //   cor.vav.m_flow_set
      //   eas.vav.dp_min
      //   eas.vav.dp_x1
      //   eas.vav.m_flow_set
      //   fanSup.rho_inlet.y
      //   fanSup.y_actual
      //   flo.portsSou[1].p
      //   heaCoi.port_b2.p
      //   nor.vav.dp_min
      //   nor.vav.dp_x1
      //   nor.vav.m_flow_set
      //   sou.vav.dp_min
      //   sou.vav.dp_x1
      //   sou.vav.m_flow_set
      //   wes.vav.dp_min
      //   wes.vav.dp_x1
      //   wes.vav.m_flow_set
      // Start values for iteration variables of non-linear system of 2 equations: 
      //   fanSup.VMachine_flow(start = 0)
      //   fanSup.port_b.p(start = fanSup.p_start)
    algorithm // Torn part
```
As you did not change this part of the model, I suspect that with the new model, the tearing algorithm (which requires some heuristics) ends up with a 3x3 rather than a 2x2 nonlinear equation. We can leave this as-is. It is only in the initial equation section and hence won't affect much the simulation time.

Can you please make sure that lines don't overlap with components (and inputs signals on the left of a component are connected from the left) and then we can finish this PR.

Another optional change we could do to clean up the model is to move the discharge air temperature and air flow rate sensor into the terminal box (as they are usually output of a terminal box). This would make the terminal box more usable for other models and clean up the system model.

AntoineGautier:
In `Buildings.Examples.VAVReheat.BaseClasses` the equation below needs to be updated.
```
Results res(
  final A=ATot,
  PFan=fanSup.P + 0,
  PHea=heaCoi.Q2_flow + cor.terHea.Q1_flow + nor.terHea.Q1_flow + wes.terHea.Q1_flow
       + eas.terHea.Q1_flow + sou.terHea.Q1_flow,
```

bravache:
> In `Buildings.Examples.VAVReheat.BaseClasses.PartialOpenLoop` the equation below for `PHea` needs to be updated.
> 
> ```
> Results res(
>   final A=ATot,
>   PFan=fanSup.P + 0,
>   PHea=heaCoi.Q2_flow + cor.terHea.Q1_flow + nor.terHea.Q1_flow + wes.terHea.Q1_flow
>        + eas.terHea.Q1_flow + sou.terHea.Q1_flow,
> ```
> 
> This is addressed in [9f78546](https://github.com/lbl-srg/modelica-buildings/commit/9f78546ff0ebfaf485ba0c1747e77c3a543bfb5b).

Good catch! Thanks

Buildings/package.mo
```diff
@@ -445,6 +445,13 @@ have been <b style=\"color:blue\">improved</b> in a
                        This is for
                        <a href=\"https://github.com/lbl-srg/modelica-buildings/issues/1502\">issue 1502</a>.
     </td>
+    <tr><td valign=\"top\">Buildings.Examples.VAVReheat.BaseClasses.VAVBranch
```
AntoineGautier, line 448: - [ ] This is non-backward compatible, to be moved to `<!-- Non-backward compatible changes to existing components -->`.
- [ ] Also a conversion script command must be added for Dymola users in `Buildings/Resources/Scripts/Dymola/ConvertBuildings_from_7_to_8.0.0.mos`.
- [ ] A note for `Buildings.Examples.VAVReheat.BaseClasses.PartialOpenLoop` is perhaps also useful.

AntoineGautier (changes requested):

AntoineGautier:
@bravache Please see my review comments. Also, happy to discuss the changes I already committed.

For reference I kept track of the impact of those changes at https://github.com/lbl-srg/modelica-buildings/issues/2059#issuecomment-774399581.

Buildings/package.mo
bravache: Done.

bravache (commented):
@AntoineGautier Let's talk about conversion scripts and the new validation test tomorrow if you have some time. I am not sure if I did the right thing there, but I am not sure how to test it.

AntoineGautier (approved):
@mwetter Please merge if the changes seem fine to you.
(I reordered alphabetically the release notes in the top-level package as part of this PR.)

bravache:
@mwetter FYI, this is ready to merge.

Buildings/Air/Systems/SingleZone/VAV/ChillerDXHeatingEconomizer.mo
```diff
@@ -627,6 +627,10 @@ feedback control of damper positions. The cooling coil is a dry coil model.
 </html>", revisions="<html>
 <ul>
 <li>
+February 25, 2021, by Baptiste Ravache:<br/>
+Inverse the sign of cooCoi.Q_flow_nominal to respect the heat flow convention.
```
mwetter, line 631: Use `<code>...</code>` when referring to variables.

Buildings/Examples/VAVReheat/BaseClasses/PartialOpenLoop.mo
```diff
@@ -543,13 +493,19 @@ partial model PartialOpenLoop
...
     PFan=fanSup.P + 0,
-    PHea=heaCoi.Q2_flow + cor.terHea.Q1_flow + nor.terHea.Q1_flow + wes.terHea.Q1_flow
-         + eas.terHea.Q1_flow + sou.terHea.Q1_flow,
+    PHea=heaCoi.Q2_flow + cor.terHea.Q2_flow + nor.terHea.Q2_flow + wes.terHea.Q2_flow
+         + eas.terHea.Q2_flow + sou.terHea.Q2_flow,
     PCooSen=cooCoi.QSen2_flow,
     PCooLat=cooCoi.QLat2_flow) "Results of the simulation";
   /*fanRet*/
 
 protected
```
mwetter, line 502: List all protected members after the public members.

Buildings/Examples/VAVReheat/BaseClasses/PartialOpenLoop.mo
```diff
@@ -926,6 +957,12 @@ shading devices, Technical Report, Oct. 17, 2006.
 </html>", revisions="<html>
 <ul>
 <li>
+February 03, 2021, by Baptiste Ravache:<br/>
+Refactored the sizing of the heating coil in the VAVBranch (renamed VAVReheatBox) class.<br/>
```
mwetter, line 961: Use `<code>...</code>`.

Buildings/Examples/VAVReheat/BaseClasses/VAVReheatBox.mo
```diff
@@ -0,0 +1,276 @@
...
+Model for a VAV terminal box with a water reheat coil and pressure independent damper.
+The pressure independent damper model includes an idealized flow rate controller
+and requires a discharge air flow rate set-point (normalized to the nominal value)
+as a control signal.
+</p>
+</html>", revisions="<html>
+<ul>
+<li>
+February 12, 2021, by Baptiste Ravache:<br/>
+First implementation, based on <code>Buildings.Obsolete.Examples.VAVReheat.BaseClasses.VAVBranch</code><br/>
```
mwetter, line 270: Use a hyperlink, e.g.,
```
<a href=\"modelica://Buildings.Obsolete.Examples.VAVReheat.BaseClasses.VAVBranch\">
Buildings.Obsolete.Examples.VAVReheat.BaseClasses.VAVBranch</a>
```
This way, the correct name is checked when exporting the html documentation, and users can easier navigate to it.

Buildings/Obsolete/Examples/VAVReheat/BaseClasses/VAVBranch.mo
```diff
@@ -221,5 +221,19 @@ The pressure independent damper model includes an idealized flow rate controller
 and requires a discharge air flow rate set-point (normalized to the nominal value) 
 as a control signal.
 </p>
+</html>", revisions="<html>
+<ul>
+<li>
+February 25, 2021, by Baptiste Ravache:<br/>
+Inverse the sign of terHea.Q_flow_nominal to respect the heat flow convention.
```
mwetter, line 228: Use `<code>...</code>`

mwetter (changes requested):
Please see inline comments.
Afterwards it is ready to be merged (once the conflicts are resolved).

mwetter (approved):

mwetter (changes requested):
@bravache and @AntoineGautier : I merged to master branch and rerun the examples but have not yet added the new reference results. The reason is that in `Guideline36Winter`, the heating energy consumption `res.EHea` changed from 5.32E6 J (master) to 4.47E6 J (this branch). This is a 16% reduction. Is this intentional in this refactoring?

Note that as I updated from E+ 9.4 to E+ 9.5 and at the same time, refactored the Spawn coupling, both of which are now on the master, this quantity in this file also changed, but only from 5.23E6 J (commit https://github.com/lbl-srg/modelica-buildings/commits/25407) to  5.32E6 J (master), or 2%. Even the 2% seems a bit high for an E+ version change. 

I did compare `NoHVAC` which uses the same building model. For this simpler model, the room air temperatures are quite similar between commit https://github.com/lbl-srg/modelica-buildings/commits/25407 and the master. If we cannot explain the 16% reduction, a next step may be to make a new model similar to `NoHVAC`, but add at each zone inlet an ideal heater that tracks the heating set point. This will help isolating whether the differences are from the E+ envelope model or from the VAV model.

Another cause of the differences may be the new Spawn coupling (the C interface). But I doubt this to be the case because in `OneZoneWithControl`, which uses a different idf file, the heat input into the zone `hea.Q_flow` is practically identical between the master and https://github.com/lbl-srg/modelica-buildings/commits/25407

bravache:
Thanks @mwetter, I'll look into it. Do you know if that discrepancy is only for this test case?

AntoineGautier:
This is unexpected. The impact on the annual heating energy use has been assessed at https://github.com/lbl-srg/modelica-buildings/issues/2059#issuecomment-774399581 for `Buildings.Examples.VAVReheat.Guideline36` and is around 2%. However, now that I look at it again, the impact on `Buildings.Examples.VAVReheat.ASHRAE2006` is about 7% which is perhaps a little higher than expected.

mwetter:
Let me add a test case through https://github.com/lbl-srg/modelica-buildings/issues/2381 to see if the changes come from EnergyPlus or the Modelica-EnergyPlus coupling.

mwetter:
@bravache @AntoineGautier : The change is not likely from E+. The test with an ideal heating/cooling system for the small office building only changes the heating energy by 2% for these 5 winter days, see https://github.com/lbl-srg/modelica-buildings/issues/2381#issuecomment-791039971

Buildings/Resources/src/ThermalZones/EnergyPlus/install.py
```diff
@@ -13,29 +13,34 @@
 import urllib.request, urllib.parse, urllib.error
```
bravache, line 13: Sorry, this file was edited because I have a file watcher that format all `.py` files, but it's all just white spaces and line breaks.

Buildings/Examples/VAVReheat/Guideline36.mo
```diff
@@ -26,39 +26,39 @@ model Guideline36
...
+    VDisHeaSetMax_flow=ratVFloHea*mNor_flow_nominal/1.2)
     "Controller for terminal unit north"
     annotation (Placement(transformation(extent={{1038,84},{1058,104}})));
   Buildings.Controls.OBC.ASHRAE.G36_PR1.TerminalUnits.Controller conVAVWes(
     V_flow_nominal=mWes_flow_nominal/1.2,
     AFlo=AFloWes,
     final samplePeriod=samplePeriod,
     VDisSetMin_flow=max(1.5*VWesOA_flow_nominal, 0.15*mWes_flow_nominal/1.2),
-    VDisHeaSetMax_flow=ratVFloHea*VWesOA_flow_nominal)
+    VDisHeaSetMax_flow=ratVFloHea*mWes_flow_nominal/1.2)
```
bravache, line 61: These were the issues that caused the discrepancies between `master` and this branch. These parameters were set to `ratVFloHea*VWesOA_flow_nominal`, when originally, they were set to `0.3*V_flow_nominal`. I think the intention of that line change was to reuse the `ratVFloHea` variable, but the wrong flow rate was used (outside air instead of mixed), which caused the controller to limit the amount of air going into the spaces.

bravache: @mwetter I'll let you take another look at the test results, but this does fix the main gap you pointed out. I can update the reference results in the morning.

mwetter:
The reason for the change in the results is that `terHea.Q_flow_nominal` is now sized differently, leading to smaller reheat coils. Also, the fan air flow rate is higher (the fan is sized the same in both versions). This in turn leads to higher outside air flow rates at the economizer, and therefore the heating energy is higher.
image

mwetter (approved):

PR #2349, opened 2021-02-03
Issue2348 documentation update

mwetter:
This closes #2384.
The release date will be updated when I prepare the release.


AntoineGautier (approved):
Not related to this PR, but I noticed that this part of the documentation may be updated for clarity as `idfName` is not exposed through the thermal zone but rather through the outer building component.
https://github.com/lbl-srg/modelica-buildings/blob/994119b1b5962cb572f5825c3712aba07c029de6/Buildings/ThermalZones/EnergyPlus/ThermalZone.mo#L376-L385

mwetter:
@AntoineGautier : Thanks, I will make a separate patch regarding the idfName.

PR #2350, opened 2021-02-03
Correct documentation of hysteresis in comparison blocks

### LINKED ISSUE #2246: Correct documentation of comparison with hysteresis blocks
In `Buildings.Controls.OBC.CDL.Continuous.Greater` this part of the documentation
`If h > 0, then the output switches to true if u1 > u2, and it switches to false if u1 > u2-h.`
should be 
`If h > 0, then the output switches to true if u1 > u2, and it switches to false if u1 < u2-h.`
The same applies to the other comparison blocks.
Also I wonder if `If h > 0` should not be deleted as anyway the parameter cannot take negative values due to `parameter Real h(final min=0)=0`.

AntoineGautier:
This closes #2246.

### DIFF
--- a/Buildings/Controls/OBC/CDL/Continuous/Greater.mo
+++ b/Buildings/Controls/OBC/CDL/Continuous/Greater.mo
@@ -162,9 +162,10 @@ Block that outputs <code>true</code> if the Real input <code>u1</code>
 is greater than the Real input <code>u2</code>, optionally within a hysteresis <code>h</code>.
 </p>
 <p>
-The parameter <code>h</code> is used to specify a hysteresis.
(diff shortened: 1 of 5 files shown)


AntoineGautier:
CI failing due to
```
Error response from daemon: login attempt to https://registry-1.docker.io/v2/ failed with status: 401 Unauthorized
The command "echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin" failed and exited with 1 during .
```

mwetter:
@AntoineGautier : I addressed the case for `h=0` and will merge when all tests pass. I think the login error is a temporary failure.

mwetter (approved):

PR #2357, opened 2021-02-09
Clarfied documentation for where idf file is obtained from

mwetter:
This clarifies where the idf name has to be specified for Spawn.

### DIFF
--- a/Buildings/ThermalZones/EnergyPlus/Building.mo
+++ b/Buildings/ThermalZones/EnergyPlus/Building.mo
@@ -109,8 +109,9 @@ to specify building-level parameters.",
 Model that declares building-level specifications for Spawn of EnergyPlus.
 </p>
 <p>
-Each EnergyPlus idf file must have one instance of this model, which
-is used to configure EnergyPlus.
+This model is used to configure EnergyPlus.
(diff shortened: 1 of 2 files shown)


AntoineGautier (approved):
I added a precision regarding the instance name `building`, see https://github.com/lbl-srg/modelica-buildings/pull/2357/commits/867e292b092e0b386198914ce70c2cac4afee84d.
I hope that it is correct. 
I also wonder if this requirement could not be verified using `getInstanceName` and a string matching condition.
There is one last thing that is not crystal clear to me in the documentation: do *all* zones in the IDF require an instance of `ThermalZone` or is it possible to only model a subset of the IDF zones in Modelica?

mwetter:
This is ready to merge after the tests pass.

@AntoineGautier : Please note https://github.com/lbl-srg/modelica-buildings/pull/2357/commits/6c8964c32992c14c435149ff2b3a8ab359235410. E.g., the same `idfName` can be used multiple times, for example if two buildings are identical. The example  `Buildings.ThermalZones.EnergyPlus.Validation.MultipleBuildings.TwoIdenticalOneZoneBuildings` illustrates this.

PR #2370, opened 2021-02-23
Fix wrong heatexchanger convention

bravache:
This PR is in response to:
https://github.com/ibpsa/modelica-ibpsa/pull/1445#discussion_r572515071

There was no issue with the regression tests in IBPSA, but some of the tests in MBL were using the wrong convention, and this PR addresses that. This is to be merged after https://github.com/ibpsa/modelica-ibpsa/pull/1445 is complete and merged into the MBL.


AntoineGautier (approved):
Maybe add revision notes to models that already have a revision section.

PR #2384, opened 2021-03-07
Cleaned up code

mwetter:
This merges some cleanup for Spawn to the master.

### DIFF
--- a/Buildings/ThermalZones/EnergyPlus/BaseClasses/SpawnExternalObject.mo
+++ b/Buildings/ThermalZones/EnergyPlus/BaseClasses/SpawnExternalObject.mo
@@ -1,7 +1,7 @@
 within Buildings.ThermalZones.EnergyPlus.BaseClasses;
 class SpawnExternalObject
   "Class used to couple the FMU to interact with a thermal zone"
-  //extends Modelica.Icons.BasesPackage;
+  extends Modelica.Icons.BasesPackage;
   extends ExternalObject;
   function constructor
     "Construct to connect to a thermal zone in EnergyPlus"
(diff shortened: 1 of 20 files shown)


Buildings/ThermalZones/EnergyPlus/BaseClasses/SpawnExternalObject.mo
```diff
@@ -1,7 +1,7 @@
 within Buildings.ThermalZones.EnergyPlus.BaseClasses;
 class SpawnExternalObject
   "Class used to couple the FMU to interact with a thermal zone"
-  //extends Modelica.Icons.BasesPackage;
+  extends Modelica.Icons.BasesPackage;
```
perost, line 4: This didn't work in OpenModelica since the specification says that classes extending from ExternalObject are not allowed to contain any other elements, and we interpreted that in the strictest sense (extends clauses are technically elements). But this seems like something that should be allowed, so I've relaxed our check to allow extends clauses as long as they don't contain any components or classes.

mwetter: Thanks for the feedback, I will remove the `extends` through https://github.com/lbl-srg/modelica-buildings/pull/2386

perost: My comment was mostly just to tell you that you don't have to change anything though, but if you wish to remove the extends anyway you're of course free to do so.

mwetter: I understand, and I rather stay compliant with the standard as this `extends` is not important.

PR #2364, opened 2021-02-11
Wet coil epsilon-NTU model

AntoineGautier:
The feature branch was created based on what is described in https://github.com/lbl-srg/modelica-buildings/pull/2001#issuecomment-776349084.

TODO before merging:

- [x] @kim1077 to include (manually) recent local modifications (posterior to 456406e) and UA nominal computation.
  - Included as `Buildings.Fluid.HeatExchangers.WetCoilEffectivenessNTUWithRated`. @AntoineGautier to refactor within `WetCoilEffectivenessNTU`.
- [x] Resolve warning 
```
The initial conditions are not fully specified. Dymola has selected default initial conditions.
Assuming fixed start value for the continuous states: hexWetNTU.dryWetCalcs.fullywet.TWatOutHat(start = 283.15)
```

PR #2417, opened 2021-03-26
Added dryFra as a public variable

mwetter:
@AntoineGautier : Would this address https://github.com/lbl-srg/modelica-buildings/pull/2364#discussion_r602607111 ? If it causes an event, we could try adding `noEvent()`. Please merge if this addresses it.


AntoineGautier:
That works for me.
I merged the branch locally and updated the mos script.
This is now on the main feature branch after b3b9759.
We can close this PR and delete the branch.

PR #2429, opened 2021-04-01
Fix use of `each` in PartialSource

### LINKED ISSUE #2428: Wrong use of `each` in Fluid.Sources.BaseClasses.PartialSource
PartialSource contains the declaration:
https://github.com/lbl-srg/modelica-buildings/blob/947abac770e5265eefdd5e8b37c88cb13bc10e55/Buildings/Fluid/Sources/BaseClasses/PartialSource.mo#L21-L28
`m_flow` is not an array though, so using `each` to modify it is incorrect. The `each` prefix should instead be moved outside such that it applies to the `ports` array.

OpenModelica only issues a warning for this so models using PartialSource still works, but it can result in quite a lot of warnings.

perost:
Fixes #2428


mwetter:
@perost : Thanks, I will fix it on the IBPSA repository and on Buildings master and the maintenance branch.

PR #2472, opened 2021-04-30
Issue2471 access to component not in constraining type

### LINKED ISSUE #2471: Avoid access to component that are not present in constraining type of declaration
Some models access components that are not in the constraining type of the declaration. In OCT, this leads to a warning such as
```
Access to component minT not recommended, it is not present in constraining type of declaration
```
This issue is to correct these models.

mwetter:
This closes #2471

@AntoineGautier : Please note the change to Experimental/DHC/EnergyTransferStations/Combined/Generation5/ChillerBorefield.mo and let me know if this is problematic. (The current version accessed a control signal that was not in the constraining type of the redeclaration).

@bravache : With regard to your refactor for the `scale_eplus_floor`  please note the changes to the VAVReheat models. Essentially, the floor area is now declared in `PartialFloor`, and access to `flo.cor.AFlo` are replaced with `flo.AFloCor` (and the same for `sou`, `nor`, ...), see for example the changes in `Buildings/Examples/VAVReheat/BaseClasses/PartialOpenLoop.mo`. This was needed because the replaceable class does not contain `flo.cor`.


AntoineGautier:
@mwetter Changing the constraining type should not be problematic here.

PR #2474, opened 2021-04-30
Issue2446 fix data centers - number of continuous time states

khinkelman:
Hi @AntoineGautier, here are my suggested changes after our discussion. 

1. I removed 4 junctions where they are not necessary to maintain correct mixing locations. However, I feel most of the newly added junctions are necessary to not create false mixing locations. For example, on both the condenser water and chilled water sides, some (if not most) plants have bypass loops, which would create incorrect mixing when many of these subsystem components are connected to the outside.
2. The above change slightly reduced the number of continuous time states, but not significantly. Thus, I changed all of the junctions to have steady state dynamics by default, thus eliminating the mixing volume. This (as I understand it) will maintain the correctly intended mixing designs, but does not increase the number of continuous time states.

I've summarized my results below. The column highlighted in gray (version that is on this PR) shows that the number of states did not increase from the current version on the master branch that does not have any junction models included (only ideal mixing). 

image

Please let me know what you think of this approach or if you have any questions. Thanks again for your insightful discussion and help with this issue!


AntoineGautier:
I tested on `Buildings.Examples.Tutorial.Boiler.System7` (simplified with no three-way valve on the primary side to reproduce the example at https://simulationresearch.lbl.gov/modelica/userGuide/bestPractice.html#modeling-of-fluid-junctions): 

- keeping the junctions with the mixing volume,
- keeping the junctions but without the mixing volume (steady-state),
- suppressing the junctions.

And simulated the model with Dymola and OCT, see plot attached for the temperature in the bypass branch of the three-way valve.
System7

It seems that the steady-state junctions are effective to guard against the incorrect mixing point.
I believe this is due to the `inStream` operator in the equations below from `Buildings.Fluid.BaseClasses.PartialResistance` that prevents the compiler from considering `port_a.h_outflow` and `port_b.h_outflow` as alias variables (which would result in the three outside connectors of the steady-state junction model having the same enthalpy and an incorrect mixing point being introduced).

```
  // Isenthalpic state transformation (no storage and no loss of energy)
  port_a.h_outflow = if allowFlowReversal then inStream(port_b.h_outflow) else Medium.h_default;
  port_b.h_outflow = inStream(port_a.h_outflow);
```

So I agree with the changes you made and the configuration of the newly added junctions in steady-state to avoid the overhead.
@mwetter Your validation would also be useful.

mwetter:
@AntoineGautier @khinkelman : I agree that the steady-state configuration works: The variable `port_internal.h_outflow`, although it is `protected`, is still computed and therefore the junctions cannot be removed or collapsed into one junction point during translation.

PR #2487, opened 2021-05-10
Reviewing Spawn Documentation

bravache:
I updated the documentation with a few suggestions that might help clarifying a few sections. Some of those changes are open to interpretation so don't hesitate to push back.


Buildings/ThermalZones/EnergyPlus/UsersGuide.mo
```diff
@@ -33,23 +33,20 @@ The <code>.epw</code> file will be used by the EnergyPlus envelope model, and th
 file will be used by the Modelica model, and must be specified by the parameter <code>weaName</code>
 in the instance <code>building</code>.
 </li>
+</ol>
+The following coupling objects can then be integrated in the model that contains the instance
```
bravache, line 37: I found peculiar that the instantiation of `ThermalZone` was a non-optional bullet, whereas any other coupling objects was optional. Isn't it technically possible to run Spawn without any `ThermalZone`? This is what I have been doing in the `ZoneSurface` comparison exercise.

I switched this section so that all the coupling objects are presented on the same level (although the ThermalZone is first because it is likely the most useful), without suggesting that some are optional and some are not.

Buildings/ThermalZones/EnergyPlus/UsersGuide.mo
```diff
@@ -33,23 +33,20 @@ The <code>.epw</code> file will be used by the EnergyPlus envelope model, and th
...
 in the instance <code>building</code>.
 </li>
+</ol>
+The following coupling objects can then be integrated in the model that contains the instance 
+<code>building</code>, or in any model instantiated by that model
+<ul>
 <li>
-In the model that contains the instance <code>building</code>,
-or in any model instantiated by that model, instantiate
-for each thermal zone an instance of
```
bravache, line 39: I was originally confused by "for each thermal zone", since it might suggest to some reader that you are required to instantiate a `ThermalZone` model for all the thermal zones in the IDF, when in fact you only need to do so for zones that you want to control (with the caveat that the other zones will be free floating).

Buildings/ThermalZones/EnergyPlus/UsersGuide.mo
```diff
@@ -168,8 +165,7 @@ If a zone is in the idf file but not modeled in Modelica, EnergyPlus will
 simulate the zone as free floating.
 </li>
 <li>
-If there is an HVAC system in the idf file, then EnergyPlus will remove it.
-</li>
+All EnergyPlus HVAC objects that are present in the idf file are removed when coupled to Spawn</li>
```
bravache, line 168: I think this change is very small, but in the original sentence, I did not agree with the "EnergyPlus will remove it", since that removal is done when compiling the model, independently of EnergyPlus.

PR #2502, opened 2021-05-24
fix enthalpy typo

carlesRT:
harmless change. typo fix


mwetter:
Thanks for the correction of the typo.

PR #2530, opened 2021-06-08
Issue2264 generic cooling

JingWang-CUB:
Hi @AntoineGautier , we'd like to start the model review process for merging the generic district cooling models into the Master branch though this PR. In this PR, we included the cooling plant model as well as cooling tower subsystem models and a few control blocks. Each model has an example model to validate the results. Unit tests have been run with Dymola and JModelica. 

There are a few items we wanted to highlight in particular:

1. Since the fluid ports in `PartialPlant` are conditional, we cannot conditionally calculate the thermodynamic states in the plant models (e.g., `Medium.ThermodynamicState sta_a(...) if show_T`). If we want the plant models to have these capabilities, `show_T` and the conditional `sta*_a/b` need to be added to the `PartialPlant` base class. (The error for reference appeared during unit tests and was of the form `The component port_aSerCoo is conditional: Access of conditional components is only valid in connect statements`);
2. We modified the base class `Buildings.Fluid.HeatExchangers.CoolingTowers.Examples.BaseClasses.PartialStaticTwoPortCoolingTower`  to allow it to be reused for the cooling tower subsystem example models. Specifically, we changed the constraining declaration:

 ```
 Buildings.Fluid.HeatExchangers.CoolingTowers.BaseClasses.CoolingTower tow     
 constrainedby
 Buildings.Fluid.HeatExchangers.CoolingTowers.BaseClasses.CoolingTower
 ```
 
 to be simply `Buildings.Fluid.Interfaces.PartialTwoPortInterface tow`. This saved many lines of code by allowing reuse of the base class. However, if you prefer a different solution, please let us know. Thank you very much!

Looping in other related team members here @ChengnanShi-Work @khinkelman .


AntoineGautier:
@JingWang-CUB Thanks for all the updates.
I will merge your PR readily so that CI tests run. I will plan the review internally (on the lbl-srg branch) and let you know when it can be done.
Regarding the couple items you mention.

1. Indeed, conditional components cannot be used outside of connect clauses. That being said, I don't follow your suggestion to add `sta*_a/b` to the `PartialPlant` base class. We would have the same problem here since the equation involves variables from the conditional fluid connectors, which is not allowed. I see no alternative but using sensors in the derived classes: those can be connected to the fluid ports, which complies with the specification. If you use them for diagnostics only, you may configure them in steady-state (`tau=0`) which should not yield significant overhead compared to the `sta*_a/b` equations. However, if you use them within control loops, better keep the state (or perform in-depth analysis before removing it), but in this case, `sta*_a/b` should not be used either.
2. We had a few discussions here in the past about the pedantic syntax with the constraining type identical to the declaration type. I could not come with a minimum example illustrating why it should be preferred to the simplified syntax that you suggest (where the constraining type is omitted and defaults to the declaration type). My vague conclusion is that it helps tools checking models against the complex requirements from the specification, for instance when modifying the constraining type in a replaceable redeclaration:

    > The constraining type of a replaceable redeclaration must be a subtype of the constraining type of the declaration it redeclares. In an element modification of a replaceable element, the modifications are applied both to the actual type and to the constraining type. In an element-redeclaration of a replaceable element the modifiers of the replaced constraining type are merged to both the new declaration and to the new constraining type, using the normal rules where outer modifiers override inner modifiers.

    I believe the corresponding syntax check will vary across tools, and even more if the constraining type is not explicit in the original declaration. So I believe the pedantic syntax is used mainly for safety and usability across tools.

Issue #2544, opened 2021-06-22
Make the connection between zones and zone groups


bravache:
The current implementation of Guideline 36 in the Modelica Buildings Library introduces the concept of zones and zone group. A building may have multiple zones and zone groups, and each controlled zones is part of one zone group that dictates how to control the associated AHU.

The objective of this feature is to provide a way to parse and connect the output signal of zones to their respective group, given a mapping between all the zones and groups.

PR #2548, opened 2021-06-23
Issue2544 ZoneGroup

bravache:
This PR is addressing issue #2544 .

## Constraints
The constraints when addressing this issue was to not rely on `for` loops when writing equation or connecting blocks together. 

Another constraints is that the zone groups are not equally sized, and so it is not possible to create a singular class that could do all the routing between zones and groups since it would require to have the output connectors be non-rectangular 2-dimensional arrays.

## Proposed solution
The solution I propose in this PR is to duplicate the signal of the zones by the number of groups, so that each iteration can be connected to a separate group, and then have that group filter only the zones that are present in it. This is a bit clunky, and there might be a better solution, but from separate test, this seems to do the job.

I've introduced two blocks for routing: `ZoneStatusDuplicator` and `ZoneGroupFilter` which are used in conjunction when parsing the zone signals. These can't be wrapped in the same class since the duplicator is used as a single object whereas the filter is used in an array of size `nGro`. 

## Potential improvements

Note that the logic of the filter could be integrated in `GroupStatus`, and the logic of the duplicator could potentially be reduced to just duplicating the bus if that's possible. This depends on feedback on the current structure or if a better solution is envisioned. 

## What it looks like in Guideline36

image


bravache:
Changes were moved to: https://github.com/lbl-srg/modelica-buildings/pull/2557

PR #2559, opened 2021-06-30
Fix DSP reset logic used in VAVReheat ASHRAE2006 model

### LINKED ISSUE #2545: Fix DSP reset logic used in VAVReheat ASHRAE2006 model
In `Buildings.Examples.VAVReheat.Controls.DuctStaticPressureSetpoint` there is an hysteresis block disabling the duct static pressure reset logic if the outdoor air temperature is below 10°C (the set point being kept minimum). 
This part of the logic is not mentioned in ASHRAE 2006 and no valid rationale can be found.
In some cases, this may starve core zones that have a cooling demand in winter. So the zone set point is not met with the VAV damper fully open.

This issue must also track the impact on annual HVAC energy use of this change.

AntoineGautier:
This closes #2545.


bravache (approved):
This looks good. I was trying to find out if there are certain implementation of duct-static pressure reset that uses a low OAT lockout, but I did not find any relevant sources. 

I wonder why the lockout was there in the first place and whether it was put to address a specific issue.

AntoineGautier:
This is ready to merge.

mwetter (approved):

PR #2573, opened 2021-07-21
Add validation of ZoneSurface by comparing results to EnergyPlus

bravache:
This branch adds two validation models, one that runs a baseline EnergyPlus model of a single family house, and another model that runs the same model but after substituting some of its surfaces with `Building.ThermalZones.EnergyPlus.ZoneSurface` objects.

The zones are all free floating and the models report surface and garage zone air temperatures which are the most likely to deviate if there is a discrepancy.


bravache:
@mwetter This PR is ready to be merged.
Also, I apologize since there was a misunderstanding previously on whether or not this was ready, and it seems that it had been ready for a bit. It previously failed a Travis test, but this was outside of the scope of this PR, so it might have been a fluke with `master` changes.
Question: Does this need a top-level package revision description given that it is only for validation models?

Buildings/ThermalZones/EnergyPlus/Validation/SurfaceComparison/EnergyPlusSurfaces.mo
```diff
@@ -0,0 +1,70 @@
...
+    name="Surface Inside Face Temperature",
+    key="Garage:EastWall",
+    y(unit="K")) "East wall garage surface temperature"
+    annotation (Placement(transformation(extent={{-80,60},{-60,80}})));
+
+  annotation (
+    Documentation(
+      info="<html>
+<p>
+This model simulates the envelope of a single family house with EnergyPlus and is used
```
mwetter, line 47: This test outputs 5 temperatures at the top level, the plot shows 4 trajectories, 3 of which match each other quite well. Are you plotting the right 4 quantities, and why not plotting the 5th temperature? It would be good to describe what the user should expect from the result, e.g., which trajectories should be close to each other?

Maybe you could use an assertion to stop the simulation if the difference is larger than what you think is acceptable.

Buildings/ThermalZones/EnergyPlus/Validation/SurfaceComparison/ZoneSurfaces.mo
```diff
@@ -0,0 +1,105 @@
...
+    A=14.87) "East wall of garage" annotation (Placement(transformation(
+        extent={{-10,-10},{10,10}},
+        rotation=270,
+        origin={70,-50})));
+
+  annotation (
+    Documentation(
+      info="<html>
+<p>
+This model simulates the envelope of a single family house with EnergyPlus but uses
```
mwetter, line 78: Similar as above, wright what the expected result is.

mwetter (changes requested):
@bravache : Please see the inline comments.

Buildings/ThermalZones/EnergyPlus/Validation/SurfaceComparison/EnergyPlusSurfaces.mo
bravache: This tests looks at the air temperature of the garage and the surface temperature of two of its exterior walls and the interior wall towards the living room. There is one additional output being read from EnergyPlus and that is the surface temperature of the same interior wall but on the side of the living room. The garage was chosen since it is an unconditonned, free-floating zone. The living room side of the interior wall is read to be used as a boundary condition to the interior wall heat transfer in the `ZoneSurface` case, but it is a bit redundant to validate its value, since an issue with the interior wall would be reflected in the other side temperature.

The main objective of this test is to compare the value obtained with and without the spawn object `ZoneSurface` to replace the heat transfer calculation through the exterior and interior walls. So one test operates as the baseline (`EnergyPlusSurfaces`) and the other is the spawn surface object validation (`ZoneSurfaces`).

I've updated the *.mos so that both file are simulated at the same time and actually compared in the same plot. I've also moved the `EnergyPlusSurfaces` example model as a base class, renamed it `ReferenceSurfaces` and removed its *mos command. Does that structure make sense? Note that BuildingsPy does not save the reference results for the second `mat` file generated, but it's not a problem in this case since the second `mat` is the reference results.

bravache:
@mwetter This should pass all the tests and ready to merge.

Buildings/Resources/Scripts/Dymola/ThermalZones/EnergyPlus/Validation/SurfaceComparison/ZoneSurfaces.mos
```diff
@@ -0,0 +1,13 @@
+﻿simulateModel("Buildings.ThermalZones.EnergyPlus.Validation.SurfaceComparison.ZoneSurfaces", stopTime=2592000, method="Cvode", tolerance=1e-06, resultFile="ZoneSurfaces");
+simulateModel("Buildings.ThermalZones.EnergyPlus.Validation.SurfaceComparison.BaseClasses.ReferenceSurfaces", stopTime=2592000, method="Cvode", tolerance=1e-06, resultFile="ReferenceSurfaces");
```
mwetter, line 2: This won't work in our CI testing, as we can only have one `simulateModel` command. You can work around this by making a top-level model that instantiates the two models that you want to compare, as in `Buildings.ThermalZones.EnergyPlus.Validation.MultipleBuildings.ThreeZonesTwoBuildings`

Buildings/ThermalZones/EnergyPlus/Validation/SurfaceComparison/package.mo
```diff
@@ -0,0 +1,15 @@
+within Buildings.ThermalZones.EnergyPlus.Validation;
+package SurfaceComparison "Collection of models used to validate zone surface coupling against EnergyPlus"
+  annotation (Documentation(info="<html>
+<p>
+Package containing models used to compare ZoneSurface results against equivalent EnergyPlus surfaces
```
mwetter, line 5: Always use the full class name with `<a href=\"modelica://Buildings....`
Otherwise, if the class gets renamed, this string becomes out of date and we have no means to detected it (hyperlinks will automatically be updated by the rename scripts, and when generating html, Dymola reports invalid hyperlinks).

mwetter (changes requested):
Please see the inline comments.

Buildings/ThermalZones/EnergyPlus/Validation/SurfaceComparison/package.mo
bravache: Updated

Buildings/Resources/Scripts/Dymola/ThermalZones/EnergyPlus/Validation/SurfaceComparison/ZoneSurfaces.mos
bravache: Done. Combined both building under `Buildings.ThermalZones.EnergyPlus.Validation.SurfaceComparison.SurfaceComparison`.

bravache:
@mwetter Ready to merge. I found what was wrong with OpenModelica, which was a difference in how variability is determined at translation. I had to move the wall composition and layers to protected parameter, even though it was working when implicitly left as continuous components in Dymola.

mwetter (approved):

PR #2619, opened 2021-09-18
Remove parameter prefixes from record definition

casella:
This commit fixes #7839 on the maint_8.0.x branch


mwetter:
Merged through https://github.com/lbl-srg/modelica-buildings/pull/2631

PR #2622, opened 2021-09-20
Fix issue link

tbeu:


mwetter:
@tbeu : Thanks for the pull request. I will merge it through https://github.com/lbl-srg/modelica-buildings/pull/2623

tbeu:
Pushed again (as there was some syntax mistake in the previous one).

PR #2627, opened 2021-09-20
Included ModelicaUtilities head file

JayHuLBL:
This PR duplicated `ModelicaUtilities.h` file from MSL to `Resources/C-Sources/` and explicitly added `#include "ModelicaUtilities.h"` to `pythonWrapper.c`.

@casella Please check if this addresses the OMC issue.


casella:
It should. Please make sure this commit is applied to maint_7.0.x, maint_8.0.x, and master

mwetter (changes requested):
@JayHuLBL : This is not correct. All that should be needed is adding the include statement. The header file must be provided by the tool (e.g, OMC, Dymola, etc.) and should not be distributed with the library.
See for example https://github.com/lbl-srg/modelica-buildings/pull/2627/files#diff-a2dac360b46a2f0e2231adecfa10c535ffad0a8ae53c8a5905a7586edaf03c23R129 which points to a Dymola macro. That won't work in other tools.

JayHuLBL:
@mwetter Thanks. I will close this PR (and the other two) for now and make the changes accordingly.

casella:
> @JayHuLBL : This is not correct. All that should be needed is adding the include statement. The header file must be provided by the tool (e.g, OMC, Dymola, etc.) and should not be distributed with the library.

@mwetter, I'm not sure. The ModelicaServices library is meant to be a stub with the interfaces, and each tool is expected to provide its own implementation. Conversely, I thought that the files in Modelica/Resources should not be touched, as I see Dymola does. 

I've asked, see [#3867](https://github.com/modelica/ModelicaStandardLibrary/issues/3867)

@adrpo, @sjoelund, could we set up the makefile so that the C compiler can look in the appropriate place and find ModelicaUtilities.h when it finds either one of the above includes, so that it is not necessary to add it explicitly adding it to the library Resources directory?

sjoelund:
> @adrpo, @sjoelund, could we set up the makefile so that the C compiler can look in the appropriate place and find ModelicaUtilities.h when it finds either one of the above includes, so that it is not necessary to add it explicitly adding it to the library Resources directory?

That's already done. The file only needs to `#include "ModelicaUtilities.h"`; it doesn't need to be present in the repository unless you also want to compile some library in the same repository (this is why MSL includes the header - so it can provide pre-compiled binaries).

Issue #2639, opened 2021-09-21
Correct bug in `Experimental.DHC.CentralPlants.BaseClasses.PartialPlant`, refactor “if” statements used in fluid ports. 


mingzhe37:
Hi Dr. Hu @JayHuLBL , could you please open a new branch for me to fix this bug? Please let me know if you have any questions. Thanks!

**Describe the bug**
The partial model `Experimental.DHC.CentralPlants.BaseClasses.PartialPlant` has a bug with the fluid ports. This model should enable and disable fluid ports functionally and graphically according to the `typ` value, but this function has been found to be invalid.

**To Reproduce**
1. Open model `Buildings.Experimental.DHC.CentralPlants.Cooling.Examples.Plant`
2. Set parameter `typ=Buildings.Experimental.DHC.Types.DistrictSystemType.Cooling`
3. As seen in image below, only fluid ports `port_aSerCoo` and `port_bSerCoo` (the pair of fluid ports on the bottom) should be enabled, and only the line showing connection between them (the pair of red and dark blue line on the bottom) should be enabled as well. However, the results shows that all fluid ports and lines are enabled. 

**Expected behavior**
Only fluid ports `port_aSerCoo` and `port_bSerCoo` (the pair of fluid ports on the bottom) should be enabled, and only the line showing connection between them (the pair of red and dark blue line on the bottom) should be enabled. 

**Screenshots**
![image](https://user-images.githubusercontent.com/46808578/134255963-1614ab4c-eada-4a30-af13-e936b884c07b.png)

**Version**
 - Modelica Buildings Library: current master branch (update to commit 806aa91); this bug is related to commit 03be483
 - Modelica simulation environment: Dymola 2021/2021X/2022, OpenModelica 1.18.0
 - Operating system: Ubuntu 18.04.5 LTS (used with Dymola 2021), Windows 10 Pro (used with Dymola 2021/2021X/2022, OpenModelica 1.18.0)

**Additional context**
This error happens to all fluid ports. The error is caused by improper expression of the value assignment in the "if" statement, which located in `Experimental.DHC.CentralPlants.BaseClasses.PartialPlant` as follows:

![partial_model_1](https://user-images.githubusercontent.com/46808578/134255782-37171357-1686-4c53-a0a5-802c3daec416.png)

![partial_model_2](https://user-images.githubusercontent.com/46808578/134253321-fe67e740-cfc9-45df-976b-087c6595945c.png)

[...]

PR #2643, opened 2021-09-24
Corrected if expressions

JayHuLBL:
This closes issue #2639.

### DIFF
--- a/Buildings/Experimental/DHC/CentralPlants/BaseClasses/PartialPlant.mo
+++ b/Buildings/Experimental/DHC/CentralPlants/BaseClasses/PartialPlant.mo
@@ -44,7 +44,7 @@ partial model PartialPlant
     redeclare package Medium = Medium,
     m_flow(min=if allowFlowReversal then -Modelica.Constants.inf else 0),
     h_outflow(start=Medium.h_default, nominal=Medium.h_default)) if
-    have_ambientPort
+    have_serAmb
     "Fluid connector for ambient water service supply line"
     annotation (
       Placement(transformation(extent={{-310,30},{-290,50}}),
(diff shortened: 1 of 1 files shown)


Buildings/Experimental/DHC/CentralPlants/BaseClasses/PartialPlant.mo
```diff
@@ -129,6 +120,16 @@ partial model PartialPlant
...
+protected
+  final parameter Boolean have_heatingPort=typ <> Buildings.Experimental.DHC.Types.DistrictSystemType.Cooling and
+  typ <> Buildings.Experimental.DHC.Types.DistrictSystemType.CombinedGeneration5
+  "Boolean flag to enable fluid connectors for heating service line";
+  final parameter Boolean have_coolingPort=typ == Buildings.Experimental.DHC.Types.DistrictSystemType.CombinedGeneration1 or
+  typ == Buildings.Experimental.DHC.Types.DistrictSystemType.CombinedGeneration2to4 or
+  typ == Buildings.Experimental.DHC.Types.DistrictSystemType.Cooling
+  "Boolean flag to enable fluid connectors for cooling service line";
+  final parameter Boolean have_ambientPort=typ == Buildings.Experimental.DHC.Types.DistrictSystemType.CombinedGeneration5
+  "Boolean flag to enable fluid connector for ambient water service line";
```
mwetter, line 132: Is there a reason for deviating from the naming convention where 3 characters are used, which is also what is used for example in `Buildings.Experimental.DHC.EnergyTransferStations.Combined.Generation5.ChillerBorefield`.
Wouldn't a consistent nomenclature use `have_hea`, `have_coo` and `have_serAmb`, or is there a compelling reason for this departure of what is used in the other classes of the DHC package?

mwetter (commented):
See inline comment.

Buildings/Experimental/DHC/CentralPlants/BaseClasses/PartialPlant.mo
mingzhe37: Thanks for pointing this out! It is indeed better to unify these names to follow the convention to be consistent with other related cases, I will revise these!

PR #2693, opened 2021-10-22
Issue2668 fan efficiency euler

### LINKED ISSUE #2668: Fan efficiency with Euler number
The fan (mover) efficiency in the current version of `Buildings.Fluid.Movers` is treated as constant when power characteristics are not provided. This could be improved by using the Euler number to describe the efficiency (see [EnergyPlus 9.6.0 Engineering Reference](https://energyplus.net/assets/nrel_custom/pdfs/pdfs_v9.6.0/EngineeringReference.pdf) ch. 16.4 and [Haves et al., 2014](https://simulationresearch.lbl.gov/sites/all/files/phil_haves_-_development_of_diagnostic_and_measurement_and_verification_tools_for_commercial_buildings.pdf) sec. 2.3.2). This approach could provide the variability to describe the fan efficiency more accurately and avoid the power approaching zero when the flow rate approaches zero which is incorrect.

hcasperfu:
This implements a model that uses the only the operation condition with peak efficiency and a correlation with Euler number to calculate mover efficiency and power. 
The key functionality has been made but there are remaining issues. This draft pull request is meant to start the conversation while other features & issues are addressed. The figure below (generated by `.BaseClasses.Validation.EulerMethod`) compares efficiency results by `PowerCharacteristic`, `EulerMethod`, and a direct interpolation of calculated efficiency (=dp*V/P) by `Buildings.Utilities.Math.Functions.smoothInterpolation` as well as the power results by the former two methods.

EulerMethod

Issues:
1. While this method is intended to be used when perhaps only one operation point (where the efficiency is at its maximum) is available, the model currently requires the input of fan/pump curves. 
1.1. This is because the `FlowMachineInterface` model requires the curve input at many points. Further modifying this model can address this issue. 
1.2. To work around it, the validation model `.BaseClasses.Validation.EulerMethod` currently identifies the peak operation point from pump/fan curves. This model is intended to compare the results of the power computation methods of `PowerCharacteristic` and `EulerNumber`, but it could be useful in a more general sense when limited operation points are available and the Euler method can perhaps provide more accurate estimation than establishing a curve.
2. Preliminary results see high calculated power at low flow (see the lower figure above). This needs further investigation.
3. The release notes have not been updated yet.

This closes #2668.


hcasperfu:
To avoid very large computed power values at low flow, now the power directly uses extrapolated values when flow is below 10% maximum (instead of those calculated from efficiency).
Euler

hcasperfu:
I have made two major changes:

1. I have added two packages of fan performance data records: `.Data.Fans.Continental` and `.Data.Fans.Greenheck` (formerly we only had pump data). Between the two, using the Greenheck data is recommended. I am considering removing the Continental package, because its power values are not precise (e.g. having 0.1 BHP for all measurement points) and hence not very helpful. 
2. I have replaced the computed power at maximum flow (zero pressure rise) also with extrapolation. The problem with zero flow work (partially addressed in the previous comment) presents at both zero flow and max flow (zero pressure rise). This was not observed previously because the pump data used to validate the model did not provide a point with zero pressure. I have also revised the validation model to evaluate all points from zero flow to max flow. 

The function `.BaseClasses.Euler.computeCurves` computes nine support points of power values (from 10% to 90% of max flow) using `.BaseClasses.Euler.correlation` with Euler number, then extrapolates the points corresponding to zero flow and max flow. It then passes the results as `.BaseClasses.Euler.computedCurves` to `.BaseClasses.FlowMachineInterface`.

The two figures below compare the results before and after replacing the two ends with extrapolation. The data record used is `.Data.Fans.Greenheck.BIDW13`. 
No extrapolation :
EulerNoExtrap
Note that the computed power approaches zero instead of infinity (as shown in the head post) because of the none-zero expression *log(* [none-zero] / [none-zero] *)* put in place in `.BaseClasses.Euler.correlation` to prevent both *log(0)* and */0* when computing *&eta;*. In the expression *P = V&#x0307; &sdot; &Delta;p &frasl; &eta;*, the denominator no longer approaches zero faster than the nominator.

With extrapolation:
EulerYesExtrap

The figure above also sees that the Euler path overestimates the power at high flow. Underestimation was observed when using `.Data.Fans.Greenheck.BIDW18`.

The figure below compares the powers values (1) computed using the power characteristic path, (2) computed using the Euler number path without extrapolation, (3) given in data records, and (4) used as the support points for (2). 
image

It can be seen that the Euler path can lose track of the power trend fairly early at high flow and this error would be carried along with extrapolation (which is still better than letting the computed power approach zero or infinity).

hcasperfu:
So far I have been testing this model almost entirely based on the validation of `FlowMachineInterface` which does not consider the speed. I am trying to make a validation model out of `PartialFlowMachine` and obtained results that do not make sense. I am therefore marking this PR back to draft as I refine the model.

hcasperfu:
The efficiency and power computations are now handled via `CombiTable2D` as functions of both flow rate and pressure rise. The look-up tables range from 10% to 100% of max flow and pressure rise at 10% increments. Values below 10% of either flow or pressure are extrapolated. The extrapolation prevents the computed power from approaching infinity as efficiency approaches zero. (Technically, efficiency probably shouldn't be extrapolated and should be left as zero at zero flow or zero pressure. This change is a to-do.)
The figure below shows the results from `.BaseClasses.Validation.Euler` (based on `.BaseClasses.FlowMachineInterface`) which compares efficiency and power computations using `powerCharacteristic` and `eulerNumber` using the same set of mover data. 
image

Another validation model `.Validation.PowerEuler` is built. Similar to `.Validation.PowerExact` and `.Validation.PowerSimplified`, it compares the results of approximate power computation with the Euler number against the correct values. The figure below shows the approximation error. 
PowerEuler

hcasperfu:
> Technically, efficiency probably shouldn't be extrapolated and should be left as zero at zero flow or zero pressure. This change is a to-do.

This change is made in 9037bc3. The efficiency values are simply set to zero when the flow or pressure is zero in the look-up table.

hcasperfu:
I have made efforts to resolve the minor unit test result changes. Likely because my ubuntu is 20.04.3, my local test results are not able to agree with those generated by Travis CI for the following models:
- Job 4
  - `Buildings.Examples.Tutorial.CDL.System5` (ControlDescriptionLanguageSystem5.mat)
  - `Buildings.Examples.DualFanDualDuct.ClosedLoop` (DualFanDualDuctClosedLoop.mat)
- Job 25
  - `Buildings.Experimental.DHC.CentralPlants.Cooling.Subsystems.Examples.Examples.CoolingTowersParallel` (CoolingTowersParallel.mat)
  - `Buildings.Experimental.DHC.Examples.Combined.Generation5.Examples.ParallelConstantFlow` (ParallelConstantFlow.mat)
  - `Buildings.Experimental.DHC.Examples.Combined.Generation5.Examples.SeriesVariableFlow` (SeriesVariableFlow.mat)
  - `Buildings.Experimental.DHC.Loads.Examples.CouplingRCZ1Valve` (CouplingRCZ1Valve2.mat)
  - `Buildings.Experimental.DHC.Loads.Examples..Examples.CouplingRCZ6` (CouplingRCZ6.mat)
- Job 34
  - `Buildings.Fluid.HeatExchangers.CoolingTowers.Examples.YorkCalc` (YorkCalc.mat)
  - `Buildings.Fluid.HeatExchangers.CoolingTowers.Examples.FixedApproachDryBulb` (FixedApproachDryBulb.mat)
  - `Buildings.Fluid.HeatExchangers.CoolingTowers.Examples.Merkel` (Merkel.mat)
- Job 37
  - `Buildings.Fluid.Movers.Validation.PumpCurveDerivatives` (PumpCurveDerivatives.mat, updated again in a99b501)

hcasperfu:
In ed8c164, conflicts were resolved by accepting the results from #2786 for the following files:

- Buildings/Resources/ReferenceResults/Dymola/Buildings_Examples_DualFanDualDuct_ClosedLoop.txt
- Buildings/Resources/ReferenceResults/Dymola/Buildings_Experimental_DHC_CentralPlants_Cooling_Subsystems_Examples_CoolingTowersParallel.txt
- Buildings/Resources/ReferenceResults/Dymola/Buildings_Experimental_DHC_Examples_Combined_Generation5_Examples_ParallelConstantFlow.txt
- Buildings/Resources/ReferenceResults/Dymola/Buildings_Experimental_DHC_Examples_Combined_Generation5_Examples_SeriesConstantFlow.txt
- Buildings/Resources/ReferenceResults/Dymola/Buildings_Experimental_DHC_Examples_Combined_Generation5_Examples_SeriesVariableFlow.txt
- Buildings/Resources/ReferenceResults/Dymola/Buildings_Experimental_DHC_Loads_Examples_CouplingRCZ1Valve.txt
- Buildings/Resources/ReferenceResults/Dymola/Buildings_Experimental_DHC_Loads_Examples_CouplingRCZ6.txt

All of them are among those cannot be agreed between Travis and my local testing.

hcasperfu:
My local tests are now able to agree with Travis after d5a0117. 
I have reconciled most result files, except:
- Job .5: Optimica timeout with `Buildings.Examples.DualFanDualDuct.ClosedLoop`.
- Job .18 & .24: Cooling tower models in general seem to have a few degrees of difference in the results at some time steps. This difference is somewhat significant. I made some efforts in investigating the cause but I haven't found anything yet.

Buildings/Fluid/Movers/Data/Fans/EnglanderNorford1992/package.mo
```diff
@@ -0,0 +1,21 @@
+within Buildings.Fluid.Movers.Data.Fans;
```
mwetter, line 1: I suggest to move this to `Buildings.Fluid.Movers.Examples.Data` as it is only used there and I don't think users will see it useful to use this particular fan. If you think the data are generally useful, then let's leave them here.

Buildings/Fluid/Movers/Data/Fans/EnglanderNorford1992/Return.mo
```diff
@@ -0,0 +1,23 @@
+within Buildings.Fluid.Movers.Data.Fans.EnglanderNorford1992;
+record Return "Data for the return fan used in Englander and Norford (1992)"
+  extends Generic(
+    etaHydMet=Buildings.Fluid.Movers.BaseClasses.Types.EfficiencyMethod.PowerCurve,
+    speed_rpm_nominal=800,
+    power(V_flow={10.2875112309074,24.9775381850853,35.4447439353099,53.414195867026},
+          P={29828,44742,51050.622,44742}),
+    pressure(V_flow={10.2875112309074,24.9775381850853,35.4447439353099,53.414195867026},
+             dp={1453.61155698234,1371.42857142857,1052.9695024077,0}));
```
mwetter, line 9: Please format the data so that they are easier to read.

mwetter (commented):
@hcasperfu : This looks almost complete and is carefully implemented. Can you please enable the Euler method for the models that are based on `Buildings.Examples.VAVReheat.BaseClasses.PartialHVAC`. This way we also have large tests that use the new model.

I added minor comments inline.

Buildings/Fluid/Movers/Data/Fans/EnglanderNorford1992/Return.mo
hcasperfu: @mwetter
Do you want me to apply some format to just the two Englander data files or all of the data records?

Buildings/Fluid/Movers/UsersGuide.mo
```diff
@@ -437,62 +437,271 @@ If the motor is cooled by the fluid, as indicated by
...
+<code>Values</code>, or <code>NotProvided</code>.<br/>
+** Although the Euler number method is defined using <i>&eta;<sub>hyd</sub></i>,
+because the motor efficiency tends to be largely constant from full load down to
+25% - 50% of part motor load (see the documentation of
+<a href=\"Modelica://Buildings.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve\">
+Buildings.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve</a>
+), the relationship between <i>&eta;</i> and
+<i>&eta;<sub>hyd</sub></i> is roughly linear for the range that matters most.
+It is therefore reasonable to approximate that the <code>EulerNumber</code> method
+can also be used to define <i>&eta;</i>.
```
mwetter, line 704: Please reformulate (see changes in paper manuscript.)

Buildings/Fluid/Movers/Data/Fans/EnglanderNorford1992/package.mo
hcasperfu: Moved as suggested in ab1946af1893e81835011269038350cd54bead06.

Buildings/Fluid/Movers/Data/Fans/EnglanderNorford1992/Return.mo
hcasperfu: Formatted in f40adfd71d5db51a65db546f38607bd4c0eef3ba and 4662c8720e861e6051a4bca605383a4e491c3e33.

Buildings/Fluid/Movers/BaseClasses/Characteristics/efficiencyParameters_yMot.mo
```diff
@@ -0,0 +1,27 @@
+within Buildings.Fluid.Movers.BaseClasses.Characteristics;
+record efficiencyParameters_yMot
+  "Record for efficiency parameters vs. motor part load ratio"
+  extends Modelica.Icons.Record;
+  parameter Real y[:](each min=0)
+    "Part load ratio, y = PEle/PEle_nominal";
+  parameter Modelica.Units.SI.Efficiency eta[size(y, 1)](each max=1)
+    "Fan or pump efficiency at these part load ratios";
```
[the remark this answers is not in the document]
hcasperfu: Revised the user guide in 6929eb416d5485e6b5cfb2c8001f8c7fbf17b6fb.

Buildings/Fluid/Movers/UsersGuide.mo
hcasperfu: Revised in 6929eb416d5485e6b5cfb2c8001f8c7fbf17b6fb.

hcasperfu:
> 

@mwetter 
I created a duplicate model of `Buildings.Examples.VAVReheat.ASHRAE2006` that uses the Euler method in 
f301fc96ea82e15c8e572fbb6d26eec9d8909abd.
The results did not change much. The fan `hvac.fanSup` was mostly just on-and-off, and the new estimated efficiency is in the range of 0.484 - 0.486, very close to the old constant value of 0.49.
I tried to make a similar model based on `Buildings.Examples.VAVReheat.Guideline36` as well, but couldn't get it to work. It appears state events would be generated endlessly near `T = 100917` when the fan was turned on for the second time. This is probably worth investigating.

Buildings/Fluid/Movers/UsersGuide.mo
```diff
@@ -437,62 +446,276 @@ If the motor is cooled by the fluid, as indicated by
...
- <i>&eta;<sub>mot</sub></i> need to be provided by the user, and hence
-the model computes
+<i>&eta;<sub>mot</sub></i> can be computed using the following choices:
 </p>
-<p align=\"center\" style=\"font-style:italic;\">
-  &eta; = &eta;<sub>hyd</sub> &nbsp; &eta;<sub>mot</sub><br/>
-  P<sub>ele</sub> = W<sub>flo</sub> &frasl; &eta;.
+<ul>
+<li>
+<code>Efficiency_VolumeFlowRate</code>
```
mwetter, line 620: This item misses the description, please add it.

Buildings/Fluid/Movers/UsersGuide.mo
```diff
@@ -437,62 +446,276 @@ If the motor is cooled by the fluid, as indicated by
...
+oversize would be used as default.
+The model then computes the efficiency the same way as the option of
+<code>Efficiency_MotorPartLoadRatio</code>.
+Also see
+<a href=\"Modelica://Buildings.Fluid.Movers.BaseClasses.Validation.FlowMachineInterface_yMot\">
+Buildings.Fluid.Movers.BaseClasses.Validation.FlowMachineInterface_yMot</a>
+as an example.
+</li>
+<li>
+<code>NotProvided</code>
```
mwetter, line 649: This item misses the description, please add it

mwetter (changes requested):

Buildings/Fluid/Movers/UsersGuide.mo
hcasperfu: Added in df96e8233e3b1cc7bb2fd7285a1813900805191f.

Buildings/Fluid/Movers/UsersGuide.mo
hcasperfu: Added in df96e8233e3b1cc7bb2fd7285a1813900805191f.

mwetter:
@hcasperfu : Can you please update this branch (and/or the one from https://github.com/ibpsa/modelica-ibpsa/pull/1646) so that the merge does not cause many diffs. Then we have one set of consistent models which is easier to review.
I run
```python
import buildingspy.development.merger as m
import os
import shutil
from git import Repo
from os.path import expanduser
home = expanduser("~")
ibpsa_dir = os.path.join(home, "github/ibpsa/modelica-ibpsa/IBPSA")
dest_dir = os.path.join(os.getcwd(), "Buildings")
mer = m.IBPSA(ibpsa_dir, dest_dir)
mer.merge()

os.remove(os.path.join(dest_dir, "Resources", "Images", "IBPSA-logo-text.png"))
os.remove(os.path.join(dest_dir, "Resources", "Images", "IBPSA-logo.png"))
```
and get various changes, including models not related to movers such as in `Air.mo` and `PerfectGases.mo`

hcasperfu:
@mwetter 

I ran the scripts and looked at the differences.
- Most of them are the difference in the change log, where I marked "<span>#</span>2668" as "Buildings, <span>#</span>2668" for documentation in the IBPSA library. I have reverted them in IBPSA.
- Some of them are minor tweaks I made when merging from Buildings to IBPSA. I have merged them to Buildings.
- `smoothInterpolation.mo` had some format changes when I copied it to IBPSA. I have reverted them in IBPSA.
- As for `splineDerivatives.mo`, I have no clue why it got changed. But Buildings has the newer version and I have merged it to IBPSA.
- The remaining files below are different in their respective master branches. I think this is because they are yet to be synced between the two libraries.
  - `HeaterCooler_u.mo`
  - `Air.mo`
  - `PerfectGas.mo`

hcasperfu:
This is merged via #3248 and #3252.

PR #2738, opened 2021-11-09
Scale eplus floor

### LINKED ISSUE #2737: Test scalability of Spawn floors compared to MixedAir floors
Test translation and run time for large office floor using either EnergyPlus envelope simulation or MBL MixedAir components.

bravache:
This closes #2737 
The changes were brought from https://github.com/lbl-srg/modelica-buildings/tree/scale_eplus_floor


bravache:
This PR adds the scalable example models from the Modelica conference paper. I didn't make any `mos` for those example since I think they would just slow down Travis testing without adding much coverage to existing models.

---

## Summary

This adds a new package in ` Buildings/Examples/ScalableBenchmarks` called `ZoneScaling` which has validation model that are used to analyze the computational impact of scaling the number of zones, using either `Buildings.ThermalZones.Detailed.MixedAir` or  `Buildings.ThermalZones.EnergyPlus.ThermalZone`.

The dependency for the spawn thermal zones are a set of large office idf files, which have been created using helper functions in `makeEplusBuilding.py`. Part of the process of creating these scalable idf files was to also manually go through the files, so this is not fully automatable, but the helper function might be useful for future project. I don't know if they belong in the MBL resources though.

---

Note: I did a wide search/replace for the term 'corridor' to become 'core', since both terms are used for referring to the same thermal zones. "Core" is actually the terminology that is used by the commercial reference building model, and is more accurate given that the layout and the load of that zone.

bravache:
To fix: There seem to be issue with running the Eplus cases, which is recent. Need to be investigated before this is merged.

mwetter:
@bravache : I suggest we try to add CI tests for all cases but with only 1 day of simulation.

If this takes too long for the CI tests, we could keep only the cases with 2 floors as `.mo` files, and assign `numFlo` to be 2, 4 or 10 with an enumeration. Then, we can add the 2 floor cases to the CI tests to ensure that they always work. This way, we don't have `.mo` files that are not tested. 
We should however keep the `.idf` files for the 2, 4 and 10 cases so they can easily be run by changing the value of the enumeration.

Buildings/Examples/ScalableBenchmarks/ZoneScaling/BaseClasses/HVACMultiFloorBuilding.mo
```diff
@@ -0,0 +1,235 @@
...
+    final mNor_flow_nominal=mNor_flow_nominal,
+    final mWes_flow_nominal=mWes_flow_nominal,
+    final mCor_flow_nominal=mCor_flow_nominal,
+    final m_flow_nominal={0.7*sum(mVAV_flow_nominal[i]) for i in 1:numFlo},
+    each final THotWatInl_nominal=THotWatInl_nominal) "HVAC system"
+    annotation (Placement(transformation(extent={{-48,-28},{40,22}})));
+
+  replaceable Buildings.Examples.VAVReheat.BaseClasses.PartialFloor flo[numFlo] constrainedby
+    Buildings.Examples.VAVReheat.BaseClasses.PartialFloor(redeclare final
+      package Medium =                                                                     MediumA)
```
mwetter, line 89: Correct format

Buildings/Examples/ScalableBenchmarks/ZoneScaling/EnergyPlus/Large10Floors.mo
```diff
@@ -0,0 +1,27 @@
+within Buildings.Examples.ScalableBenchmarks.ZoneScaling.EnergyPlus;
+model Large10Floors
+  "Model of a large building with 10 floors and 50 conditioned zones"
+  extends
+    Buildings.Examples.ScalableBenchmarks.ZoneScaling.EnergyPlus.Large2Floors(      final numFlo=10);
```
mwetter, line 5: Fix format.

Buildings/Examples/ScalableBenchmarks/ZoneScaling/EnergyPlus/Large2Floors.mo
```diff
@@ -0,0 +1,49 @@
+within Buildings.Examples.ScalableBenchmarks.ZoneScaling.EnergyPlus;
```
mwetter, line 1: This model is not valid:
```
checkModel("Buildings.Examples.ScalableBenchmarks.ZoneScaling.EnergyPlus.Large2Floors");
 = false
```

Buildings/Examples/ScalableBenchmarks/ZoneScaling/EnergyPlus/Large2Floors.mo
```diff
@@ -0,0 +1,49 @@
...
+    weaName=weaName,
+    epwName=epwName,
+    computeWetBulbTemperature=false)
+    "Building-level declarations"
+    annotation (Placement(transformation(extent={{-20,60},{0,80}})));
+
+    annotation (
+experiment(
+      StopTime=432000,
+      Tolerance=1e-05),
```
mwetter, line 32: There needs to be a call to the .mos script.

Buildings/Examples/ScalableBenchmarks/ZoneScaling/MixedAir/Large10Floors.mo
```diff
@@ -0,0 +1,37 @@
+within Buildings.Examples.ScalableBenchmarks.ZoneScaling.MixedAir;
+model Large10Floors
+  "Model of a large building with 10 floors and 50 conditioned zones"
+  extends
+    Buildings.Examples.ScalableBenchmarks.ZoneScaling.MixedAir.Large2Floors(      numFlo=10)
```
mwetter, line 5: Fix format

Buildings/Examples/ScalableBenchmarks/ZoneScaling/MixedAir/Large2Floors.mo
```diff
@@ -0,0 +1,32 @@
...
+    Buildings.Examples.ScalableBenchmarks.ZoneScaling.BaseClasses.HVACMultiFloorBuilding(
+    numFlo=2,
+    redeclare final Buildings.Examples.ScalableBenchmarks.ZoneScaling.MixedAir.BaseClasses.LargeOfficeFloor flo(
+      each sampleModel=true),
+    redeclare final Buildings.Examples.VAVReheat.BaseClasses.Guideline36 hvac);
+
+    annotation (
+experiment(
+      StopTime=432000,
+      Tolerance=1e-05),
```
mwetter, line 15: call to .mos file is missing

Buildings/Examples/ScalableBenchmarks/ZoneScaling/MixedAir/Large4Floors.mo
```diff
@@ -0,0 +1,27 @@
+within Buildings.Examples.ScalableBenchmarks.ZoneScaling.MixedAir;
+model Large4Floors
+  "Model of a large building with 4 floors and 15 conditioned zones"
+  extends
+    Buildings.Examples.ScalableBenchmarks.ZoneScaling.MixedAir.Large2Floors(      final numFlo=4);
```
mwetter, line 5: fix format

Buildings/Examples/ScalableBenchmarks/ZoneScaling/MixedAir/package.mo
```diff
@@ -0,0 +1,19 @@
+within Buildings.Examples.ScalableBenchmarks.ZoneScaling;
+package MixedAir "Package with scalable large office buildings modeled with MBL MixedAir"
```
mwetter, line 2: Use full words, users may not know what MBL MixedAir is, and if MixedAir gets renamed, this string will be invalid. If you want to link to a specific model, use `<a href=\"modelica://....` in the info section.

Buildings/Resources/Python-Sources/makeEplusBuilding.py
```diff
@@ -0,0 +1,212 @@
+import re
```
mwetter, line 1: This file should not be in this directory as the directory is used for files that are invoked from Modelica.
A better place would be `Buildings/Resources/src/Examples/ScalableBenchmarks/ZoneScaling/EnergyPlus`, and put all the idf template files that are required to generate the actual idf file. 

Then, also put some README.md file that explains what files such as `Buildings/Resources/Data/Examples/ZoneScaling/raw/RefBldgLargeOfficeNew2004_Chicago_template_adiabatic.idf` are.
It also should not be in `raw` I think, but maybe in `Resources/src` if it is used to generated the actual idf file.

Lastly, add to the travis CI tests a test that invokes `makeEplusBuilding.py`, and returns an error if the generated file is not the same as the one that is distributed with the library. Otherwise, they get out of sync and we won't notice for a while.

What is the process for updating these files as the EnergyPlus version changes? I assume the '*_template_*' files can be run through the EnergyPlus conversion program, and then `makeEplusBuilding.py` be invoked, but this is not clear.

mwetter (changes requested):
Please see inline comments.

Also, I don't see the San Francisco idf files anywhere used. These can be deleted from the Buildings library, and the cases with 2 zones need to be run as part of the unit tests. They are currently not run, and neither of them passes the model check.

Buildings/Resources/Python-Sources/makeEplusBuilding.py
bravache: I moved this file and the original and template models to `Buildings/Resources/src/Examples/ScalableBenchmarks/ZoneScaling/`. I have also removed the San Francisco models, which included the adiabatic version. These were created and used when we were iterating on the final product but we don't need to keep them anymore.

I've added a `README.md` that provides info about the utility and the files, and also give instruction about updating the Eplus version.

bravache (commented):
@mwetter I've addressed most comment, but I am having trouble with making `.mos` test for the scalable models. They run in Dymola, although it takes 2-3 minutes for the smallest (2 floors) cases, but if I run them through BuildingsPy, the EnergyPlus case takes between 5 and 10 minutes and the MixedAir case never completes while also creating a very large `.mat` file. A couple of time, the simulation stops because my disk partition was full from a 40Gb output.

Let's talk about it at the next coordination meeting, and also on how to create new CI test for the python utilities.

mwetter:
@bravache : What is the status of these remaining fixes? I have an application that would benefit from having this model on the master.

mwetter:
I will close this without merging as scaling the idf file is cumbersome to port to future versions.

Issue #2781, opened 2021-11-24
Edit CO2 generation calculation


dhblum:
This issue is just to update the calculation of CO2 generation in ``Buildings.Examples.VAVReheat.Validation.BaseClasses.Floor`` so as to directly use the output from internal gain schedule and assumed 0.05 ppl/m^2 population density, rather than a latent load per person assumption.  This should not change unit test results significantly, if at all, for ``Buildings.Examples.VAVReheat.Validation.TraceSubstances``.

mwetter:
This is now on the master.

PR #2782, opened 2021-11-24
issue2781 co2 gen update

dhblum:
This is for #2781.  It should not change unit test results significantly if at all.


dhblum:
@mwetter This is a small change.

Buildings/Examples/VAVReheat/Validation/BaseClasses/Floor.mo
```diff
@@ -61,8 +66,8 @@ Buildings.Examples.VAVReheat.Validation.TraceSubstance</a>.  It extends
 <a href=\"modelica://Buildings.Examples.VAVReheat.BaseClasses.Floor\">
 Buildings.Examples.VAVReheat.BaseClasses.Floor</a> with CO<sub>2</sub> generation
 from people and CO<sub>2</sub> from outside air infiltration.  Calculation of the
-generation from people uses the gain schedule for latent load and assumes
-80 W per person latent load to calculate the number of people.  Then, it assumes
+generation from people uses the internal gain schedule and uses 0.05 ppl/m^2
```
mwetter, line 69: @dhblum : do you mean PPM/m<sup>2</sup> during occupied hours rather than ppl?

mwetter (changes requested):
See inline comment

Buildings/Examples/VAVReheat/Validation/BaseClasses/Floor.mo
dhblum: No, 0.05 is the assumed design occupancy density in the space so it should be ppl (or people).  PPM per person is discussed in the next sentence.

mwetter: Ah, now I understand what ppl means. I made a commit that spells out the abbreviation.

dhblum: Sounds good thanks 👍

PR #2784, opened 2021-11-25
Include ModelicaUtilities.h for crypto hash

sjoelund:
The default prototype for ModelicaAllocateString will be int32, which
does not fit a whole pointer causing crashes in models using the
function.

Note: This should probably be ported to maintenance branches as well. It was reported in https://github.com/OpenModelica/OpenModelica/issues/8219

### DIFF
--- a/Buildings/Resources/C-Sources/cryptographicsHash.c
+++ b/Buildings/Resources/C-Sources/cryptographicsHash.c
@@ -24,6 +24,7 @@ A million repetitions of "a"
 /* for uint32_t */
 #include <stdint.h>
 #include "cryptographicsHash.h"
+#include "ModelicaUtilities.h"
 
 
 #define rol(value, bits) (((value) << (bits)) | ((value) >> (32 - (bits))))


mwetter:
@sjoelund : Thanks, I applied it to the master, maint_7.0.x and maint_8.0.x branch, and will apply it to IBPSA through https://github.com/ibpsa/modelica-ibpsa/pull/1558

casella:
This should improve the Buildings 7.0.x success ratio quite significantly. Let's see the next regression report, Martin has already started [a run of the master test job](https://test.openmodelica.org/jenkins/job/LibraryTesting/job/LibraryTest/4151/).

casella:
@Mathadon pointed out in [his comment](https://github.com/ibpsa/modelica-ibpsa/pull/1558#issuecomment-979375209) that we first removed `ModelicaUtilities.h`, and now we're adding it again, which is a bit contradictory.

I guess there is some misunderstanding between adding or removing _the file_ and adding or removing _the include directive_.

The issue is discussed in MSL ticket [#3687](https://github.com/modelica/ModelicaStandardLibrary/issues/3867). The conclusion of the discussion is that the content of `ModelicaUtilities.h` is implicitly defined by the Modelica language specification in [Sect. 12.9.6](https://specification.modelica.org/maint/3.5/functions.html#utility-functions), which introduces several C functions that are expected to be available to external functions in all Modelica installations. However, different Modelica tools could implement it in slightly different ways (e.g. because of 32- vs. 64-bits or something), so it doesn't make sense to have a one-fits-all `ModelicaUtilities.h` in the MSL, nor it makes sense to have a one-fits-all `ModelicaUtilities.h` in the library Resource files.

If I am not mistaken, the conclusion of the discussion were that:
- the `ModelicaUtilities.h` file as such should be removed from the ModelicaStandardLibrary repository (done in MSL 4.1.0)
- each tool vendor should create their own one, based on their actual implementation (already done in OMC and Dymola)
- each tool vendor should make avaliable a proper `ModelicaUtilities.h` file in their environment, so that external functions can access it when doing `#include "ModelicaUtilities.h"` (already done in OMC and Dymola)

So, 
- you do need to add the `#include "ModelicaUtilities.h"` `directive` to the C function, as @sjoelund did in 691706c, otherwise you may get bad behaviour, i.e. a segfault. The C language can be quite nasty...
- you shouldn't add the `ModelicaUtilities.h` _file_ to the Resource directory, but rather rely on the one that is supplied by the Modelica tool you're using
- if the C source code is directly provided in the Include annotations (hence compiled on the fly every time you compile the Modelica model), the Modelica tool is configured so that `#include "ModelicaUtilities.h"` actually includes the right .h file from the Modelica tool's installation directory
- note that `#include <ModelicaUtilities.h>` won't work, you need to use `#include "ModelicaUtilities.h"`

@Mathadon does this address your concerns?

sjoelund:
And if there is a bug, fix it in the tool. Not including ModelicaUtilities.h will not work. I'm wondering if the other issue was a problem for FMUs in OM 1.17... Not going to check it now

casella:
Eight more models running successfully in all tested versions of Buildings, see [report](https://libraries.openmodelica.org/branches/history/master/2021-11-25%2011:41:34..2021-11-25%2015:40:12.html).

Buildings 7.0.x [now at 1125/1180](https://libraries.openmodelica.org/branches/master/Buildings_maint.7.0.x/Buildings_maint.7.0.x.html) = 95.3% success rate.

Issue #2820, opened 2021-12-16
merge ice tank model


YangyangFu:
@mwetter can you please create a new branch so that I can make a pull request?

Or you can get the model here: https://github.com/BE-HVACR/modelica-ice-tank/tree/release_for_mbl.

The current implementation is validated against onsite measurement data, which contain three discharging experiments (with various discharging rate) and one charging experiment (with fixed charging rate). The data and descriptions are also documented in the implementation. More details can be found in the following reference.

```
Li, G., Fu, Y., Pertzborn, A., Wen, J., & O’neill, Z. (2021, September). 
An Ice Storage Tank Modelica Model: Implementation and Validation. 
In Modelica Conferences (pp. 177-185).

mwetter:
@YangyangFu : Thanks, I created the branch `issue2820_iceTank`. Can you please make a PR against that branch. Note that this branch is based on MSL 3.2.3 which is the same version as your ice tank model. I think it will be easier for you to integrate it into that version. I therefore suggest we integrate it into that branch, and then I can upgrade that branch to MSL 4.0.0 which is what the master is currently at. See also https://github.com/lbl-srg/modelica-buildings/issues/2812

mwetter:
The current tank model contains built-in control for the outlet temperature during discharging. This makes it hard to use model for configurations such as this (unless cascading control where used, which is not needed and adds complexity):
![image](https://user-images.githubusercontent.com/1329080/149524496-6f55b782-b315-4aeb-a106-517fb57becb2.png)
(Image from https://www.evapco.eu/sites/evapco.eu/files/2017-03/Thermal%20Ice%20Storage%20Application%20%26%20Design%20Guide.pdf)
or the one in http://www.calmac.com/stuff/contentmgr/files/0/c0e62901daa3ddb1722d04aaeaf31997/files/ice_storage_how_and_why_lessons_learned_17_garcia_coulard4.pdf

Therefore, the model should be edited to move the control out of the component.

PR #2827, opened 2021-12-20
Issue2820 ice tank

YangyangFu:
@mwetter this is for issue #2820.

Major activities are:
1. implementation of an ice tank model;
2. validation of the ice tank model


mwetter:
@YangyangFu : Thanks for the pull request. I will merge it and further integrate as needed. I also assume you are fine if I remove the license file (https://github.com/lbl-srg/modelica-buildings/pull/2827/files#diff-c693279643b8cd5d248172d9c22cb7cf4ed163a3c98c8a3f69c2717edd3eacb7) which seems consistent with LBL releasing it under the BSD license that the Modelica Buildings Library already uses.

PR #2833, opened 2021-12-22
Modify comments of supply temperature setpoint limits

### LINKED ISSUE #2828: make parameters in Multizone.VAV.Controller clearer
In `G36_PR1.AHUs.Multizone.VAV.Controller`, the parameter `TSupSetDes` is not clear if this is for heating or cooling. It probably is for cooling as it is used in
```modelica
parameter Real minSetSupTem(
    final unit="K",
    final displayUnit="degC",
    final quantity="ThermodynamicTemperature")=TSupSetDes
    "Minimum setpoint for supply temperature control"
    annotation (Dialog(tab="Supply air temperature", group="Trim and respond for reseting TSup setpoint"));
``` 
However, this raises the question why a user is given the option to use a different value for `minSetSupTem`. The model should explain why or when different settings are needed, or else don't allow different settings.
The same applies to `TSupSetMax`, which is assigned to `maxSetSupTem` in the top-level controller.

This should be made clear, in particular for the release of the official version of Guideline 36.

JayHuLBL:
This closes #2828.
- [x] improved the comments of design supply temperature, minimum cooling supply air temperature
- [x] made the trim and respond limit parameters to be protected


mwetter (approved):
@JayHuLBL : This is ready to merge, please let me know if you agree with my two commits, and also apply them to the implementation of the official release of Guideline 36.

JayHuLBL:
@mwetter I agree with your commits and will apply the changes to the official release.

PR #2868, opened 2022-01-28
Issue2866 disallow tower model reversal

### LINKED ISSUE #2866: Cooling tower sta_a.T and sta_b.T swap due to flow reversal in computation
First reported in comments of PR #2693.
When simulating `Buildings.Fluid.HeatExchangers.CoolingTowers.Examples.Merkel` and other models that use a cooling tower, the temperatures of the two ports of the cooling tower `sta_a.T` and `sta_b.T` may swap at times. This occurs when the flow rate of the tower is supposed to be zero, but its actual value is a very small negative number (see figures below for a snapshot of a few time steps).
image
image

This could be solved by adding the modifier `final allowFlowReversal=false` when `Buildings.Fluid.HeatExchangers.CoolingTowers.BaseClasses.CoolingTower` extends `Buildings.Fluid.Interfaces.TwoPortHeatMassExchanger`.

`Buildings.Fluid.HeatExchangers.CoolingTowers.Examples.Merkel`
Before:
Screen Shot 2022-01-27 at 16 46 52

After:
Screen Shot 2022-01-27 at 16 44 50

This will change the results of quite a few models.

hcasperfu:
This closes #2866.


hcasperfu:
Will solve the issue with a different approaching.

PR #2878, opened 2022-02-04
Merge IBPSA, issue1580_OpenModelicaCI

mwetter:
This also merges the update to the calculation of sta_a and sta_b


mwetter:
@hcasperfu : FYI, this will merge your changes to the calculation of `sta_a` and `sta_b`.

Issue #2848, opened 2022-01-10
Adding a steam pressure regulating valve model


SaranyaAnbarasu:
**Pressure regulating valve (PRV) for Steam District Heating with the URBANopt Project** 

@bravache: Hi Baptiste, could you please open a new branch for the steam PRV models as described below? If you have any questions or comments, please let us know. Thank you!

(@mwetter: Dr. Wetter, we're mentioning you here as well in case you would like to provide any particular input. Thanks!)
(@khinkelman, mentioning you for your reference).

The existing `Buildings.Fluid.Actuators.Valves.TwoWayLinear` are specific to incompressible media and do not include density to compute the pressure drop across valves.  Whereas steam is a compressible medium, so the pressure drop varies with changes in density (density of the steam medium is considered variable over the range of pressure conditions seen in real systems).  The existing valve models do not capture this thermodynamic property. 

This valve model will be essential for district heating applications on the demand side (reducing the high-pressure steam from boilers before passing through the heat exchangers). 

**Our proposed solution** 

To capture the correct performance, we are proposing new steam valves:
(1) SteamTwoWayValve – Two-way valve with opening `y` input, similar to `Buildings.Fluid.Actuators.Valves.TwoWayLinear`
 
(2) SteamTwoWayValveSelfActing -  SteamTwoWayValve with a PI controller to maintain the outlet pressure to the required setpoint. From testing, this seems to be the most numerically efficient way to mimic the self-actuating (damped springs) in the physical valves.
 
The implementation is similar to `Buildings.Fluid.Actuators.Valves.TwoWayLinear`, extended from the base class `SteamValveParameters` and `PartialSteamTwoWayValveKv`. The valve parameters are implemented based on ISA 75.01.01/IEC 60534 steam (compressible medium).  The instantaneous inlet density is used in the calculation of pressure drop across the valve. 

We had earlier proposed the steam valve models in the DHC package (https://docs.google.com/document/d/12GaIloYaMds4USyhZt1W_s2hB2W2_GWR/edit), but `Buildings.Fluid.Actuators.Valve` package seems to be a better fit. Please let us know your thoughts on the package structure below:
```
Buildings.Fluid.Actuators
•	Valves
    •	SteamTwoWayValve
    •	SteamTwoWayValveSelfActuating
    •	Examples
        •	SteamTwoWayValve
        •	SteamTWoWayValveSelfActing
    •	Validation
        •	SteamTwoWayValveParameterization
[...]

PR #2908, opened 2022-03-03
Issue2848 steam valves

SaranyaAnbarasu:
Hi @bravache 
Here is PR for the issue #2848 on steam valve models for your review. 

We have simplified the self-acting valve model by using `Buildings.Fluid.Movers.BaseClasses.IdealSource,` that allows the pressure drop to be prescribed independently of mass flow rate. This eliminates the complex dependencies between `m_flow `and `dp `for compressible mediums (such as steam).

If you have any questions or require any revisions, I and @khinkelman will be available to help. 
Thank you


SaranyaAnbarasu:
Hi @bravache, 
This PR is now ready for your review. We have revised according to Dr. Wetter's suggestions (from GED meeting).  
Please let me know if you have any questions. 
Thanks!

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mo
```diff
@@ -0,0 +1,85 @@
...
+    annotation (Line(points={{40,0},{30,0}}, color={0,127,255}));
+  annotation (Icon(coordinateSystem(preserveAspectRatio=false)),
+    Diagram(coordinateSystem(preserveAspectRatio=false)),
+    __Dymola_Commands(file="modelica://Buildings/Resources/Scripts/Dymola/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mos"
+    "Simulate and plot"),
+    experiment(StopTime=15, Tolerance=1e-06),
+    Documentation(info="<html>
+<p>
+This example model demonstrates the performance of self-acting steam pressure regulator 
+with a noisy/varying inlet pressure signal and a ramped mass flow rate signal. The inlet
```
bravache, line 73: There is no ramp in the mass flow rate. You could actually add a varying pressure signal to the outlet boundary so that we would observe that the mass flow rate is dictated by the `dp` on the outlet.

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mo
```diff
@@ -0,0 +1,85 @@
...
+    Diagram(coordinateSystem(preserveAspectRatio=false)),
+    __Dymola_Commands(file="modelica://Buildings/Resources/Scripts/Dymola/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mos"
+    "Simulate and plot"),
+    experiment(StopTime=15, Tolerance=1e-06),
+    Documentation(info="<html>
+<p>
+This example model demonstrates the performance of self-acting steam pressure regulator 
+with a noisy/varying inlet pressure signal and a ramped mass flow rate signal. The inlet 
+pressure conditions are selected to demonstrate how the downstream pressure is maintained 
+at the setpoint, unless the inlet pressure drops below the setpoint (pressure drop is then zero).
```
bravache, line 75: The current inlet signal doesn't illustrate that behavior (the inlet pressure never drops below the setpoint). It could be interesting to change the signal so that this behavior is shown.

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mo
```diff
@@ -0,0 +1,85 @@
+within Buildings.Experimental.DHC.Loads.Steam.BaseClasses.Examples;
+model ValveSelfActing
+  "Example model for self-acting steam valve with a noisy inlet pressure input"
+  extends Modelica.Icons.Example;
+  package MediumSte = Buildings.Media.Steam "Medium model";
+  parameter Modelica.Units.SI.MassFlowRate m_flow_nominal=1
+    "Nominal mass flow rate";
+
+  Buildings.Experimental.DHC.Loads.Steam.BaseClasses.ValveSelfActing prv(
+    redeclare package Medium = MediumSte,
```
bravache, line 10:
```suggestion
    redeclare final package Medium = MediumSte,
```

(and all the other assignment later)

bravache (approved):
Great job @SaranyaAnbarasu , it looks clean and great! I have a couple of comments, mostly around parameters value, but it should be fairly quick to resolve. I'll merge to the `lbl-srg` repo once you have a look at that.

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/ValveSelfActing.mo
```diff
@@ -0,0 +1,136 @@
...
+    final m_flow_start=m_flow_nominal,
+    final m_flow_small=m_flow_small,
+    final show_T=show_T,
+    final control_m_flow=false,
+    final control_dp=true) "Ideal source controlling dp"
+    annotation (Placement(transformation(extent={{40,-10},{60,10}})));
+  Modelica.Blocks.Sources.RealExpression pbSet(final y=pb_nominal)
+    "Downstream pressure setpoint"
+    annotation (Placement(transformation(extent={{-80,46},{-60,66}})));
+  Modelica.Blocks.Math.Add dpReq(final k1=-1)
```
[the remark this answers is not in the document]
SaranyaAnbarasu: Replaced a Subtract block instead as add doesn't allow specifying `k1=-1`

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/ValveSelfActing.mo
```diff
@@ -0,0 +1,136 @@
...
+  extends Buildings.Fluid.Interfaces.PartialTwoPortInterface;
+  parameter Modelica.Units.SI.Pressure pb_nominal(
+    displayUnit="Pa",
+    min=101325)
+    "Nominal outlet pressure"
+    annotation (Dialog(group="Nominal condition"));
+
+  Buildings.Fluid.Movers.BaseClasses.IdealSource ideSou(
+    redeclare final package Medium = Medium,
+    final allowFlowReversal=allowFlowReversal,
```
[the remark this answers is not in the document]
SaranyaAnbarasu: While setting `allowFlowReversal = false` within the extended `PartialTwoPortInterface` is not enforcing zero flow reversal, whereas there is an error-type assert warning in `IdealSource` that actually enforces zero flow reversal (and stops the simulation when `m_flow<0`). 

Hence, within `IdealSource`, should we set `final allowFlowReversal = allowFlowReversal`, which will trigger an error and stops the simulation, or should we set` final allowFlowReveral = true` only within this block so that the error is not triggered? 

Please let me know your thoughts.

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/ValveSelfActing.mo
```diff
@@ -0,0 +1,136 @@
...
+    final dp_start=pb_nominal,
```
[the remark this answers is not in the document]
SaranyaAnbarasu: Exposed the parameter `dp_start` and set a default valve

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mo
SaranyaAnbarasu: Since this is an example model, I believe `final` may not be required.  
Please let me know what you think.

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mo
SaranyaAnbarasu: Agreed. 
I have revised the input signal to drop below the setpoint. Now we could see the change in `m_flow` based on the `dp` as well as the triggered warnings.

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mo
SaranyaAnbarasu: Sorry, this is a documentation mistake that I missed revising. Thanks for pointing it.
We have not used a ramped mass flow rate and by revising the inlet signal values (to fall below setpoint) we are able to observe the variation in mass flow rate.

SaranyaAnbarasu:
@bravache 
Thank you for reviewing the steam valves. 
We agree with all the suggested changes and have revised the branch accordingly.  Yet, the suggestion on `allowFlowReversal` is triggering an error and stopping the simulation.  Could you please have look at the comment added to the unresolved conversations? 

Many thanks!

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mo
bravache: You are right that it is not required, whether in a model or in an example, and if you do a broad search on the Buildings library, you'll see about an equal amount of `redeclare package Medium = Medium` and `redeclare final package Medium = Medium`.
But I personally prefer to make those redeclarations final since there should not be any conceivable cases where a user would extend or use these models and cutomize the underlying component with a different Medium. Even though it is an example, a user might want to extend it to test other parameters value, in which case there shouldn't be an option to change the propagation of the medium from the top-level class down to the components.

Does that make sense?

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/ValveSelfActing.mo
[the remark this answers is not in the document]
bravache: Well, I am a bit torn about this. So, in the current situation, if you were to propagate `allowFlowReversal = false` down to the ideal source component, the example fails when the flow reverse. If you set it to `true`, then when the source pressure drops below the sink pressure, the flow reverse. The valve is now a passthrough, with no pressure drop (port_b.p == port_a.p).

My issue with this is that it seems that in a real self-acting valve, this is not realistic. I believe (but please let me know if you think I'm wrong), that when the outlet pressure is greater than the inlet pressure, the valve closes, which result in no flow. If that's the case, then you should add a check valve (`Buildings.Fluid.FixedResistances.CheckValve`) somewhere on the path of the valve.

I have tried with the following parameters which seems to work:
```modelica
  Fluid.FixedResistances.CheckValve cheVal(
    redeclare final package Medium = Medium,
    final m_flow_nominal=m_flow_nominal,
    dpValve_nominal=1,
    rhoStd=Medium.density_pTX(
        pb_nominal,
        273.15 + 100,
        Medium.X_default),
    dpFixed_nominal=0)
```

In that case, the `allowFlowReversal` can be left to its default value (true).

The other proposition would be to actually set the `allowFlowReversal` to `false`, and actually let the model fail when the conditions are not handled by the model. In that case, you should just change the inlet parameters in the example so that it doesn't fail anymore.

What do you think?

bravache:
@SaranyaAnbarasu I've replied to the `allowFlowReversal` comment. If I can paraphrase (maybe my comment is not super clear): the current self-acting valve is not realistic if the flow is reversed (in that case, it behaves like a lossless pipe, as we can see in the example). We should not allow this behavior, even with a warning. We can either prevent it from happening by raising an error (by fixing `allowFlowReversal` to `false`), or we can change the model so that it is more congruent with the physical behavior (the valve closes when the outlet pressure rises above the inlet).

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/ValveSelfActing.mo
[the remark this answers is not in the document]
SaranyaAnbarasu: I think the second option suggested would be a better fit for a real self-acting steam pressure reducing valve (PRV).  

Steam systems will have a number of valves connected in series along the pipelines, and each valve has a specific function (e.g.: pressure reduction, back-flow prevention, mass flow control etc.). Moreover, having a check valve within the self-acting valve would still be a problem, as self-acting PRVs are typically added before the heat exchanger, and zero `m_flow` from PRV (due to check valve) will eventually fail the heat exchanger. Hence, setting the `allowFlowReversal` to `false` and letting the model fail would be the best option.  With this revision, I have removed the `assert ()` warning for negative `m_flow` as it is no longer needed. 

As suggested, I have changed inlet pressure signal in the example model to successfully run the simulations.

Buildings/Experimental/DHC/Loads/Steam/BaseClasses/Examples/ValveSelfActing.mo
SaranyaAnbarasu: Thanks for explaining!
I have added `final` to medium declarations in the example.

SaranyaAnbarasu:
Hi @bravache 

Thanks for the comments and I have revised the models accordingly. I prefer to set `allowFlowReversal` to `false` and let the model fail, which represents a real steam pressure reducing valve. Please review it and let me know if any other changes are required.

SaranyaAnbarasu:
In addition, I have encountered a minor issue in warning message triggered.  The assert condition `dpReq.y > 0` is implemented as below:
```
assert (dpReq.y > 0, "pb_nominal is set higher than the upstream pressure in "
  + getInstanceName() + ", which results in a negative pressure drop. 
  This is not typical of real systems and should be verified.", AssertionLevel.warning);
```
But the warning message that is output has `Failed condition: prv. dpSet.u2 > 0`. Technically `dpReq.y = dpSet.u2`, but I am not sure why Modelica doesnot output the warning based on the condition specified in the assert statement. 

Please let me know if you have suggestions to resolve this issue. 

image

image

bravache:
> 

That's a good question, and it has to do with how Dymola interpret the models and optimize its definition before simulation. Basically, to reduce the size of the system, Dymola removes any constant variables (a = k with k being constant in time) and any alias variable (a = b or a = -b where b is another variable).

More information can be found here:
https://www.claytex.com/tech-blog/model-translation-and-symbolic-manipulation/

If you want to solve that issue, you can just swap out the order of the declaration for the `dpSet` and `dpReq` blocks, so that `dpReq` is defined first, and when the alias reduction is done, the alias that will be kept will be `dpReq.y`. Does that make sense?

SaranyaAnbarasu:
@bravache 
Many thanks for the suggestion. 
I have swapped the declaration and the model triggers the defined warning. The branch is now ready for your review.

bravache:
@SaranyaAnbarasu Thanks for the quick revision! This branch is good to go. I'm merging it.

PR #2957, opened 2022-04-11
Issue2884 pid autotuning

mwetter:
This merges the PID autotuning to the master.


Buildings/Controls/OBC/CDL/Continuous/AMIGOWithFOTD.mo
```diff
@@ -0,0 +1,129 @@
...
+  Real y0 "Initial value of the process output";
+  Real u0 "Initial value of the process input";
+  Interfaces.RealInput tau "Normalized time delay"
+    annotation (Placement(transformation(extent={{-140,80},{-100,120}}), iconTransformation(extent={{-140,80},{-100,120}})));
+  Interfaces.RealInput dtOFF "Half-period length of a relay tuner"
+    annotation (Placement(transformation(extent={{-140,-120},{-100,-80}}),iconTransformation(extent={{-140,-120},{-100,-80}})));
+
+  Interfaces.RealInput dtON "Half-period length of a relay tuner"
+    annotation (Placement(transformation(extent={{-140,-72},{-100,-32}}),
+                                                                        iconTransformation(extent={{-140,-78},{-100,-38}})));
```
mwetter, line 18: Use camelCase (dtOff and dtOn), and correct formatting.

Buildings/Controls/OBC/CDL/Continuous/AMIGOWithFOTD.mo
```diff
@@ -0,0 +1,129 @@
...
+  Real Kp "Static gain";
+  Real T "Time constant";
+  Real L "Time delay";
+  Real k
+    "Gain of controller";
+  Real Ti
+    "Time constant of integrator block";
+  Real Td
+    "Time constant of derivative block";
```
mwetter, line 27: Add units.

Buildings/Controls/OBC/CDL/Continuous/NormalizedDelay.mo
```diff
@@ -0,0 +1,64 @@
...
+  Min min annotation (Placement(transformation(extent={{-60,-64},{-40,-44}})));
+  Divide rho "Half-period ratio" annotation (Placement(transformation(extent={{-16,-10},{4,10}})));
+  Modelica.Blocks.Sources.RealExpression zeroCheck(y=noEvent(if min.y > 0 then min.y else 1))
+    annotation (Placement(transformation(extent={{-58,-30},{-38,-10}})));
+equation
+  assert(abs(gamma-1)>0.01, "The relay signal should be asymmetric");
+//  assert(rho.y>1, "The half-period ratio should be larger than 1");
+//  assert(y>1 or y<0, "Invalid value for the normalized time delay");
+  y=(gamma-rho.y)/(gamma-1)/(0.35*rho.y+0.65);
+  connect(max.u1, dtON) annotation (Line(points={{-64,56},{-80,56},{-80,60},{-120,60}}, color={0,0,127}));
```
mwetter, line 20: Don't mix graphical modeling with textual, see https://simulationresearch.lbl.gov/modelica/userGuide/development.html#equations-and-algorithms

Buildings/Controls/OBC/CDL/Continuous/NormalizedDelay.mo
```diff
@@ -0,0 +1,64 @@
...
+  Interfaces.RealInput dtON "Half-period length for the higher value"
+    annotation (Placement(transformation(extent={{-140,40},{-100,80}}), iconTransformation(extent={{-140,40},{-100,80}})));
+  Interfaces.RealInput dtOFF "Half-period length of the lower value"
+    annotation (Placement(transformation(extent={{-140,-80},{-100,-40}}), iconTransformation(extent={{-140,-80},{-100,-40}})));
+  Interfaces.RealOutput y "Normalized time delay"
+    annotation (Placement(transformation(extent={{100,-20},{140,20}})));
+  Max max annotation (Placement(transformation(extent={{-62,40},{-42,60}})));
+  Min min annotation (Placement(transformation(extent={{-60,-64},{-40,-44}})));
+  Divide rho "Half-period ratio" annotation (Placement(transformation(extent={{-16,-10},{4,10}})));
+  Modelica.Blocks.Sources.RealExpression zeroCheck(y=noEvent(if min.y > 0 then min.y else 1))
```
mwetter, line 13: This relays on inequality of a `Real` variable which is not allowed per the Modelica Language Standard.

Buildings/Controls/OBC/CDL/Continuous/NormalizedDelay.mo
```diff
@@ -0,0 +1,64 @@
...
+          thickness=1),
+        Text(
+          extent={{226,60},{106,10}},
+          textColor={0,0,0},
+          textString=DynamicSelect("",String(y,
+            leftJustified=false,
+            significantDigits=3)))}),
+    Documentation(
+      info="<html>
+<p>Block that calculates the normalized time delay based on results of a relay experiment for PID tuning</p>
```
mwetter, line 57: Expand info section. It needs to be self-explanatory.

Buildings/Controls/OBC/CDL/Continuous/Validation/AMIGOWithFOTD.mo
```diff
@@ -0,0 +1,66 @@
+within Buildings.Controls.OBC.CDL.Continuous.Validation;
+model AMIGOWithFOTD "Validation model for the AMIGOWithFOTD block"
+
+  Buildings.Controls.OBC.CDL.Continuous.AMIGOWithFOTD FOTD(controllerType=Buildings.Controls.OBC.CDL.Types.SimpleController.PI)
+    annotation (Placement(transformation(extent={{20,-10},{40,10}})));
+  Modelica.Blocks.Sources.Sine sine(
+    amplitude=1.2,
+    f=1/60,
+    offset=0.3,
+    startTime=0.025)                                      "Sine source"
```
mwetter, line 10: Fix format

Buildings/Controls/OBC/CDL/Continuous/Validation/AMIGOWithFOTD.mo
```diff
@@ -0,0 +1,66 @@
...
+    annotation (Placement(transformation(extent={{-82,30},{-62,50}})));
+  Modelica.Blocks.Sources.Constant const(k=0) annotation (Placement(transformation(extent={{-82,-10},{-62,10}})));
+  Buildings.Controls.OBC.CDL.Continuous.NormalizedDelay NormalizedDelay(gamma=3) annotation (Placement(transformation(extent={{-40,-60},{-20,-40}})));
+  Discrete.Relay                            relay(
+    yHigher=4,
+    yLower=-0.5,
+    deadBand=0.5)
+                annotation (Placement(transformation(extent={{-40,20},{-20,40}})));
```
mwetter, line 18: add missing comment

Buildings/Controls/OBC/CDL/Continuous/PIDWithAutoTuning.mo
```diff
@@ -0,0 +1,515 @@
...
+  Modelica.Blocks.Sources.RealExpression gainI(y=errI2.y*k/Ti) if
+                                                               with_I
+                                                          annotation (Placement(transformation(extent={{-106,22},{-86,42}})));
+  Modelica.Blocks.Sources.RealExpression antWinGai(y=antWinErr.y/(k*Ni)) if
+                                                               with_I annotation (Placement(transformation(extent={{178,-34},{158,-14}})));
+  Buildings.Controls.OBC.CDL.Discrete.Relay
+        relay(
+    yHigher=yHigher,
+    yLower=yLower,
+    deadBand=deadBand)
```
mwetter, line 116: Add missing comment.

Buildings/Controls/OBC/CDL/Continuous/PIDWithAutoTuning.mo
```diff
@@ -0,0 +1,515 @@
...
+    annotation (Placement(transformation(extent={{120,80},{140,100}})));
+
+  Modelica.Blocks.Sources.RealExpression P(y=errP.y*k) annotation (Placement(transformation(extent={{-50,130},{-30,150}})));
+  Modelica.Blocks.Sources.RealExpression gainD(y=Td*k*errD.y) if
+                                                               with_D annotation (Placement(transformation(extent={{-106,60},{-86,80}})));
+  Modelica.Blocks.Sources.RealExpression gainI(y=errI2.y*k/Ti) if
+                                                               with_I
+                                                          annotation (Placement(transformation(extent={{-106,22},{-86,42}})));
+  Modelica.Blocks.Sources.RealExpression antWinGai(y=antWinErr.y/(k*Ni)) if
+                                                               with_I annotation (Placement(transformation(extent={{178,-34},{158,-14}})));
```
mwetter, line 111: add missing comment and fix format

Buildings/Controls/OBC/CDL/Continuous/PIDWithAutoTuning.mo
```diff
@@ -0,0 +1,515 @@
...
+    "Limiter"
+    annotation (Placement(transformation(extent={{120,80},{140,100}})));
+
+  Modelica.Blocks.Sources.RealExpression P(y=errP.y*k) annotation (Placement(transformation(extent={{-50,130},{-30,150}})));
+  Modelica.Blocks.Sources.RealExpression gainD(y=Td*k*errD.y) if
+                                                               with_D annotation (Placement(transformation(extent={{-106,60},{-86,80}})));
+  Modelica.Blocks.Sources.RealExpression gainI(y=errI2.y*k/Ti) if
+                                                               with_I
+                                                          annotation (Placement(transformation(extent={{-106,22},{-86,42}})));
+  Modelica.Blocks.Sources.RealExpression antWinGai(y=antWinErr.y/(k*Ni)) if
```
mwetter, line 110: antWinErr is conditionally removable and therefore the Modelica Language Standard does not allow using it in an equation. This causes the model check to fail.

Buildings/Controls/OBC/CDL/Continuous/Validation/PIDWithAutoTuning.mo
```diff
@@ -0,0 +1,62 @@
+within Buildings.Controls.OBC.CDL.Continuous.Validation;
+model PIDWithAutoTuning "Test model for PID controller"
+  Sources.Sine                                        sin1(freqHz=1/60, offset=0.3,
+    startTime=0.025)
+                 "Setpoint"
+    annotation (Placement(transformation(extent={{-90,14},{-70,34}})));
```
mwetter, line 6: fix format to make it easier to read.

Buildings/Resources/Scripts/Dymola/Controls/OBC/CDL/Continuous/Validation/PIDWithAutoTuning.mos
```diff
@@ -0,0 +1,14 @@
+simulateModel("Buildings.Controls.OBC.CDL.Continuous.Validation.PIDWithAutoTuning", stopTime=120, tolerance=1e-06, resultFile="PIDWithAutoTuning");
+removePlots(false);
+createPlot(id=1, position={15, 10, 900, 700}, y={"limPID.u_s", "limPID.u_m"}, range={0.0, 120.0, -5.0, 5.0}, grid=true, colors={{238,46,47}, {0,140,72}});
+createPlot(id=1, position={15, 10, 900, 700}, y={"limPID.u_s", "limPID.y"}, range={0.0, 120.0, -5.0, 5.0}, grid=true, subPlot=102, colors={{238,46,47}, {0,140,72}});
+createPlot(id=1, position={15, 10, 900, 700}, y={"limPID.k"}, range={0.0, 120.0, 0.5, 1.5}, grid=true, subPlot=103, colors={{28,108,200}}, displayUnits={"s-1"});
+createPlot(id=1, position={15, 10, 900, 700}, y=fill("", 0), range={0.0, 1.0, 0.0, 1.0}, grid=true, subPlot=104);
+createPlot(id=1, position={15, 10, 900, 700}, y=fill("", 0), range={0.0, 1.0, 0.0, 1.0}, grid=true, subPlot=105);
+createPlot(id=1, position={15, 10, 900, 700}, y=fill("", 0), range={0.0, 1.0, 0.0, 1.0}, grid=true, subPlot=106);
+createPlot(id=1, position={15, 10, 900, 700}, y=fill("", 0), range={0.0, 120.0, -2.0, 62.0}, grid=true, subPlot=107);
```
mwetter, line 9: What are these lines for?

mwetter (commented):
@SenHuang19: This requires considerable more work. Please also see the email I sent.

SenHuang19:
@mwetter Do we need to close this PR? I am not using it anymore because 1) it is merging into the master branch 2) I completely redesigned the model and it will be very hard to track the review history. A new PR has been made in #3089

mwetter:
@SenHuang19 Since the PR #3089 is also for this branch, we can leave it, and the comments should get resolved if the respective section of the code changed.

SenHuang19:
@mwetter thanks for your prompt response. I will respond to your comments in this PR.

Buildings/Controls/OBC/CDL/Continuous/AMIGOWithFOTD.mo
SenHuang19: The camelCase is enforced and the formatting is corrected.

Buildings/Controls/OBC/CDL/Continuous/AMIGOWithFOTD.mo
SenHuang19: Units are added

Buildings/Controls/OBC/CDL/Continuous/AMIGOWithFOTD.mo
```diff
@@ -0,0 +1,129 @@
...
+  Interfaces.RealInput ProcessOutput "Half-period length of a relay tuner"
+    annotation (Placement(transformation(extent={{-140,28},{-100,68}}), iconTransformation(extent={{-140,28},{-100,68}})));
+  Interfaces.RealInput RelayOutput "Half-period length of a relay tuner"
+    annotation (Placement(transformation(extent={{-140,-20},{-100,20}}), iconTransformation(extent={{-140,-20},{-100,20}})));
```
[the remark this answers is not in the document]
SenHuang19: The naming convention is enforced

Buildings/Controls/OBC/CDL/Continuous/AMIGOWithFOTD.mo
```diff
@@ -0,0 +1,129 @@
...
+        rotation=-90,
+        origin={-56,120})));
+
+  Interfaces.BooleanInput experimentEnd "If the relay experiment ends" annotation (Placement(transformation(
+        extent={{-20,-20},{20,20}},
+        rotation=-90,
+        origin={40,120}), iconTransformation(
+        extent={{-20,-20},{20,20}},
+        rotation=-90,
+        origin={54,120})));
```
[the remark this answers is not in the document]
SenHuang19: The order is enforced

Buildings/Controls/OBC/CDL/Continuous/AMIGOWithFOTD.mo
```diff
@@ -0,0 +1,129 @@
...
+        k = 1/Kp*(0.15+0.35*T/L-T^2/(T+L)^2);
+        Ti = (0.35+13*T^2/(T^2+12*T*L+7*L^2))*L;
+        Td = 0;
+     else
+        k = 1/Kp*(0.2+0.45*T/L);
+        Ti = (0.4*L+0.8*T)/(L+0.1*T)*L;
+        Td = 0.5*T*L/(0.3*L+T);
+     end if;
+  end when;
+  connect(noReset.y, Iy.trigger) annotation (Line(points={{-19,20},{8,20},{8,38}},     color={255,0,255}));
```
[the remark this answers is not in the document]
SenHuang19: Only graphical modeling (with components from CDL) is used in this package

Buildings/Controls/OBC/CDL/Continuous/NormalizedDelay.mo
SenHuang19: Only graphical modeling (with components from CDL) is used in this package

Buildings/Controls/OBC/CDL/Continuous/NormalizedDelay.mo
SenHuang19: Removed all the RealExpression blocks

Buildings/Controls/OBC/CDL/Continuous/NormalizedDelay.mo
SenHuang19: The info section has been substantially expanded.

Buildings/Controls/OBC/CDL/Continuous/Validation/AMIGOWithFOTD.mo
SenHuang19: The format is fixed

Buildings/Controls/OBC/CDL/Continuous/Validation/AMIGOWithFOTD.mo
SenHuang19: Comments are added to all the blocks of the models in this package

Buildings/Controls/OBC/CDL/Continuous/PIDWithAutoTuning.mo
SenHuang19: Comments are added to all the blocks of the models in this package

Buildings/Controls/OBC/CDL/Continuous/PIDWithAutoTuning.mo
SenHuang19: Comments are added to all the blocks of the models in this package and the format is fixed

Buildings/Controls/OBC/CDL/Continuous/PIDWithAutoTuning.mo
SenHuang19: This block has been removed and we eliminate the usage of the RealExpression block

Buildings/Controls/OBC/CDL/Continuous/Validation/PIDWithAutoTuning.mo
SenHuang19: The format is fixed

Buildings/Controls/OBC/CDL/Continuous/Validation/PIDWithAutoTuning.mo
```diff
@@ -0,0 +1,62 @@
...
+      Tolerance=1e-06,
+      __Dymola_Algorithm="Dassl"),
+    __Dymola_Commands(
+      file="modelica://Buildings/Resources/Scripts/Dymola/Controls/OBC/CDL/Continuous/Validation/PIDWithAutoTuning.mos" "Simulate and plot"),
+    Documentation(
+      info="<html>
+<p>
+Validation test for the block
+<a href=\"modelica://Buildings.Controls.OBC.CDL.Continuous.PIDWithAutoTuning\">
+Buildings.Controls.OBC.CDL.Continuous.PIDWithAutoTuning</a>.
```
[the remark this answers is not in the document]
SenHuang19: The info section is significantly expanded to include more information

Buildings/Resources/Scripts/Dymola/Controls/OBC/CDL/Continuous/Validation/PIDWithAutoTuning.mos
SenHuang19: Those lines are removed in the script

SenHuang19:
> @SenHuang19: This requires considerable more work. Please also see the email I sent.

@mwetter: I have tried my best to address your comments. Please take a look at the responses to your comments. I believe your comments in the email have also been reflected.
The new implementation has been reviewed by Jianjun and currently resides in [issue2884_PID_autotuning](https://github.com/lbl-srg/modelica-buildings/tree/issue2884_PID_autotuning).
Could you please let me know if you have more comments?

JayHuLBL:
@SenHuang19
In the composed block `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.FirstOrderAMIGO`, there is no detailed documentation of how to use the  controller. For instance, by only looking into the model, the user will not know how the parameters `yHig`, `yLow`, `deaBan`, etc. affect the results and how to set the parameters.

Also, does the controller have the difference for direct acting and reverse acting?

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/HalfPeriodRatio.mo
```diff
@@ -0,0 +1,212 @@
...
+<p>This block calculates the half-period ratio of the output from a relay controller.</p>
+<h4>Main equations</h4>
+<p align=\"center\" style=\"font-style:italic;\">
+&rho; = max(t<sub>on</sub>,t<sub>off</sub>)/ min(t<sub>on</sub>,t<sub>off</sub>),
+</p>
+<p>
+where <code>t<sub>on</sub></code> and <code>t<sub>off</sub></code> are the
+length of the On period and the Off period, respectively.
+</p>
+<p>During an On period, the relay switch signal becomes True;
```
JayHuLBL, line 174: This block does not have the relay switch input. Should this section be removed?

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/AutoTuner/AMIGO/Validation/PI.mo
```diff
@@ -0,0 +1,318 @@
+within Buildings.Controls.OBC.Utilities.PIDWithAutotuning.AutoTuner.AMIGO.Validation;
+model PI "Test model for PI"
+  Buildings.Controls.OBC.Utilities.PIDWithAutotuning.AutoTuner.AMIGO.PI PI
+    "Calculate the parameters for a PI controller"
+    annotation (Placement(transformation(extent={{-10,-10},{10,10}})));
+  Buildings.Controls.OBC.CDL.Continuous.Sources.TimeTable RefDat(table=[0,1,0.5,
```
JayHuLBL, line 6: Putting the time series data directly in the validation model should be avoid. Please see the model `Buildings.Fluid.CHPs.Validation.ElectricalFollowing` of how to add the time series data for the validation purpose.

JayHuLBL (changes requested):
@SenHuang19 Please see my inline comments.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/HalfPeriodRatio.mo
SenHuang19: Sorry for the confusion. The relay switch is referred to define the On period and the Off period. I modified the doc to avoid confusion.

SenHuang19: see #3291

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/AutoTuner/AMIGO/Validation/PI.mo
SenHuang19: I was under the impression that only the CDL components can be used in the validation. Could you please let me know if this is still the case?

SenHuang19:
> @SenHuang19 Please see my inline comments.

@JayHuLBL Could you please let me know if your earlier comments are addressed?

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/AutoTuner/AMIGO/Validation/PI.mo
JayHuLBL: For this time series data, how did you generate them? Is it possible to implement the calculation here in the validation model (use only OBC blocks)?

SenHuang19: This data is generated via a Python script which is the equivalent implementation of <code>Buildings.Controls.OBC.Utilities.PIDWithAutotuning.AutoTuner.AMIGO.PI</code>.

SenHuang19: @JayHuLBL just wanted to follow up on my previous response to see if you need more clarifications.

JayHuLBL: I cannot see if there is anyway to log the time series efficiently. However, it seems that in the whole `AutoTuner` package the calculation is relatively simple. Like in the class `PIDGain`, it basically do the calculation `k = 1/k_p + (0.2 + 0.45T/L),`. I think there is no need to verify the Modelica calculation with Python calculation. 

I suggest to just create a simple time table with much less data point (e.g. <10 points). Or, when you generate the python data point, generate less points (10 points).

SenHuang19: Very good point!. I have made https://github.com/lbl-srg/modelica-buildings/pull/3326  to address this comment

SenHuang19: @JayHuLBL could you please let me know if more changes are needed for this issue?

mwetter:
Next: @JayHuLBL to review.

JayHuLBL:
@mwetter It will be ready for your review when CI tests passed.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/AutoTuner/AMIGO/Validation/PI.mo
```diff
@@ -0,0 +1,62 @@
...
+</li>
+</ul>
+</html>", info="<html>
+<p>
+Validation test for the block
+<a href=\"modelica://Buildings.Controls.OBC.Utilities.PIDWithAutotuning.AutoTuner.AMIGO.PI\">
+Buildings.Controls.OBC.Utilities.PIDWithAutotuning.AutoTuner.AMIGO.PI</a>.
+</p>
+<p>
+The reference data is imported from a raw data that is generated with a Python implementation of this block.
```
mwetter, line 59: I don't understand this comment. There is no python implementation and such a python implementation is in my view not needed for this block.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/AutoTuner/AMIGO/Validation/PI.mo
```diff
@@ -0,0 +1,62 @@
...
+<li>
+June 1, 2022, by Sen Huang:<br/>
+First implementation<br/>
+</li>
+</ul>
+</html>", info="<html>
+<p>
+Validation test for the block
+<a href=\"modelica://Buildings.Controls.OBC.Utilities.PIDWithAutotuning.AutoTuner.AMIGO.PI\">
+Buildings.Controls.OBC.Utilities.PIDWithAutotuning.AutoTuner.AMIGO.PI</a>.
```
mwetter, line 56: We really need to stop stating the obvious (this block [which has an instance of class A] validates A). Rather, state concisely what the model tests, e.g., what is the purpose and the result of the test?
This should be changed throughout. 
I know we still have quite a few models that state "Validation test for the block XXX" but there is really no useful information in such a statement.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
mwetter, line 1: This block has many blocks which just have a white icon. Please add some graphics as this really helps understanding what these blocks are doing if the graphics is well done as block diagrams are a visual language.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/NormalizedTimeDelay.mo
```diff
@@ -0,0 +1,117 @@
...
+    annotation (Placement(transformation(extent={{0,10},{20,30}})));
+  Buildings.Controls.OBC.CDL.Continuous.MultiplyByParameter gai(
+    final k=0.35)
+    "Gain for the half period ratio"
+    annotation (Placement(transformation(extent={{-80,-70},{-60,-50}})));
+  Buildings.Controls.OBC.CDL.Continuous.Multiply mul
+    "Calculate the product of the output of addPar1 and that of addPar2"
+    annotation (Placement(transformation(extent={{0,-48},{20,-28}})));
+  Buildings.Controls.OBC.CDL.Utilities.Assert assMes(
+    final message="The asymmetry level should be larger than the half period ratio.")
```
mwetter, line 38: I assume this assertion can trigger during a simulation. Can you give some advice to the user for how to fix this situation rather than only stating what went wrong?

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/SystemIdentification/FirstOrderTimedelayed/Validation/ControlProcessModel.mo
```diff
@@ -0,0 +1,70 @@
+within Buildings.Controls.OBC.Utilities.PIDWithAutotuning.SystemIdentification.FirstOrderTimedelayed.Validation;
+model ControlProcessModel "Test model for ControlProcessModel"
+  Buildings.Controls.OBC.Utilities.PIDWithAutotuning.SystemIdentification.FirstOrderTimedelayed.ControlProcessModel
+    conProMod(yLow=0.1, deaBan=0.05)
+    "Calculate the parameters of  of a first-order model"
```
mwetter, line 5: typo

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/SystemIdentification/FirstOrderTimedelayed/Validation/ControlProcessModel.mo
```diff
@@ -0,0 +1,70 @@
+within Buildings.Controls.OBC.Utilities.PIDWithAutotuning.SystemIdentification.FirstOrderTimedelayed.Validation;
+model ControlProcessModel "Test model for ControlProcessModel"
```
mwetter, line 2: Don't use instance name in comment

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/SystemIdentification/FirstOrderTimedelayed/Validation/ControlProcessModel.mo
```diff
@@ -0,0 +1,70 @@
...
+First implementation<br/>
+</li>
+</ul>
+</html>", info="<html>
+<p>
+Validation test for the block
+<a href=\"modelica://Buildings.Controls.OBC.Utilities.PIDWithAutotuning.SystemIdentification.FirstOrderTimedelayed.ControlProcessModel\">
+Buildings.Controls.OBC.Utilities.PIDWithAutotuning.SystemIdentification.FirstOrderTimedelayed.ControlProcessModel</a>.
+</p>
+</html>"));
```
mwetter, line 69: You need to explain what the model is doing. This is not clear at all. It reads a lot of data from a table that says it is a validation data, and then uses it to validate a model, but what do you try to do?
Think about a user or developer looking at that and having to debug it. There is no chance to understand the intent of this model.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/OnOffPeriod.mo
```diff
@@ -0,0 +1,131 @@
+within Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay;
+block OnOffPeriod
+  "Calculate the lengths of the on period and the off period"
+  Buildings.Controls.OBC.CDL.Interfaces.RealInput tim(
+    final quantity="Time",
+    final unit="s")
+    "Simulation time"
+    annotation (Placement(transformation(extent={{-140,20},{-100,60}}),
+        iconTransformation(extent={{-140,40},{-100,80}})));
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanInput On
```
mwetter, line 10: Use lower case as that is the convention. Being consistent helps filtering results.

mwetter (commented):
Please see inline comments for feedback and needed changes.

mwetter:
@SenHuang19 : Please let me know if you need anything else from me regarding the above feedback.

SenHuang19:
> @SenHuang19 : Please let me know if you need anything else from me regarding the above feedback.

@mwetter Thanks for checking with me. I was just back from a person leave this week and started to work on it yesterday. I will let you know if I need any more guidelines.

mwetter:
> > @SenHuang19 : Please let me know if you need anything else from me regarding the above feedback.
> 
> @mwetter Thanks for checking with me. I was just back from a personal leave this week and started to work on it yesterday. I will let you know if I need any more guidelines.

Great, thanks for the update.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -0,0 +1,303 @@
+within Buildings.Controls.OBC.Utilities.PIDWithAutotuning;
+block FirstOrderAMIGO
+  "A autotuning PID controller with an AMIGO tuner and a first order time delayed system model"
+  parameter Buildings.Controls.OBC.CDL.Types.SimpleController controllerType=Buildings.Controls.OBC.CDL.Types.SimpleController.PI
+    "Type of controller";
```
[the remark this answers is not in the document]
SenHuang19: This seems to be resolved by cf11ed882851670684cf615f4adb440edef8f73f

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/Controller.mo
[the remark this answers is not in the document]
SenHuang19: Added one Boolean input to determine when the auto tuning starts.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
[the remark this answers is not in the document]
SenHuang19: Added <code>triTun</code>.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/AutoTuner/AMIGO/PIDGain.mo
```diff
@@ -0,0 +1,107 @@
...
+    final quantity="Time",
+    final unit="s",
+    min=100*Buildings.Controls.OBC.CDL.Constants.eps)
+    "Time delay of a first order time-delayed model"
+    annotation (Placement(transformation(extent={{-140,-80},{-100,-40}}),
+        iconTransformation(extent={{-140,-80},{-100,-40}})));
+  Buildings.Controls.OBC.CDL.Interfaces.RealOutput k
+    "Control gain signal"
+    annotation (Placement(transformation(extent={{100,-20},{140,20}})));
+  Buildings.Controls.OBC.CDL.Continuous.Divide div1
```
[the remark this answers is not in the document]
SenHuang19: Those blocks have been made <code>protected</code>. The same principle has been applied to other modules as well.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/AutoTuner/AMIGO/Validation/PI.mo
SenHuang19: The <code>reference data</code> has been removed.
All the validation example has been redesigned and documented.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/NormalizedTimeDelay.mo
SenHuang19: The advice has been added to the <code>message</code>.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/SystemIdentification/FirstOrderTimedelayed/Validation/ControlProcessModel.mo
SenHuang19: The typo has been fixed.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/SystemIdentification/FirstOrderTimedelayed/Validation/ControlProcessModel.mo
SenHuang19: Replaced all the instance names in comment.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
SenHuang19: Icons are added to the main blocks.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/SystemIdentification/FirstOrderTimedelayed/Validation/ControlProcessModel.mo
SenHuang19: The validation model is redesigned to eliminate the table and the model doc is updated

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/AutoTuner/AMIGO/Validation/PI.mo
SenHuang19: I kept the statement "Validation test for ..." but added more descriptions to the validation example.
Hope the updated version meet your requirement.

SenHuang19:
> > > @SenHuang19 : Please let me know if you need anything else from me regarding the above feedback.
> > 
> > 
> > @mwetter Thanks for checking with me. I was just back from a personal leave this week and started to work on it yesterday. I will let you know if I need any more guidelines.
> 
> Great, thanks for the update.

@mwetter can you please take a look at the updated model based on your comments?

JayHuLBL:
@SenHuang19 I see the CI tests are failing. I can fix some of them and will let you know if there is any help from you.

SenHuang19:
> @SenHuang19 I see the CI tests are failing. I can fix some of them and will let you know if there is any help from you.

@JayHuLBL Thanks a lot! Please do let me know the issue so I can avoid it in the future.

SenHuang19:
@JayHuLBL thank you so much for you edits. There are indeed a lot of details to be addressed. Can you please let me know if you need me to do anything at this point?

JayHuLBL:
@SenHuang19

I did some editing on the documentation and comments. However, I have some additional comments:
- [ ] For the class `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.BaseClasses.Validation.HalfPeriodRatio`,
   - the documentation indicates that at 0.1 s, the output switches from On to Off. So What `output` are you referring? Do you mean `triSta` becomes `true`? However in fact, both of these happen at 0.2 s.
   - the documention indicates that at 0.7 s, the output switches to On. But in the plot, there is no change either `triSta` or `triEnd`.
   - in the documentation of `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.BaseClasses.HalfPeriodRatio`, it indicates that `The tuning period is triggered to end when either tOn or tOff changes after they become positive`. However, in the test model, at 0.7 s, the `tOn` changes from 0.1 to 0.7 but the `triEnd` remains `false`.
- [ ] For the class `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.BaseClasses.Validation.OnOffPeriod`, the `tOn` remain zero even though the `on` is `true`. According to the documentation in `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.BaseClasses.OnOffPeriod`, it calculates the length of the on perid. So, shouldn't the `tOn` output will increase from 0 to 0.06. rather than remains 0 and step changes to 0.06?
- [ ] it would be better to improve the comments for `trigger` and `yOn` in `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.Controller`. I think the `trigger` is `True: enables the relay controller` and the `yOn` is `True: tuning starts`.
- [ ] in class `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.ResponseProcess`, what is the difference between `yOn` and `triSta` and `triEnd`? I think
    - when `yOn` is `true`, the `triSta` is `true` and `triEnd` is `false`.
    - when `yOn` is `false`, the `triSta` is `false` and `triEnd` is `true`.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -0,0 +1,322 @@
...
+This block is built based on
+<a href=\"modelica://Buildings.Controls.OBC.Utilities.PIDWithInputGains\">
+Buildings.Controls.OBC.Utilities.PIDWithInputGains</a>
+and inherits most of the parameters of the latter. However, through the parameter
+<code>controllerType</code>, the controller can only be configured as PI or
+PID controller.
+In addition, the output of this block is limited from 0 to 1.
+</p>
+<p>
+Note that the autotuning can be performed only once.
```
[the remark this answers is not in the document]
SenHuang19: I limit it to be only once to avoid the following scenario:
- the autorunning has not been completed yet but the <code>triTun</code> becomes True.

Please let me know your thoughts on it.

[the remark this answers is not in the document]
JayHuLBL: One scenario may need to double check:
- [ ] If `triTun` is `true` and `resPro.triEnd` is `false` (means autotuning has not been completed yet but the `triTun` becomes True), then `or1.y` will be `false`. The `y` will be `swi.u3`, which equals to `rel.y`. However, the `rel.y` would not equal to `yHig`, in block `PIDWithAutotuning.Relay.Controller`, as the `trigger` input in the block is `true`.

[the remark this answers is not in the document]
SenHuang19: Yes, <code>rel.y</code> may not be equal to <code>yHig</code>.
Can you please let me know if <code>rel.y</code> needs to be equal to <code>yHig</code>?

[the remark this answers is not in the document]
JayHuLBL: Is there a scenario that the `triTun` becomes `false` but the `resPro.triEnd` is `false` (means the autotuning triggering signal becomes `false` but the tuning has not been completed), or anytime when the `triTun` becomes `false`, the `resPro.triEnd` will become `true`?
Is it fine to change the text as below?
```
Note that each time when the tuning triggering input `triTun` becomes `true`, the autotuning can be performed and performs only one time.
```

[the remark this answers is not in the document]
SenHuang19: The current implementation assumes that <code>triTun</code> keeps <code>true</code> after it becomes <code>true</code>.
The scenario you described is unexpected.
The text looks good but do we need to emphasize that <code>triTun</code> needs to keep <code>true</code>?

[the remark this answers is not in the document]
JayHuLBL: so will it be possible to retune the controller in the middle of the system operation? what is the typical scenario that the controller should be tuned? I basically wants to understand where the input `triTun` receives the signal.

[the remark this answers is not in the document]
SenHuang19: Retuning is possible only when the ongoing tuning is completed.
That means when we can retune can't be predefined.
We could add a Boolean output to indicate if the retuning window opens.

I thought the main reason to add this <code>triTun</code> is to avoid tuning this PI during the unoccupied period.
In this case, the <code>triTun</code> should be connected with the occupied signal.
Please let me know your thoughts.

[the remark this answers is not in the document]
JayHuLBL: Is it possible to re-trigger the tuning process? For current implementation, at the first time when `triTun` changes from `false` to `true`, the `resPro.triEnd` is `false` (is it correct?), then the `swi.y` equals to `rel.y`, which I think the tuning process starts. 
After the tuning is done, the `resPro.triEnd` becomes `true` and keeps being `true`. And then the `triTun` changes to be `false`. (Will this make the `resPro.triEnd` becomes `false`?)
If the `resPro.triEnd` keeps being `true`, it will be not possible to re-trigger the tuning process.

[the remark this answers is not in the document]
SenHuang19: Yes, you are right. The current implementation doesn't allow to re-trigger the tuning process. I didn't fix this issue mainly because we couldn't predict when the tuning is done. I thought it basically eliminated the possibility for performing re-tuning.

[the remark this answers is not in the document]
mwetter: The implementation must allow re-tuning. We want to be able to trigger it from an external block whenever a control loop has poor performance.

JayHuLBL:
@mwetter It will be ready for your review when the tests are passed.

JayHuLBL:

mwetter (commented):
In `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.PIWithFirstOrderAMIGO`, the value of `PIWitTun.inTunPro.y` does not reset after the second tuning when simulated in OpenModelica (OpenModelica 1.23.0~dev-59-g7a41456):

image

This then leads to different results when simulated in OpenModelica. With OCT, this quantity looks the same as in Dymola. This may be related to the `Latch` block.

Also, the following needs to be address:
- [ ] Add a test with a model for `reverseActing=false`. Otherwise we don't know if the model is correct for this setting.
- [ ] Add a test that uses it with an actual control loop, such as `WetCoilCounterFlowPControl` which could be copied and adjusted to provide a realistic example.
- [ ] Explain why in `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.PIDWithFirstOrderAMIGO`, the quanity `PIDWitTun.con.Td` differs after the first and second tuning period. Shouldn't it be the same?

image

@JayHuLBL : Can you please address these issues.

SenHuang19:
@mwetter there seems to be a bug which causes the following issue.
> * [ ]  Explain why in `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.PIDWithFirstOrderAMIGO`, the quanity `PIDWitTun.con.Td` differs after the first and second tuning period. Shouldn't it be the same?

Do you want to me to work on this?

mwetter:
@SenHuang19 : It would be great if you can work on this. I wonder if the plant has too low an order so that the identification is indeterminate?

SenHuang19:
> @SenHuang19 : It would be great if you can work on this. I wonder if the plant has too low an order so that the identification is indeterminate?

@mwetter 

Hi Michael, initially I thought this was caused by the fact that some samplers are not re-initialized but I believe you are right.

Specifically, it is associated with

image

When <code>τ</code> is close to 1,  <code>L/T</code> is very sensitive to the change of <code>τ</code>.

image

Should I re-design the validation case to avoid this scenario? Maybe just use the following case?

> * [ ]  Add a test that uses it with an actual control loop, such as `WetCoilCounterFlowPControl` which could be copied and adjusted to provide a realistic example.

mwetter:
@SenHuang19 : I suggest you redesign the validation case, as having a small simple validation case is useful. You may want to see if the 1st order room model that is used in the OptimalStart validation works better; and/or add two first order filters rather than the discrete time delay at the output of the autotuning controller. Such filters would approximate a series of control volumes.

JayHuLBL:
The results difference between OpenModelica and Dymola is due to the results different of `PIDWitTun.resPro.TunMon.tInc.us`. 
See the one from OpenModelica:
re
And from Dymola:
re

SenHuang19:
> The results difference between OpenModelica and Dymola is due to the results different of `PIDWitTun.resPro.TunMon.tInc.us`. See the one from OpenModelica: re And from Dymola: re

Should I add a re-init to `resPro.TunMon.samAddtOntOff`?

JayHuLBL:
The root cause of the OpenModelica and Dymola results difference is from the instance `PIDWitTun.resPro.TunMon.gretOnAndtOff`.

See the OpenModelica results:
Omc-1
Omc-2

And the Dymola results:
dymola

In both simulation, the `CDL.Reals.Greater` has the same inputs (`u1` and `u2`), with hysteresis set to be zero. The Dymola gives expected results, but the OpenModelica gives wrong output. This could be an issue in OpenModelica.

mwetter:
@JayHuLBL : If it is an issue with OpenModelica, can you submit a bug report with instructions for how to reproduce it.

JayHuLBL:
@mwetter I will submit a bug report to OpenModelica.

JayHuLBL:
There was two discussion about the OpenModelica issue, in their ticket [8600](https://github.com/OpenModelica/OpenModelica/discussions/8600) and [11049](https://github.com/OpenModelica/OpenModelica/issues/11049), and the issue we created [11468](https://github.com/OpenModelica/OpenModelica/issues/11468#issuecomment-1785664197). The comparison between two real variables would cause results difference between OpenModelica and Dymola. Two avoid the issue, we could change the input `u2` of `PIDWitTun.resPro.TunMon.gretOnAndtOff` from previous `0` to `1e-5`, which is usually greater than the solver's tolerance. The test shows there is no more different `PIDWitTun.resPro.TunMon.gretOnAndtOff.y` .

However, the unit test with OpenModelica will still give difference results at the very beginning of the simulation, see below:
Screenshot 2023-10-27 at 3 13 16 PM

Both the instances `uniDel1` and `derivative1` have parameter `y_start` and they have been default set to `0`. Dymola did not show the start value but the OpenModelica showed it.

mwetter:
@JayHuLBL : Did you test whether your correction leads to the same results, other than the initialization? I still see substantial differences when I run the unit test with OpenModelica (1.23.0~dev-103-gd230b4a) toward the end of the simulation, for example for `Validation.PIWithFirstOrderAMIGO`.

@SenHuang19 : Please add in the model `info` section of `TuningMonitor` what the `mul` instance is doing. It multiplies two inputs with units of time, so the output is `s^2`. That does not seem to be right, or it needs a clear explanation.

JayHuLBL:
@mwetter I further improved the real comparison in the class `Relay.BaseClasses.TuningMonitor`, now we can only see the results difference during the initialization.

SenHuang19:
> @JayHuLBL : Did you test whether your correction leads to the same results, other than the initialization? I still see substantial differences when I run the unit test with OpenModelica (1.23.0~dev-103-gd230b4a) toward the end of the simulation, for example for `Validation.PIWithFirstOrderAMIGO`.
> 
> @SenHuang19 : Please add in the model `info` section of `TuningMonitor` what the `mul` instance is doing. It multiplies two inputs with units of time, so the output is `s^2`. That does not seem to be right, or it needs a clear explanation.

I will update the model `info` section to provide a clear explanation.

mwetter:
@JayHuLBL : The result differences during the initialization are fine. Looking at the csv file from OpenModelica shows for example 
time | PI.u_m
-- | --
0 | 0
0 | 0
0 | 0.72

for `Validation/PIWithFirstOrderAMIGO.mo`

JayHuLBL:
@SenHuang19 will add more validations. There will be another PR to address the comments.

JayHuLBL:
@SenHuang19 Would you please check the class `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.Controller`? First check if the conditions documented is correct. I think the assignment of the `yErr` is correct, but not sure if the `y` and `yOn` are correct.
Screenshot 2023-12-14 at 4 07 09 PM

The reason why I have this doubt is that in your implementation when using `OnOffController`, the output may not be the desired one. I reimplemented to avoid using `OnOffController` (with `Subtract` and `Hysteresis`), but it seems to be opposite to the documentation.
Screenshot 2023-12-14 at 4 03 08 PM

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/Controller.mo
```diff
@@ -0,0 +1,229 @@
...
+    iconTransformation(extent={{-20,-20},{20,20}},rotation=90,origin={-60,-120})));
+  Buildings.Controls.OBC.CDL.Interfaces.RealOutput y
+    "Control output"
+    annotation (Placement(transformation(extent={{100,30},{140,70}}),
+        iconTransformation(extent={{100,40},{140,80}})));
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanOutput yOn
+    "Relay switch output, true when control output switches to the higher value"
+    annotation (Placement(transformation(extent={{100,-80},{140,-40}})));
+  Buildings.Controls.OBC.CDL.Interfaces.RealOutput yErr
+    "Control error"
```
JayHuLBL, line 36: in the code implementation, the `yErr = u_m - u_s`, but in the documentation, the `yErr` could be `u_m - u_s` or `u_s - u_m`. Would you please confirm if the one in the code should be the same as described in the documentation? If it should be different, then we need to differentiate them.

JayHuLBL:
@SenHuang19 
Run the example class `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControlAutoTuning`, it shows that the controller is not working properly:
WetCoilCounterFlowPControlAutoTuning-2
However, running the class `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControl` gives reasonable result:
WetCoilCounterFlowPControl

JayHuLBL:
@mwetter it's ready for your review.

JayHuLBL:
@SenHuang19 
- [ ] better explanation of how to set the parameters, like `yHig`, `yLow`, `yRef`, and `deaBan`. We may update the comments for each parameter, or better documentation. Users cannot understand how these parameters affect the control performance. 
- [ ] the default value of parameter `deaBan` is inconsistent. In `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.FirstOrderAMIGO`, the value is 0.1. While in the subblocks like in `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.Controller` (and other blocks), the value is 0.5.
- [ ] Make sure the PID controller with autotunning, can use to replace the standard PID controller.
  - the controller instantiates the PID controller with input gains (`con`). The `con.yMax` and `con.yMin` are set to be 1 and 0 and have been finalized. Can we not finalize the `yMax` and `yMin`? Does it mean we need to update the implementation?
- [ ] in the example model, `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControlAutoTuning`, it's not necessary to use `gai1` and update the `con.r` to the appropriate value. For the `con.deaBan`, why should we use `1/298.15`? It would be useful after you address the comment 1, and so the user then knows how to set the values.

JayHuLBL:
@SenHuang19 Please let me if any help is needed to address the comments. We'd like to close this PR soon if possible.

SenHuang19:
> @SenHuang19 Please let me if any help is needed to address the comments. We'd like to close this PR soon if possible.

@JayHuLBL thanks for checking with me. I am working on another PR which is our milestone for this quarter and plan to work on it next week.  I believe I have all I need to proceed and should be able to make a PR by next week. Does this timeframe work for you?

JayHuLBL:
@SenHuang19 Thanks for the update. The plan looks good.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -0,0 +1,503 @@
...
+     final min = 0,
+     final max = 1) = 1
+    "Higher value for the relay output";
+  parameter Real yLow(
+     final min = 0,
+     final max = 1)
+    "Lower value for the relay output";
+  parameter Real deaBan(
+    final min=1E-6)
+    "Deadband for holding the output value";
```
JayHuLBL, line 31: Is it `Deadband for holding the relay output`?

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -0,0 +1,503 @@
...
+<br>
+The following procedure can be used for determining the values of those parameters. 
+</p>
+<ol>
+<li>
+The <code>r</code> should be adjusted so that the output of the relay controller,
+<code>rel.yDif</code>, is within the range from 0 to 1.
+</li>
+<li>
+The <code>yRef</code> can be determined by dividing the set point by the sum of the
```
JayHuLBL, line 368: How do use know the set point and the minimum and maximum values of the measurement? Both the set point and measurement are inputs.
Do you mean the typical values?

JayHuLBL:
@SenHuang19 See inline comments.

Buildings/Fluid/HeatExchangers/Examples/WetCoilCounterFlowPControlAutoTuning.mo
```diff
@@ -0,0 +1,213 @@
...
+    duration=2000) "Water temperature, raised to high value at t=3000 s"
+    annotation (Placement(transformation(extent={{-80,54},{-60,74}})));
+  Buildings.Controls.OBC.CDL.Logical.Sources.Constant resSig(k=false)
+    "Reset signal"
+    annotation (Placement(transformation(extent={{-80,160},{-60,180}})));
+  Buildings.Controls.OBC.CDL.Logical.Sources.Pulse autTunSig(
+    width=0.1,
+    period=4000,
+    shift=400)
+    "Signal for enabling the autotuning"
```
JayHuLBL, line 115: If we change the value as 
```
width=0.2,
period=2000,
shift=100
```
The simulation will fail with the error:
```
Integration started at T = 0 using integration method DASSL
(DAE multi-step solver (dassl/dasslrt of Petzold modified by Dassault Systemes))
Error: The following error was detected at time: 2493.834817548029

Model error - Modelica.Math.log (con.conProMod.timConDel.log.u) = Modelica.Math.log (-25.3528)
```

This is the error due to the instance `log` in block `PIDWithAutotuning.SystemIdentification.FirstOrderTimedelayed.BaseClasses.TimeConstantDelay`, as the block cannot have a negative input.

SenHuang19: Yes, this autotuning process employs an empirical equation:
image.
Since multiple variables are involved, avoid this error seems to be difficult.
Should we discuss it in our meeting next week?

SenHuang19: Based on the discussion on 6/12/2024

- [ ]  Add an assert to check if the setpoint changes during the autotuning process. If the setpoint changes, stop the autotuning. In this case, the autotuning completes unsuccessfully.
- [ ]  Add a Boolean output to indicate if the autotuning completes successfully or not.
- [ ]  If the autotuning completes unsuccessfully, use the gains before the autotuning starts

SenHuang19: With the updated model, the simulation output is shown below:

image

image

Specifically, at time 2100 seconds, the model detects an abnormal change in the setpoint;
As a result, the autotuning fails at ~2493 seconds, no changes are made to the control gains.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -0,0 +1,555 @@
...
+The PID tuning process ends automatically
+(see details in
+<a href=\"modelica://Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.BaseClasses.TuningMonitor\">
+Buildings.Controls.OBC.Utilities.PIDWithAutotuning.BaseClasses.Relay.TunMonitor</a>),
+at which point this block turns back to a PID controller but with tuned PID parameters.
+</p>
+<p>
+<b>Note:</b> If an autotuning is ongoing, i.e., <code>inTunPro.y = true</code>,
+a new request for performing autotuning will be ignored.
+In addition, the set point should not be changed during the autotuning process.
```
JayHuLBL, line 400: What will happen if the setpoint is changed during the tuning process? I see that you have a assert block`assMes3`. The assertion will give a warning not the error, so the tuning/simulation will not stop. So, should the tuning stop?

If the intention is to use the setpoint value at the moment when the tuning request becomes true, you may use the sampling value (`sam_u_s.y`) as input for the tuning process. If during the process the setpoint changes, it would give warning, but the tuning still use the old setpoint value. However, the question still is, should we allow the setpoint change?

SenHuang19: Yes, we could stop the tuning once the setpoint changes. However, I am not sure if changing setpoint is the only reason for the square root negative number. Thus, the current implementation stops the tuning once the square root negative number occurs.

JayHuLBL: So I am assuming that when there is setpoint change during the tuning process, current implementation will stop the tuning process (no matter if or not the setpoint change causes the negative square root number)?

However, when check the result of `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControlAutoTuning`, the variable `con.inTunPro.y` has following profile:
Screenshot 2024-08-22 at 2 14 37 PM

It shows that although the setpoint `con.u_s` is changed at 2400 seconds, but the  `con.inTunPro.y` keeps being `true` till to 2493.84 seconds, meaning that the tuning process does not stop at 2400 seconds. Am I missing something?

SenHuang19: Sorry for the confusion.
In the current implementation,
- The tuning doesn't stop if the setpoint changes. A warning will be triggered.
- When the tuning stops, we will check if the tuning has been completed successfully or not. If not, we will use the control gains before the tuning.
- The flag that determines whether the tuning completes successfully or not will be false when the negative square root number occurs.

In my previous reply, I was saying we could change the current implementation into:
- The tuning stops if the setpoint changes. In this case, we will use the control gains before the tuning.

But my concern is that we cannot handle the case where the negative square root number is caused by other reasons.

Let me know if we need to have a quick call to discuss it.

SenHuang19: 08/26/2024 meeting note:

-  Need to stop the tuning process once the setpoint changes. In this case, we will use the control gains before the tuning.

SenHuang19: Changes have been made in #3987  to stop the tuning process if the setpoint changes. In such a case, control gains will not be changed. If the setpoint is constant during the tuning, the tuning will be processed without interruption. We will check if the negative square root number occurs once the tuning is completed. If so, the tuning is considered to fail and control gains will not be changed.

JayHuLBL:
@SenHuang19 Run the model `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControlAutoTuning`, we have the results as below:
Screenshot 2024-11-19 at 10 50 26 PM
It seems that although the tuning process is triggered at 100 seconds and 2100 seconds, the control gain `con.con.k` and `con.con.Ti` constantly equal the start values. Could you please double check the results?

SenHuang19:
> @SenHuang19 Run the model `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControlAutoTuning`, we have the results as below: Screenshot 2024-11-19 at 10 50 26 PM It seems that although the tuning process is triggered at 100 seconds and 2100 seconds, the control gain `con.con.k` and `con.con.Ti` constantly equal the start values. Could you please double check the results?

@JayHuLBL Before merging the master branch, the results appear to be fine. I will thoroughly review it.

SenHuang19:
> > @SenHuang19 Run the model `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControlAutoTuning`, we have the results as below: Screenshot 2024-11-19 at 10 50 26 PM It seems that although the tuning process is triggered at 100 seconds and 2100 seconds, the control gain `con.con.k` and `con.con.Ti` constantly equal the start values. Could you please double check the results?
> 
> @JayHuLBL Before merging the master branch, the results appear to be fine. I will thoroughly review it.

@JayHuLBL 

- The control gains, con.con.k and con.con.Ti, remain constant at the initial values because the first tuning attempt fails, as detailed in the log file. 
- The failure of the initial tuning is due to recent modifications to the valve, which appear to reduce the delay effects. 
- I made several adjustments to the example to ensure it works, including minor changes to the valve parameters and those of the relay controllers (see #4054) based on the updated response.

mwetter:
The error with the license check occasionally occurs when the license manager takes too long to start and in the meantime Dymola proceeded without checking out the license. Restarting the CI tests helps.

JayHuLBL:
@mwetter Thanks! It now passed the test. I will have another look and then let you know if it is ready for your review.

JayHuLBL:

mwetter (commented):
@SenHuang19 @JayHuLBL : 
This PR still has a few issues. First, note that I deleted the parameter `u_s_start` as it has no effect. The start value is overridden as soon as the tuning is initiated. So I set the start value of the sampler to 0.

- [ ] A minor thing to address is the default value for `yHig` and `yLow`. The parameter `yHig` has a default of `1` but `yLow` has no default. This is inconsistent. Is there a rational for it? If so it needs to be documented.
- [ ] The documentation of `yLow` and `yHig` needs to be improved. It is not clear what they are used for.
- [ ] The "guidance for setting the parameters" is not clear. This needs to be made clearer.
- [ ] The main issue is that the example `WetCoilCounterFlowPIControlAutoTuning`, formerly called `WetCoilCounterFlowPControlAutoTuning`, does not work: The autotuning is interrupted and the message is `Autotuning failed. The controller gains are unchanged`. The only change I think compared to the older version is that the stroke time of the valve is now 120 s (the default) and not 240 s. Changing the stroke time is not a design choice in HVAC control. This change was somewhat accidental as I introduced a base class, but I have not seen a valve with a stroke time larger than 120 s. I am surprised why a faster process fails the tuning, as it should be easier to control. This hints towards some bug for which there needs to be an explanation and/or a better example.

SenHuang19:
> @SenHuang19 @JayHuLBL : This PR still has a few issues. First, note that I deleted the parameter `u_s_start` as it has no effect. The start value is overridden as soon as the tuning is initiated. So I set the start value of the sampler to 0.
> 
> * [ ]  A minor thing to address is the default value for `yHig` and `yLow`. The parameter `yHig` has a default of `1` but `yLow` has no default. This is inconsistent. Is there a rational for it? If so it needs to be documented.
> * [ ]  The documentation of `yLow` and `yHig` needs to be improved. It is not clear what they are used for.
> * [ ]  The "guidance for setting the parameters" is not clear. This needs to be made clearer.
> * [ ]  The main issue is that the example `WetCoilCounterFlowPIControlAutoTuning`, formerly called `WetCoilCounterFlowPControlAutoTuning`, does not work: The autotuning is interrupted and the message is `Autotuning failed. The controller gains are unchanged`. The only change I think compared to the older version is that the stroke time of the valve is now 120 s (the default) and not 240 s. Changing the stroke time is not a design choice in HVAC control. This change was somewhat accidental as I introduced a base class, but I have not seen a valve with a stroke time larger than 120 s. I am surprised why a faster process fails the tuning, as it should be easier to control. This hints towards some bug for which there needs to be an explanation and/or a better example.

Thank you so much for the edits and comments. I will create a PR to address the issues.

SenHuang19:
@mwetter @JayHuLBL 
Hi Michael and Jianjun, just report some preliminary findings from my investigation on the example `WetCoilCounterFlowPIControlAutoTuning`.

The tuning failure is due to a symmetric response from the controlled system. An asymmetric response is typically easier to achieve during the transient period at the start of the simulation. When the valve stroke time is 240s, the transient period is longer, resulting in an overlap with the tuning period. In this case, the tuning completes successfully.
image

However, when the valve stroke time is reduced to 120s, this overlap no longer occurs, causing the tuning to fail.
image

I attempted to adjust the `yHig` and `yLow` parameters to modify the response when the valve stroke time is reduced to 120s, and it appears to be working now.
image

SenHuang19:
> @SenHuang19 Would you please let me know if you had a chance to address the comments?

@JayHuLBL Yes, I’ve addressed all the issues and plan to do some proofreading today and tomorrow before submitting the PR.

JayHuLBL:

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -0,0 +1,649 @@
...
+</li>
+<li>
+If the set point is changed during an autotuning process, a warning will be
+generated. The ongoing tuning process will be halted, and no adjustments will be made
+to the PID parameters.
+</li>
+<li>
+The autotuning must be conducted when the process is in a stable state.
+The user should monitor changes in the independent variables and
+the controlled variables (e.g., mass flow rate, temperature) over time.
```
mwetter, line 466: @SenHuang19 : The documentation from here on and below uses "controlled variable", "independent variable" and "measurement". I think you mean by "controlled variable" the "measurement". Can you please use the same term for the same thing. Perhaps easiest is to use "controller input" and "controller output".

Afterwards this is ready to merge.

mwetter (commented):
@SenHuang19 : Please see inline note for one last clarification in the documentation.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
SenHuang19: @mwetter: Yes, the "controlled variable" is the same as the "measurement". I'll create a PR to update the document according to your suggestion.

mwetter (commented):
This is ready to merge.

JayHuLBL (approved):

PR #2968, opened 2022-04-15
Remove model and associated files

### LINKED ISSUE #2967: Incorrect hydronics in Experimental.DHC.Examples.Combined.ParallelConstantFlow
Simulating the model yields a negative ETS pump power due to incorrect hydronic principles.
The model is indeed a blunt adaptation of `SeriesConstantFlow` to a two-pipe network, developed primarily to illustrate the use of `Experimental.DHC.Networks.Combined.UnidirectionalParallel`. 
However, the hydronic layout of the ETS (as presented in Sommer 2020) is suited for "passive networks" where the ETS is not exposed to a high differential pressure.
For active networks (as it is the case here) a control valve should be used instead of a pump for the heat pump source circuits.
In addition, as an ideal pump model is used (that takes mass flow rate as an input) the model logically computes a negative flow work and a negative power.
I suggest removing that model from the library (as it may also be confused with the bidirectional network presented in Sommer 2020).
Maybe an assert statement would also be useful in `Fluid.Movers.FlowControlled_m_flow` to generate a warning in case of a negative flow work.

AntoineGautier:
This closes #2967.


mwetter (approved):
@AntoineGautier : Thanks for the PR. Can you please also remove these files on the `maint_8.1.x` branch. Technically this change is not backwards compatible, but I think it is fine as it is in the `Experimental` package and only affects an example.

AntoineGautier:
@mwetter Sure, I will make those changes in `maint_8.1.x` too.

PR #2971, opened 2022-04-18
Issue2905 Steam DH Full Systems - CI Test Errors

khinkelman:
Hi @bravache,

These changes address the comments in [PR #2961 Comment 1099548714](https://github.com/lbl-srg/modelica-buildings/pull/2961#issuecomment-1099548714). Here is a summary of the changes:

- The initial condition warnings from `*.SingleBoiler` are no longer present. To address: I added a Boolean parameter `fixed_p_start` for the `ControlVolumeEvaporation` base class to optionally set the initial pressure condition as fixed. For example, the default `fixed=false` is retained in the `ControlVolumeEvaporation` and `Boiler` example models where the pressure state is defined through the direct connection with a pT boundary component. While in the steam plant, we set `fixed=true` because this fixed boundary is not present, and this condition was required to solve the error/warning. 
- I revised the base class `Networks.BaseClasses.PartialConnection2Pipe2Medium` to only include the common models, and the connection pipe declarations have been added to higher levels. This was needed because `*.DistributorCollector` requires only one (lossless) pipe declaration on the connection side to be nonsingular, while two pipe declarations (supply & return) are possible for other network connection models. 
- I set `T_start` to be `final` in `PartialSaturatedControlVolume` because the model is constrained at a saturated state. Thus, only setting `p_start` is required.

With the above changes, the unit tests for all DHC models now run without errors with Dymola. (I don't yet have Optimica/OpenModelica set up, so I was unable to verify with those tools.)
One last note however: My unit tests did present some small errors with tolerances being exceeded for several DHC models developed by Antoine. I assume this is due to a small difference in my workstation/software configuration compared to the CI Docker setup. As such, it is possible that the `Steam.*.mo` models I developed may present small tolerance differences as well; if you could please rerun the unit tests and update the results as needed, I would greatly appreciate it.

If you have any questions, please let me know. Thank you again very much for your timely help and attention!


bravache (approved):
Thanks for working on that so quickly Katy!

PR #3002, opened 2022-05-11
Issue2905 steam dhs full sys patch single boiler

mwetter:
@dhblum @khinkelman : This allows `Buildings.Experimental.DHC.Plants.Steam.Examples.SingleBoiler` to compile in OpenModelica.
Note that the `constrainedby` clause was missing. Without it, the parameter assignments are not preserved (some tools preserve the medium assignment, but it is not required without the `constrainedby` clause).

Simulation fails with `division by zero at time 0, (a=0) / (b=0), where divisor b expression is: 0.0`. I think this is an open issue in OpenModelica as we have other fluid models with this behavior.

I also changed the expansion vessel pressure assignment from `p` to `p_start`, as `p` turns out to be not used and will be removed.

@dhblum : Please go ahead and merge it to your development branch if you agree with the changes, and delete this branch `issue2905_steamDHS_fullSys...issue2905_steamDHS_fullSys_patch_SingleBoiler`

### DIFF
--- a/Buildings/Experimental/DHC/Networks/BaseClasses/PartialConnection2Pipe2Medium.mo
+++ b/Buildings/Experimental/DHC/Networks/BaseClasses/PartialConnection2Pipe2Medium.mo
@@ -9,16 +9,18 @@ partial model PartialConnection2Pipe2Medium "Partial model for connecting an
     "Medium model for return fluid";
 
   replaceable model Model_pipDisSup =
-      Buildings.Fluid.Interfaces.PartialTwoPortInterface (
-    redeclare final package Medium = MediumSup,
-    final m_flow_nominal=mDis_flow_nominal,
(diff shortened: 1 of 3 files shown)


dhblum:
@mwetter Thanks for the patch.  Interestingly, ``Buildings.Experimental.DHC.Plants.Steam.Examples.SingleBoiler`` does not have any problem simulating with OpenModelica 1.20.0~dev-29-g07dc981 before your patch.  Your patch does not change this, but good that it helps with translation with the earlier version.

Also, are you saying ``p`` will be removed as a parameter of the expansion vessel component model in a coming Buildings release?

mwetter:
Yes, p will be removed, this is now in the CI testing. You can use p_start
instead of p, which is what the model already uses.

dhblum:
@mwetter Ok I see sounds good, thanks.

Issue #3026, opened 2022-05-20
OCT error when define a single pane window


JayHuLBL:
In class `Buildings.HeatTransfer.Windows.BaseClasses.CenterOfGlass`, when the number of glass layer `nGlaLay` is 1, there is no gas layer. So the `Buildings.HeatTransfer.Windows.BaseClasses.GasConvection gas[nGlaLay-1]` has size 0. In Dymola (version 2022), it will be seems as disabled instance. However, it would trigger error in OCT
```
Error at line 37, column 15, in file '/home/jianjun/GitFolder/modelica-buildings/Buildings/HeatTransfer/Windows/BaseClasses/CenterOfGlass.mo':
  Array size mismatch in modification of gas, expected size is [0] and size of binding expression is [1]
```
The parameter `glaSys` has type `Buildings.HeatTransfer.Data.GlazingSystems.Generic`, in which the vector `gas` in default has one element `Buildings.HeatTransfer.Data.Gases.Air(x=-1)`.

Because of this setting, when we define a single pane window as 
```
record Win670 =
    Buildings.HeatTransfer.Data.GlazingSystems.Generic (
    final glass={Buildings.ThermalZones.Detailed.Validation.BESTEST.Data.Glass600()},
    UFra=1.4) "Single pane, clear glass 3.048mm"
```
It will trigger the OCT error as above.

**To Reproduce**
Run model `Buildings.ThermalZones.Detailed.Validation.BESTEST.Cases6xx.Case670` from branch issue3005_BEST_update (commit 423b95f7365d27cbe068c7e6820c59da4a5b3b21) with the OCT. It should have error as above.

**Possible solution**
declare a parameter as 
```
final parameter Boolean have_GasLay = nGlaLay > 1
    "True if it has gas layer";
```
and make the `gas` to be conditional:
```
  Buildings.HeatTransfer.Windows.BaseClasses.GasConvection gas[nGlaLay-1](
    each final A=A,
    final gas=glaSys.gas,
    each final til=til,
    each linearize=linearize,
    each final homotopyInitialization=homotopyInitialization) if have_GasLay
    "Window gas layer"
    annotation (Placement(transformation(extent={{20,10},{40,30}})));
```

mwetter:
This is fixed on the `master` and `maint_8.1.x` branches

PR #3032, opened 2022-05-23
updated existing BESTEST cases and added new cases

### LINKED ISSUE #3005: Update BESTEST cases based on the changes in ASHRAE 140 version 2020
There are following changes from version 2017 to version 2020:
- in case 600FF
  - the infiltration rate should be adjusted for the altitude (1650 m), change the rate from 0.5 ACH to 0.414 ACH. 
  - exterior surface texture should change from `medium` to `rough`
  - the `Buildings.ThermalZones.Detailed.Validation.BESTEST.Data.Win600`, some windows properties have changed:
    - thickness changed from 3.175mm to 3.048mm
    - the air gap changed from 13 mm to 12 mm
    - the thermal conductivity changed from 1.06 to 1 W/m.K
    - the solar transmittance changed from 0.86156 to 0.834
    - the solar reflectance changed from 0.0434 to 0.075
    - the Infrared absorptivity changed from 0.9 to 0.84
    - the u-value of the frame changed from 3 to 2.1
- in case 640:
  - From 8:00 hours to 23:00 hours, same
  - from 23:00 hours to 7:00 hours, same,
  - but from 7:00 hours to 8:00 hours, different: the thermostat set point shall vary linearly with time from 10°C to 20°C.
- in case 650, 650FF, 950 and 950FF:
  - the ventilation fan capacity is 1700 (not 1703.16) m3/h, and then adjusted for the altitude (1650 m). The value becomes  1409 m3/h
- new cases since version 2020:
  - case 660: same as case 600, but with low-e window
  - case 670: same as case 600,  but with single pane window
  - case 680: same as case 600, with increased wall and roof insulation
  - case 685: same as case 600, but with "20,20" Tstat (Heat=ON if Temp.>20 degC, Cool=ON if Temp.>20 degC)
  - case 695: same as case 680, but with increased wall and roof insulation with "20,20" test
  - case 680FF: same as case 680 but without mechanical heating and cooling
  - case 980: same as case 900, but with increased wall and roof insulation
  - case 985: same as case 900, but with "20,20" Tstat
  - case 995: same as 980, but with increased wall and roof insulation with "20,20" test
  - case 980FF: same as case 980, but without mechanical heating and cooling

JayHuLBL:
This closes #3005 and #3026.
- [x] updated existing BESTEST cases according to the ASHRAE 140-2020
- [x] added new cases
- [x] fixed the OCT error caused by implementation of heat transfer of windows if there is single pane glass, see #3026.


mwetter (changes requested):
- [ ] The bar plots of the free floating case look wrong. They plot case 960, which has very different results. Shouldn't the plot be for 960FF?
- [ ] Can you make a PR for `CenterOfGlass.mo` against the `maint_8.1.x` branch
- [x] Update the reference results that fail the CI tests.

JayHuLBL:
@mwetter 
- [x] The bar plots for the free floating case indeed have wrong results. Case 960 has two zones, back zone and sun space zone. The back zone is conditioned with HVAC while the sun space is free floating. In previous plot, the back zone temperature was plot but we in fact should plot the temperature in the sun space. The error has been corrected.
- [x] Made PR for the changes in `CenterOfGlass.mo` against branch `maint_8.1.x`
- [x] Updated the reference results for the classes that uses the `Case600` as the base class. The unit tests have been done with the Dymola version 2022x. (Note that the results are different when run with version 2022.).

mwetter (approved):
@JayHuLBL : Thanks for the validation, I will merge when the tests pass.

PR #3038, opened 2022-05-24
Issue3026 single pane

### LINKED ISSUE #3026: OCT error when define a single pane window

JayHuLBL:
This closes #3026.

@mwetter I am not sure how to add the release note. Should it be in the version 8.1.2?


JayHuLBL:
Close this PR as the new change should base on the maint.8.1.x. Will make another PR soon.

PR #3049, opened 2022-05-26
Issue3027 wet coil sensible heat

### LINKED ISSUE #3027: The wet coil generates an unrealistic sensible heat exchange rate
In <code>Buildings.Fluid.HeatExchangers.WetCoilCounterFlow</code>, the sensible heat exchange rate is calculated by:
```
  Modelica.Units.SI.HeatFlowRate QSen2_flow=Q2_flow - QLat2_flow
    "Sensible heat input into air stream (negative if air is cooled)";
```
If <code>Q2_flow</code> is lower than <code>QLat2_flow</code>, then <code>QSen2_flow</code> is positive even <code>Q2_flow</code> and<code>QLat2_flow</code> are negative.
In this case, the coil is heating the air.

This error can be triggered when the water flow rate is lower while the air flow rate is still high (e.g., the air-side economizer is on). In this case, the <code>Q2_flow</code> is lower but <code>QLat2_flow</code> may be higher since it is not directly related to the water side condition, as shown below:

image

mwetter:
This closes #3027

@dhblum : Can you please look at the new formulation of Buildings/Fluid/HeatExchangers/BaseClasses/HexElementLatent.mo

@SenHuang19: FYI, this should resolve the issue you reported. Can you test on your larger example? As a hack, all you would need to do is delete the instance `heaConVapCoi` in `HexElementLatent`.

We plan to include it in the 9.0.0 release that is scheduled for this Tuesday if this solves the issue.


SenHuang19:
Thanks for your information.
It seems to alleviate the issue a lot but doesn't eliminate the issue.

Before:
image


After:
image

dhblum (approved):
I think removing ``heaConVapCoi`` is the right thing to do.  (The following insight is derived from Heat Transfer by A.F. Mills Chapter 7).

An energy balance for an element of condensate has:
(+) Enthalpy entering from condensate drip from above elements due to gravity.
(+) Enthalpy entering from condensation from air.
(-) Enthalpy leaving from condensate drip to below elements due to gravity.
(-) Conduction through the thickness of the condensate element due to the difference between T_sat and T_coil.  This last part hinders condensation some since it effectively adds a thermal resistance between the coil surface and the bulk moist air.   

Thus, it is not correct to have the heat of condensation sensibly added to the coil frame explicitly.  I may suggest adding a line in the documentation about how the model does not explicitly consider the film thickness.  In reality, this thickness is not negligible, as there's research on hydrophobic coatings to quickly remove condensate layers to improve overall heat transfer coefficients, though in practice its hard to keep these coatings clean and intact.  But, this coil model's overall heat transfer coefficient is parameterized with a nominal UA anyway, which I think would include whatever heat transfer coefficient a user wants to define at nominal conditions.

mwetter:
@dhblum : Thanks, I will merge it for the 9.0.0 release, and also update on the `maint_8.1.x` branch.

PR #3069, opened 2022-06-27
Added version number

asrjy:
Added version number  to match the folder structure in the download package


mwetter:
@asrjy : Thanks for the pull request. We prefer to leave it the old way as the " 9.0.0" is only added in some distributions.

PR #3071, opened 2022-07-03
Package with standard hydronic configurations

### LINKED ISSUE #1884: Package for standard hydronic configurations
Note: The feature branch is `issue1884_hydronics`.

This is to develop a package where we collect different hydronic configurations for heating or cooling systems. 

- For these it would also be nice to have CDL control sequence. 
- The configurations either have a differential pressure or not at their supply, and have either no valve, a two-way or three-way valve, optionally with a bypass if the primary supply temperature is higher than the secondary heating design supply temperature (or cooler for cooling systems). 
- These are the repetitive elements that can then be used in a variety of projects. As such, this is a shared task between OBC, URBANopt and Spawn.

See figure for some examples of such standard configurations.

Screen Shot 2020-04-07 at 5 11 27 PM

AntoineGautier:
This closes #1884. 
The developments are within `Buildings.Fluid.HydronicConfigurations`.

TODO before merging:

- [x] Restore full CI testing

Note:

- Default parameters:
  
  - `energyDynamics=Modelica.Fluid.Types.Dynamics.FixedInitial` as opposed to the default setting `DynamicFreeInitial` in MBL because the latter triggers the warning `The initial conditions are not fully specified. Dymola has selected default initial conditions.`. To be discussed.
  - `fraK=1.0` as opposed to the default setting `fraK=0.7` for 3-way valves in MBL, see the justification at `Buildings.Fluid.HydronicConfigurations.UsersGuide.ControlValves`.

- The logic to enable/disable `dp(1|2)_nominal` is not interpreted properly by Dymola at UI runtime. I will open a FE ticket at DS: to be referenced in this page. The logic is properly handled by Modelon Impact.

  - [EDIT 8/4] Ticket #SR00922000-01 submitted to DS support.

- The logic to disregard unassigned parameters that are not enabled is not implemented in Modelon Impact that generates a warning. I will open a FE ticket at Modelon: to be referenced in this page. The logic is implemented in Dymola that throws a clear info message at compile time but no warning.

- Some models fail to translate or simulate with OCT, see `Buildings/Resources/Scripts/BuildingsPy/conf.yml`. I will open tickets at Modelon: to be referenced in this page.

- All models fail to simulate with OpenModelica and are excluded from CI testing. Error message:
  ```
  Matrix singular!
  under-determined linear system not solvable!
  division by zero at time 0, (a=-10714.99753345652) / (b=0), where divisor b expression is: 1.0 - $cse79
  simulation terminated by an assertion at initialization
  ```
  OM debugger shows this equation in `WetCoilWetRegime` as faulty:
  ```
  hSatSurEff = Buildings.Utilities.Math.Functions.smoothMax(
    x1 = hSatSurEffMinM.hSat,
    x2 = hAirIn  +(hAirOut - hAirIn) / (1 - exp(-NTUAirSta)),
    deltaX = delta*1E4); 
  ```


dhblum:
Hey @AntoineGautier thanks so much for the PR and sorry for my delayed review.  I've started to go through it, although I'm simultaneously also still learning about the applications/issues with each of these configurations, and want to still look through it in some more detail to hopefully provide better feedback.  But overall, the structures and implementations seem reasonable.  Just a couple of questions/comments to start:

1. Internal valve control options.  I'm not sure if I'm confused or if there are indeed some implementation errors.  But from the documentation, it seems ``Decoupling`` should not have ``typCtl`` and related parameters available to users to modify.  The documentation indicates the valve is to be controlled to maintain a constant bypass mass flow.   Similarly, ``SingleMixing`` documentation table says no built in controls available, but the ``typCtl`` is offered to users to change and the model block diagram indeed shows at least the possibility of feedback control based on consumer supply temperature.

2. Pump type options.  ``Decoupling``, ``InjectionThreeWay``, ``TwoWay``, ``TwoWayCheckValve``, and ``SingleMixing`` seem like they should require a pump to be present.  Although, the parameter ``typPum`` allows for a user to select ``No pump``, which I'm not sure should be possible.  Along those lines too, the pumps in those models are not conditionally instantiated based on anything as far as I can tell, so I'm not clear on what actually changes if a user selects ``No pump``.

AntoineGautier:
@dhblum 

### 1. Internal valve control options

Model `Decoupling`: 

- I wanted to support the use case where an external control signal is used in conjunction with a modulating valve, as illustrated in `Buildings.Fluid.HydronicConfigurations.ActiveNetworks.Examples.DecouplingTemperature` which represents a configuration that I encountered in some projects and that is also documented in Petitjean, 1994. To support that, the parameter `typCtl` is needed to disable the built-in controls. However, it is used a little differently than in other models. I added a clarification in the model documentation in my last commit:
  >  The specific built-in control option implemented in this model does not depend on the actual function of the consumer circuit (such as cooling, heating, or change-over). Therefore, the model remains the same whatever the value assigned to the parameter typCtl except if None (no built-in controls) is selected. In that latter case only, no built-in controls are included and the user must connect a control signal to modulate the valve. 
- In the default model configuration, a self-acting Δp control valve is used. In that case the parameter `typVar` should indeed be disabled, which is not the case due to a bug in Dymola (see below). I chose to leave the other control parameter `k` enabled in order for a user to modify the proportional band the match the self-acting Δp control valve specification. And you are right `controllerType` should be made final, which is tha case in my last commit.
- Bug in Dymola (DS ticket #SR00922000-01): in the attached model `TestEnable.mo` the input field for `p` remains enabled in the parameter dialog despite the `enable` attribute being evaluated to `false` at compile time. This seems to be due to the declaration of the Boolean parameter `flag_enable` inside the protected section. As a temporary workaround I propose to add those declarations inside the public section of the interface class and systematically assign final values to those parameters in the derived models.

  [TestEnable.mo](https://github.com/lbl-srg/modelica-buildings/files/9260579/TestEnable.mo.txt)

Model `SingleMixing`:

- The documentation is incorrect. It was edited in the last commit to mention the control of the supply temperature as a possible option.

### 2. Pump type options

See `Buildings.Fluid.HydronicConfigurations.UsersGuide.ModelParameters` section "Distribution pump":

> For configurations with a secondary distribution pump, the user may chose whether to include the pump in the configuration model or not, using the parameter typPum from the enumeration Buildings.Fluid.HydronicConfigurations.Types.Pump.

I think that this level of flexibility could be useful to support connecting a model that already includes a secondary distribution system.

The implementation of that option relies on the container (or wrapper) class `Buildings.Fluid.HydronicConfigurations.Components.Pump`. See the documentation of that model, or the documentation of the containing package `Buildings.Fluid.HydronicConfigurations.Components` for further description.
(Or more conceptually in this paper: https://www.researchgate.net/publication/225021336_Towards_Improved_Class_Parameterization_and_Class_Generation_in_Modelica)

mwetter:
@AntoineGautier @dhblum : Is there any reason for not completing this PR? There was no activity since Aug 2022 and I think these additions would be valuable to have on the master. Note that the OpenModelica issue `(a=-10714.99753345652) / (b=0)` is known and the developers work on it, so file with this problem can be put on the exclude list.

dhblum:
The delay has been my own in completing my review after Antoine addressed initial comments.  I don't think there are any major modifications required though, and will complete as soon as possible.

dhblum (changes requested):
This is looking pretty good, I made just a few edits and also merged the latest master.  There are some remaining things to address however:

- [x] ``k`` parameters for the constant source blocks in ``Buildings.Fluid.HydronicConfigurations.Components.Pump`` needed to be specified with decimal point rather then as integer in order to avoid the following translation error in Dymola:
```
Argument number 2 in smooth(0, (if con.pum.swi.u2 then 1 else 0))
is of type Integer, but it must be continuous and only containing Real.
```
I believe this has to do with the update made in ``Buildings.Controls.OBC.CDL.Continuous.Switch`` in the latest master.  See https://github.com/lbl-srg/modelica-buildings/pull/3071/commits/d5d7ca71b3d386eef5a36a571c9a1fabc285692d.
*Comment by mwetter*: Yes the constants need to be declared with a decimal point. This will be fixed in future versions of Dymola, but since Dymola 2023x requires it, we need to adjust our models.


- [x] Need to increment the integers assigned for ``Buildings.Fluid.HydronicConfigurations.Controls.OperatingModes`` by 1 in order to avoid the warning generated by the block ``Buildings.Controls.OBC.CDL.Routing.RealExtractor`` after the update in https://github.com/lbl-srg/modelica-buildings/issues/3125 to handle index inputs that are out of range and require them to be > 0.  The warning is for example:
```
Warning: The following was detected at time: 0
  In DecouplingTemperature.loa.loa.ctl.extIndSig: The extract index is out of the range.
  Failed condition: loa.mode > 0 and loa.mode <= 2
```


- [x] Fix initialization error when simulating ``Buildings.Fluid.HydronicConfigurations.ActiveNetworks.Examples.DiversionOpenLoop`` and ``Buildings.Fluid.HydronicConfigurations.ActiveNetworks.Examples.DiversionOpenLoop`` related to the new mover model using the Euler method implemented in https://github.com/lbl-srg/modelica-buildings/issues/2668.  The error is:
```
Error: The following error was detected at time: 0

Model error - division by zero: ((pressure.V_flow[nPre]-pressure.V_flow[nPre-1])*pressure.dp[nPre]) / (pressure.dp[nPre]-pressure.dp[nPre-1]) = (0) / (0)

The stack of functions is:
Buildings.Fluid.Movers.BaseClasses.Euler.getPeak
Buildings.Fluid.Movers.BaseClasses.Euler.getPeak(
con.perPum.pressure, 
con.perPum.power)

FixInitials:Init: Integrator failed to start model.
```

- [ ] @AntoineGautier do you have scripts to regenerate the plot figures in ``Buildings.Fluid.HydronicConfigurations.ActiveNetworks.Examples.DiversionOpenLoop`` and ``Buildings.Fluid.HydronicConfigurations.ActiveNetworks.Examples.ThrottleOpenLoop``?

- [x] Update reference results as needed once issues are fixed

AntoineGautier:
@dhblum Thanks for the review. I committed the fixes that were required after the changes made on the master (to Movers and RealExtractor) since the PR was first open. The only noticeable impact is described in the message of commit [e9393df](https://github.com/lbl-srg/modelica-buildings/pull/3071/commits/e9393df4a3e89413a8a2e709a3844777205adf86). 
Yes, the scripts to generate the plots used in the documentation of the example models are included, see:
```
./Buildings/Resources/Images/Fluid/HydronicConfigurations/ActiveNetworks/Examples/DiversionOpenLoop.py
./Buildings/Resources/Images/Fluid/HydronicConfigurations/ActiveNetworks/Examples/ThrottleOpenLoop.py
```

dhblum (approved):
Thank you very much @AntoineGautier for addressing the comments and making the changes.  Looks like unit tests passed.  I just also added the new package to the release notes.  I think this is ready to merge.  Perhaps @mwetter wants a final look.

mwetter (approved):

PR #3085, opened 2022-08-08
Handle spaces in paths when installing Spawn

sjoelund:
Note: `shell=True` could probably be changed to `False` since no shell interpretation should be needed anymore. I created a minimal fix instead.

```
modelica-buildings$ cp -a Buildings/ Buildings\ blabla
modelica-buildings$ Buildings\ blabla/Resources/src/ThermalZones/install.py
/bin/sh: 1: /home/marsj/modelica-buildings/Buildings: Permission denied
```

Should probably also be fixed for 9.x maintenance

### DIFF
--- a/Buildings/Resources/src/ThermalZones/install.py
+++ b/Buildings/Resources/src/ThermalZones/install.py
@@ -21,9 +21,9 @@ if __name__ == '__main__':
   # Parse the arguments
   args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


mwetter:
@sjoelund Thanks for the change. There is not yet a 9.x maintenance branch, we will create it when the 9.1 development (on the master) would start to differ from what will be on a 9.x maintenance branch.

PR #3086, opened 2022-08-08
Changed subprocess to use shell=False

mwetter:


modelonrobinandersson (commented):
Hi @mwetter, I have been trying to use this script and also ran into issues on Windows due to `shell=True`. However, I noticed that changing this causes another error   
`OSError: [WinError 193] %1 is not a valid Win32 application`  
but this I resolved by adding the name of the python binary to the list `proc`, i.e.:  
```  
---proc=[os.path.join(os.path.dirname(os.path.realpath(__file__)), "EnergyPlus_9_6_0", "install.py")]  
+++proc=["python", os.path.join(os.path.dirname(os.path.realpath(__file__)), "EnergyPlus_9_6_0", "install.py")] 
```  

Is it possible to add this change as well to this PR and possibly merged to master soon?

mwetter:
@modelonrobinandersson : I anticipate that this will be on the master by tomorrow, with your suggested change included.

PR #3090, opened 2022-08-12
Created duty cycle (Logical.VariablePulse)

### LINKED ISSUE #3042: Add duty cycle blocks in CDL.Logical
This issue is to add duty cycle blocks as discussed in ASHRAE231p meeting:
- [x] `CDL.Logical.ConstantDutyCycle`: produces an output that cycles on and off according to the specified length of time for the cycle, and the specified percentage of the time that the output should be on. It only cycles the output when the go input is on; if the go input is off, the output remains on.
- [x] `CDL.Logical.VariableDutyCycle`: produces an output that cycles on and off according to the specified length of time for the cycle and the value of the input, which indicates the percentage of the cycle time the output should be on. It only cycles the output when the input value is greater than zero. If the input is zero, the output ~~remains off~~ remains on.

For the `VariableDutyCycle`, if for example the cycle time is 10 minutes and the cycle on percentage is 50%, then at first 5 minutes the output will be `True` and the next 5 minutes will be `False`. If in the middle of one cycle (say 10 minutes cycle time), the cycle percentage changes from 50% to 25%, the output will become `True` immediately (a new timing period is established) for 2.5 minutes and remain `False` for 7.5 minutes.

JayHuLBL:
This closes #3042.


Buildings/Controls/OBC/CDL/Logical/ConstantDutyCycle.mo
```diff
@@ -0,0 +1,176 @@
...
+        Text(
+          extent={{-100,140},{100,100}},
+          textColor={0,0,255},
+          textString="%name")}),
+    Documentation(
+      info="<html>
+<p>
+Block that produces an output that cycles on and off according to the specified
+length of time for the cycle, and the specified percentage of the time that the
+output should be on. It cycles the output when the <code>go</code> input is on;
```
mwetter, line 160: Use `true` or `false`, not "on" as there is no such data type.
Also, it is odd to put the input to `false` and then this means the output is `true`. Is this the intent?

Buildings/Controls/OBC/CDL/Logical/VariableDutyCycle.mo
```diff
@@ -0,0 +1,270 @@
...
+  Buildings.Controls.OBC.CDL.Continuous.Abs abs1
+    "Find the absolute value change"
+    annotation (Placement(transformation(extent={{-10,50},{10,70}})));
+  Buildings.Controls.OBC.CDL.Continuous.GreaterThreshold greThr
+    "Check if there is input value change"
+    annotation (Placement(transformation(extent={{20,50},{40,70}})));
+  Buildings.Controls.OBC.CDL.Continuous.Switch swi
+    "If there is input value change, use the new value"
+    annotation (Placement(transformation(extent={{60,20},{80,40}})));
+  Buildings.Controls.OBC.CDL.Continuous.GreaterThreshold greThr1
```
mwetter, line 43: This needs a hysteresis, else if uCycOn is the output of a PI controller, noise will trigger it.

Buildings/Controls/OBC/CDL/Logical/VariableDutyCycle.mo
```diff
@@ -0,0 +1,270 @@
...
+  Buildings.Controls.OBC.CDL.Logical.Sources.Constant swi2(
+    final k=true) "Remain true output"
+    annotation (Placement(transformation(extent={{0,-80},{20,-60}})));
+protected
+  Cycle cycOut(final period=period) "Produce cycling output"
+    annotation (Placement(transformation(extent={{0,-20},{20,0}})));
+
+  block Cycle
+  "Produce output that cycles on and off according to the specified pulse period and width input"
+    parameter Real period
```
mwetter, line 58: Specify unit and min value

Buildings/Controls/OBC/CDL/Logical/VariableDutyCycle.mo
```diff
@@ -0,0 +1,270 @@
...
+    Interfaces.BooleanInput go
+      "True: cycle the output"
+      annotation (Placement(transformation(extent={{-140,-100},{-100,-60}}),
+          iconTransformation(extent={{-140,-100},{-100,-60}})));
+    Interfaces.RealInput uCycOn
+      "Ratio of the cycle time that the output should be on"
+      annotation (Placement(transformation(extent={{-140,-20},{-100,20}}),
+          iconTransformation(extent={{-140,-20},{-100,20}})));
+    Interfaces.BooleanOutput y
+      "Connector of Boolean output signal"
```
mwetter, line 69: The comment should be more descriptive.

Buildings/Controls/OBC/CDL/Logical/VariableDutyCycle.mo
```diff
@@ -0,0 +1,270 @@
...
+    final unit="1")
+    "Ratio of the cycle time that the output should be on"
+    annotation (Placement(transformation(extent={{-140,-20},{-100,20}}),
+        iconTransformation(extent={{-140,-20},{-100,20}})));
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanOutput y
+    "Connector of Boolean output signal"
+    annotation (Placement(transformation(extent={{100,-60},{140,-20}}),
+        iconTransformation(extent={{100,-20},{140,20}})));
+
+  Buildings.Controls.OBC.CDL.Discrete.Sampler sam(
```
mwetter, line 27: A lot of these blocks below should be protected, no?

Buildings/Controls/OBC/CDL/Logical/VariableDutyCycle.mo
```diff
@@ -0,0 +1,270 @@
...
+          textColor={0,0,255},
+          textString="%name")}),
+    Documentation(
+      info="<html>
+<p>
+Block that produces an output that cycles on and off according to the specified
+length of time for the cycle and the value of the input <code>uCycOn</code>,
+which indicates the percentage of the cycle time the output should be on.
+It only cycles the output when the input value is greater than zero.
+If the input is zero, the output remains on.
```
mwetter, line 255: So, a duty cycle of 1 means it is off, and vice versa? This seems wrong.

Buildings/Controls/OBC/CDL/Logical/Validation/ConstantDutyCycle.mo
```diff
@@ -0,0 +1,92 @@
+within Buildings.Controls.OBC.CDL.Logical.Validation;
+model ConstantDutyCycle
```
mwetter, line 2: It would be more user friendly to call these `DutyCycleVariable` and `DutyCycleConstant` as then, they will be next to each other in a menu that allows selecting blocks.

mwetter (changes requested):
See inline comments for a brief initial review.

JayHuLBL:
@mwetter It will be ready for your review when CI tests are passed.

Buildings/Controls/OBC/CDL/Logical/Validation/VariablePulseNegativeStartTime.mo
```diff
@@ -0,0 +1,117 @@
...
+  __Dymola_Commands(
+      file="modelica://Buildings/Resources/Scripts/Dymola/Controls/OBC/CDL/Logical/Validation/VariablePulseNegativeStartTime.mos" "Simulate and plot"),
+  Documentation(info="<html>
+<p>
+Validation test for the block
+<a href=\"modelica://Buildings.Controls.OBC.CDL.Logical.VariablePulse\">
+Buildings.Controls.OBC.CDL.Logical.VariablePulse</a>.
+</p>
+<p>
+This validates the blocks with a start time of <i>-3600</i>.
```
mwetter, line 93: Add units, and better write <i>t=-3600</i> seconds.

Buildings/Controls/OBC/CDL/Logical/Validation/VariablePulsePositiveStartTime.mo
```diff
@@ -0,0 +1,117 @@
...
+    final samplePeriod=120,
+    final period=900)
+    "Variable pulse with width input changing from zero to positive value"
+    annotation (Placement(transformation(extent={{60,-70},{80,-50}})));
+  Buildings.Controls.OBC.CDL.Continuous.Sources.Pulse conPul2(
+    final amplitude=0.5,
+    final width=0.5,
+    final period=3600,
+    final shift=300,
+    final offset=0.3)    "Continuous pulse signal"
```
mwetter, line 65: Use one empty space or better put comment on separate line as above

Buildings/Controls/OBC/CDL/Logical/Validation/VariablePulsePositiveStartTime.mo
```diff
@@ -0,0 +1,117 @@
...
+  __Dymola_Commands(
+      file="modelica://Buildings/Resources/Scripts/Dymola/Controls/OBC/CDL/Logical/Validation/VariablePulsePositiveStartTime.mos" "Simulate and plot"),
+  Documentation(info="<html>
+<p>
+Validation test for the block
+<a href=\"modelica://Buildings.Controls.OBC.CDL.Logical.VariablePulse\">
+Buildings.Controls.OBC.CDL.Logical.VariablePulse</a>.
+</p>
+<p>
+This validates the blocks with a start time of <i>4800</i>.
```
mwetter, line 93: Add unit of time.

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
```diff
@@ -0,0 +1,329 @@
...
+  "Generate boolean pulse with the width specified by input"
+
+  parameter Real samplePeriod(unit="s")
+    "Sample period of the width input";
+  parameter Real period(unit="s")
+    "Time for one pulse period";
+  parameter Real chaWidThr=0.05
+    "Threshold for checking if the width input is changed. It is the ratio of the value change to the original value"
+    annotation (Dialog(tab="Advanced"));
+  parameter Real zerWidThr=0.01
```
mwetter, line 12: Let's discuss. I think what this means and what you want to accomplish is something like "Minimum value for `u` below which the output remains always `false`.

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
```diff
@@ -0,0 +1,329 @@
+within Buildings.Controls.OBC.CDL.Logical;
+block VariablePulse
+  "Generate boolean pulse with the width specified by input"
+
+  parameter Real samplePeriod(unit="s")
+    "Sample period of the width input";
+  parameter Real period(unit="s")
+    "Time for one pulse period";
+  parameter Real chaWidThr=0.05
```
mwetter, line 9: Maybe this should be "Minimum required change in `u` to re-trigger an update of the output interval`

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
```diff
@@ -0,0 +1,329 @@
...
+    "Ratio of the period that the output should be true"
+    annotation (Placement(transformation(extent={{-180,-20},{-140,20}}),
+        iconTransformation(extent={{-140,-20},{-100,20}})));
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanOutput y
+    "Boolean pulse when the input is greater than zero"
+    annotation (Placement(transformation(extent={{140,-130},{180,-90}}),
+        iconTransformation(extent={{100,-20},{140,20}})));
+
+protected
+  Buildings.Controls.OBC.CDL.Discrete.Sampler sam(
```
mwetter, line 28: I don't think this should be time sampler. Shouldn't it rather be triggered whenever the downstream `greThr` triggers?

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
```diff
@@ -0,0 +1,329 @@
...
+    "Sample period of the width input";
+  parameter Real period(unit="s")
+    "Time for one pulse period";
+  parameter Real chaWidThr=0.05
+    "Threshold for checking if the width input is changed. It is the ratio of the value change to the original value"
+    annotation (Dialog(tab="Advanced"));
+  parameter Real zerWidThr=0.01
+    "Threshold for checking if the width input is greater than zero"
+    annotation (Dialog(tab="Advanced"));
+  Buildings.Controls.OBC.CDL.Interfaces.RealInput uWid(
```
mwetter, line 15: Better just call this `u` if you agree it can't be anything else then the width.

mwetter (changes requested):
Please see inline comments. Let's discuss briefly about the time sampler as I don't think it does the right thing. Imagine a `u` being connected to a ramp with duration 120 s, and the sampling period is 120 s. I think this will then only trigger an update in `y` once.

Buildings/Controls/OBC/CDL/Logical/Validation/VariablePulseNegativeStartTime.mo
JayHuLBL: updated.

Buildings/Controls/OBC/CDL/Logical/Validation/VariablePulsePositiveStartTime.mo
JayHuLBL: updated.

Buildings/Controls/OBC/CDL/Logical/Validation/VariablePulsePositiveStartTime.mo
JayHuLBL: added.

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
JayHuLBL: changed to `Minimum value of input below which the output remains always false`.

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
JayHuLBL: changed to `u`.

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
JayHuLBL: updated.

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
JayHuLBL: Thanks for the suggestion. I replaced the sampler with the triggered one, with the `pre` to break to loop.

JayHuLBL:
@mwetter I have addressed your comments. It will be ready for your review when the CI tests passed.

JayHuLBL:
@mwetter it's ready for you review.

JayHuLBL:

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
```diff
@@ -0,0 +1,340 @@
+within Buildings.Controls.OBC.CDL.Logical;
+block VariablePulse
+  "Generate boolean pulse with the width specified by input"
+
+  parameter Real period(unit="s")
+    "Time for one pulse period";
+  parameter Real chaWidThr=0.01
+    "Minimum input change to re-trigger an update of the output interval. It is the ratio of the value change to the original value"
```
mwetter, line 8: Do we really want to make a relative comparison? Wouldn't it be simpler to just check for absolute changes? In that case, the test for smaller than a tolerance can also be removed and the limit can be imposed on the width of the pulse. This will avoid state events caused by `greThr1`

Buildings/Controls/OBC/CDL/Logical/Validation/VariablePulse.mo
mwetter, line 1: Please remove the `final` in the validation tests; otherwise a user cannot change them. There is no need to make them final

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
```diff
@@ -0,0 +1,340 @@
...
+    annotation (Dialog(tab="Advanced"));
+  parameter Real minTruFalHol(
+    final quantity="Time",
+    final unit="s",
+    final min=Constants.small)=1
+    "Minimum time to hold true or false";
+  parameter Real zerWidThr=0.01
```
mwetter, line 15: Why not using `minTruFalHol` to set the bound on `u` below which it is considered always off, or always on. E.g., if `u < uMin` with `uMin = minTruFalHol/period` then `y=false` always, wheree `minTruFalHod` cannot be longer than `period/2`.

mwetter (commented):
Please see inline comments.

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
JayHuLBL: Thanks! I removed the relative comparison and avoided using `greThr1`.

Buildings/Controls/OBC/CDL/Logical/Validation/VariablePulse.mo
JayHuLBL: Removed the `final` in the validation tests.

Buildings/Controls/OBC/CDL/Logical/VariablePulse.mo
JayHuLBL: changed to use the `uMin` and set it as suggested. It is then added to the protected section.

mwetter (approved):

PR #3100, opened 2022-08-26
Issue3095 battery discharge

### LINKED ISSUE #3095: Potential bug in battery model
In class `Buildings.Electrical.DC.Storage.BaseClasses.Charge`, the actual power flowing through the battery `Pact` is computed as
```
// Charge balance of battery
  PAct = if P > 0 then etaCha*P else (2-etaDis)*P;
```
where the `etaCha` is charging efficiency, the `eatDis` is discharging efficiency and the `P` is the power that should be provide or taken from the battery.

The power taken from the battery during discharge may should be computed as `(1/etaDis)*P`, rather than `(2-etaDis)*P`. The two implements have difference results. However, when the discharge efficiency is higher, for example 0.9 showing in below plot, the difference is small.
Battery2

But when the efficiency is small, for example 0.1 in below plot, the difference becomes significant.
Battery

This issue is to check if it is indeed a bug.

- [ ] check if current implementation has reference to support it
- [ ] if not, reimplement the calculation.

JayHuLBL:
This closes #3095.


JayHuLBL:
@mwetter This is ready for your review.

mwetter (approved):

PR #3113, opened 2022-09-19
Extended functionality to install.py when installing required binaries

modelonrobinandersson:
Hi!

This pull request resolves two issues:

1. Possibility to invoke the script `Buildings/Resources/src/ThermalZones/install.py` using `Python 3.6`. This is not possible today due to the use of f-strings. The reason why I would like this to be supported is because when you install Python 3 on CentOS 7.4 or Ubuntu 18.04, you will get version `Python 3.6.`. Moreover, end of life for CentOS 7 is not until 2024.
2. Using CentOS 7.4 in particular, you will likely have the executable `/usr/bin/python` to point at `Python 2.7`, while `/usr/bin/python3` points at `Python 3.6`. With the changes in this PR I check both `python` and `python3` to see if any of them correspond to `Python 3.x`, in order to invoke the `subprocess.run(...)` using the correct executable. The previous behavior would always invoke `subprocess.run(...)` using the executable `python`, which could be `Python 2.7`. I have also added some error checking in order to make it easier for a user to understand what is going wrong (in case something goes wrong). This is optional but I think it would be nice to have in case someone else will end up debugging this.

I have tested my changes on CentOS 7.4 using `Python 3.6.8` and Ubuntu 22.04 using `Python 3.10.4`.


modelonrobinandersson:
Is there any way I can see the build error? Clicking at the links above I only noticed some errors with "docker login" but cannot see how that is related to my changes.

mwetter:
@modelonrobinandersson : Thanks for the PR. I will merge it through https://github.com/lbl-srg/modelica-buildings/pull/3115 once it goes through the CI testing.
The "docker login" fails as access is disabled for remote pull requests.

modelonrobinandersson:
> @modelonrobinandersson : Thanks for the PR. I will merge it through #3115 once it goes through the CI testing. The "docker login" fails as access is disabled for remote pull requests.

I see, thank you!

mwetter:
@modelonrobinandersson : Thanks, I added the change to the PR.

PR #3093, opened 2022-08-22
Issue2912 district cooling

JayHuLBL:
This closes #2912 and #2308.

PR #3116, opened 2022-09-21
Issue2912 district cooling - Partial Review Comments for PR #3093 

khinkelman:
Hi @JayHuLBL. These commits address the inline comments in PR #3093 and solve some of the OM errors. There is one outstanding error that may be an open OM issue (`division by zero at time 0, (a=0) / (b=0), where divisor b expression is: 0.0`, see for reference PR #3002). I am still debugging my OCT installation so am not yet able to evaluate those errors. However in the meantime, if you could please pull these changes (assuming you approve), that would be great; it is possible some of the fixes I made herein will resolve the OCT errors as well. 

In addition to addressing comments, these changes include:
- Minor formatting fixes
- Cleaning and fixing bugs in the simulation scripts (e.g., deleted repeated variables and unneeded lines).
- Propagating dynamics and initialization parameters in the cooling plant model `DHC.Plants.Cooling.ElectricChillerParallel`. In addition to solving one of the OM errors, these changes allow users to apply filters to CHW pumps, CW pumps, and the chillers, which can improve the numerical stability.  

For reference, the division by zero error occurs in both the plant model (`DHC.Plants.Cooling.Examples.ElectricChillerParallel`) and the complete DC example (`DHC.Examples.Cooling.DirectUncontrolledETS`). The error points to psychometric functions for calculating wetbulb properties, including `TWetBul_TDryBulPhi`, `XiSat`, and `pSat(TWetBul)` (all within `Buildings.Utilities.Psychrometrics`). 

If you have any questions/comments, please let me know. Thanks again for your help!

@hoxj180: I am copying you here as well so that you can track the progress.


JayHuLBL:
@khinkelman With the change, we can run the model with Optimica. However, it still cannot run with the OpenModelica with the same `division by zero` error. I also noticed that there are results change. Is it intentional? See below:
Screen Shot 2022-09-21 at 3 32 25 PM

khinkelman:
Hi @JayHuLBL thanks for the message. I double checked the simulation and yes, the revised results are physically correct. These changes were caused (primarily) by my modification of the terminal unit heat exchangers at the buildings. 

I updated the reference results with my latest commit (sorry I forgot to earlier!). If you need anything else from me to merge these with the master, please let me know. Many thanks for your help. 

Best, 
Katy

PR #3126, opened 2022-09-30
Changed temperature time constant to non zero

### LINKED ISSUE #3123: Two-port temperature sensors should set the tau to non-zero
In class `Buildings.Air.Systems.SingleZone.VAV.ChillerDXHeatingEconomizer`, the temperature sensors `senTSup` and `senTMixAir` set `tau=0`.

This issue is to change the `tau` to a non-zero value, say `tau=1`. The `non-zero` value should be used as suggested in `Buildings.Fluid.Sensors.UsersGuide`.

JayHuLBL:
This closes #3123.


mwetter (changes requested):
@JayHuLBL : For this model, `tau` can be set to zero as the model prescribes the mass flow rate and sets `allowFlowReversal=false`. Hence, the upstream temperature can be used at the sensor unless I miss something here.

JayHuLBL:
As explained, the time constant can be zero in this case. The G36 control will be improved to avoid the algebraic loop issue.

PR #3073, opened 2022-07-05
Undo removal of latent heat transfer to compute the coil state in HexElementLatent 

AntoineGautier:
This closes #3065.

Issue #3065, opened 2022-06-14
Investigate WetCoilCounterFlow energy balance


AntoineGautier:
**Describe the bug**

The commit df7415ebeb introduced some changes that break the energy balance of the model `Fluid.HeatExchangers.WetCoilCounterFlow` under the conditions simulated in `Fluid.HeatExchangers.Validation.WetCoilEffectivenessNTUCounterFlow`.
Basically the heat flow rate on the water side `*Q1_flow` remains equal to the sensible heat flow rate, leaving the latent part unbalanced, see attached plot.

![WetCoilEffectivenessNTUCounterFlow](https://user-images.githubusercontent.com/10965994/173615302-3a42aff9-768a-4a67-9cfa-103badda46f0.png)

The main change was to remove the heat connection to the mass (heaConVapCoi), see details at https://github.com/lbl-srg/modelica-buildings/issues/3027#issuecomment-1138849935.

**Expected behavior**

`*Q1_flow` should be the opposite of `*Q2_flow`.

**Version**
 - Modelica Buildings Library: a4bf2006b2

AntoineGautier:
### Additional notes

Simulating `Buildings.Fluid.HeatExchangers.Validation.WetCoilCounterFlowLowWaterFlowRate` up to 6000 s shows that, despite a zero water mass flow rate beyond 2700 s, there is a steady state 

- latent heat transfer in commit df7415ebeb (where the modification was introduced),
- latent and sensible heat transfer that cancel each other out in the previous commit dff251c2a1 (just before the modification),

So none of the two models seem correct.

Commit df7415ebeb
![HexElementLatent](https://user-images.githubusercontent.com/10965994/177191290-5606c8a8-733a-4db5-b8a9-ddaa96f57e05.png)

Commit dff251c2a1
![HexElementLatent](https://user-images.githubusercontent.com/10965994/177191465-b1633c76-a6a9-4370-b958-1eeae9f63f2a.png)

dhblum:
[...]

PR #3138, opened 2022-11-02
Undo removal of latent heat transfer to compute the coil state in HexElementLatent - Maint 8.1.x

dhblum:
This is for https://github.com/lbl-srg/modelica-buildings/issues/3065, following https://github.com/lbl-srg/modelica-buildings/pull/3073, to update the Maintenance 8.1.x branch.

@mwetter, the udpated model and documentation files pushed here so far.  Have not updated reference results yet.  Two things:

1. I made the changes needed for using for MSL 3.2.3.  I then ran unit tests with buildingspy 2.2.dev0 and Dymola 2022x, per the Dymola "About Dymola" window, and found at least one unit test that failed the reference check that I didn't expect to (though insignificant to the human eye).  This makes me wonder if I'm running the tests with the right tool versions.  Is Dymola 2022x the same as ``docker pull "$DOCKER_USERNAME"/travis_ubuntu-1804_dymola:2022-x86_64_rev-1;``defined in the travis yml?  I also ran the tests in Ubuntu 20 instead of 18.  If not, I'll need to get the installation for that version of Dymola to run the tests and generate results for this PR.

2.  ``Buildings.Fluid.HeatExchangers.Validation.WetCoilCounterFlowLowWaterFlowRate`` is a test that is in the master and maint_9.0.x but isn't in maint_8.1.x.  The files I've pushed here add this test, but let me know if this shouldn't be added in the maintenance branch.  Then, per 1. above, we can generate the appropriate reference results.


dhblum:
And depending on your answer to 2. above, I'll fix the failing test because WetCoilCounterFlowLowWaterFlowRate isn't in the package.order.

mwetter:
@dhblum Dymola 2022x is different from `travis_ubuntu-1804_dymola:2022-x86_64_rev-1` (note that travis is 2022, not 2022x). The right version is
https://github.com/lbl-srg/modelica-buildings/blob/bbdf2c6b56e3ce584cdce138be644246cc651b2c/.travis.yml#L99

Adding a new test on the maintenance branch is fine.

dhblum:
@mwetter I updated reference results for the Dymola version.  I also removed the test that was added in version 9 as it was throwing a division by zero error that might be fixed by something related to the update to version 9.  So then not sure if it belongs in a version 8 maintenance.  Although I didn't look into it any further.

mwetter (approved):

dhblum:
@mwetter Turns out my unexplained simulation fails I mentioned to you in person were due to hard drive space issues on my end.  I've rerun unit tests, checked against previous commit https://github.com/lbl-srg/modelica-buildings/commit/dbad43cdeec9fe34d86e68beec7ba49b445878b6, and updated reference results.

dhblum:
Jobs [.41](https://app.travis-ci.com/github/lbl-srg/modelica-buildings/jobs/588546192) and [.38](https://app.travis-ci.com/github/lbl-srg/modelica-buildings/jobs/588546189) were failing overnight due to stall, the only ones failing in the [Build](https://app.travis-ci.com/github/lbl-srg/modelica-buildings/builds/257818239).  I restarted them both and now they're both passing.

PR #3163, opened 2022-11-18
Issue2510 pipe autosize dh start

### LINKED ISSUE #2510: PipeAutosize causes initialization error when dh.start attribute too far from solution
The attached model fails to initialize with Dymola and either Cvode or Dassl.
```
Warning: Failed to solve nonlinear system using Newton solver.
  During initialization at time: 0
  Tag: initialization.nonlinear[2]
Error: could not solve simplified initialization for homotopy method.
Error: Integrator failed to start model.
```
Modifying the default value of the `start` attribute of the parameter `dh` from 0.2 to 0.05 (the solution being 0.034) solves the issue. It seems that only a start attribute too *high* from the solution causes an issue. The same modified default (0.05) leads to a solution for a nominal flow rate in the range of [0.3, 500] kg/s.

[BugAutosize.mo.txt](https://github.com/lbl-srg/modelica-buildings/files/6545961/BugAutosize.mo.txt)

dhblum:
This is to close #2510.  It:
- adds a new validation model to test the pipe autosizing at high and low ``dh`` values
- changes the ``start`` attribute to 0.01.

This is to merge to ``master``.  Corresponding PR to be made to ``maint9.1.x`` shortly.


dhblum:
@mwetter The MSLAIT model failed so I restarted that job.  Optimica also failed to initialize the new validation model, which Dymola and OpenModelica were able to pass.  I'm thinking to submit the problem to Modelon. What's the process for that?  Should the model be added to do not test list for Optimica for the time being as well?

PR #3206, opened 2022-12-19
Issue3189 g36 enumeration

### LINKED ISSUE #3189: improve enumerations used in G36
This issue is to improve the wordings used in the enumerations for G36 sequences, based on the feedback from ctrl-flow work.
- [x] Update wording for options in OA control
  - Single damper for ventilation and economizer with AFMS  
  - Seperate dampers for ventilation and econmomizer with AFMS  
  - Seperate dampers for ventilation and economizer with DP sensor.  
  - spell out these abbreviations
- [x] no need to list `energy code` in `ASHRAE 90.1-2016 energy code`
- [x] no need to list `ventilation code` in `ASHRAE 62.1-2016 ventilation code`
- [x] no need to list the years in `ASHRAE 90.1-2016`, `ASHRAE62.1-2016` and `Title 24-2020`

JayHuLBL:
This closes #3189.
- [x] removed `Not_Specified` option for the energy standard and ventilation standard.
- [x] removed the year suffix for the energy standard, such like changing the  `ASHRAE90_1_2016` to `ASHRAE90_1`.
- [x] removed `no_economizer` option.
- [x] improved the descriptions of the configuration options of outdoor air section.


Buildings/Controls/OBC/ASHRAE/G36/AHUs/MultiZone/VAV/Economizers/Controller.mo
```diff
@@ -6,18 +6,19 @@ block Controller
...
-  parameter Buildings.Controls.OBC.ASHRAE.G36.Types.EnergyStandard eneStd=Buildings.Controls.OBC.ASHRAE.G36.Types.EnergyStandard.Not_Specified
+  parameter Buildings.Controls.OBC.ASHRAE.G36.Types.EnergyStandard eneStd=
+    Buildings.Controls.OBC.ASHRAE.G36.Types.EnergyStandard.ASHRAE90_1
     "Energy standard, ASHRAE 90.1 or Title 24";
   parameter Buildings.Controls.OBC.ASHRAE.G36.Types.ControlEconomizer ecoHigLimCon
     "Economizer high limit control device";
-  parameter Buildings.Controls.OBC.ASHRAE.G36.Types.ASHRAEClimateZone ashCliZon=
-    Buildings.Controls.OBC.ASHRAE.G36.Types.ASHRAEClimateZone.Not_Specified
+  parameter Buildings.Controls.OBC.ASHRAE.G36.Types.ASHRAEClimateZone ashCliZon(
+    start=Buildings.Controls.OBC.ASHRAE.G36.Types.ASHRAEClimateZone.Zone_3A)
```
mwetter, line 15: What is the rationale for providing a `start` value? 
This value is not meaningful for many climate zones, hence it would be good to understand why we don't leave it unassigned or `Not_Specified`.

mwetter (commented):
@JayHuLBL : The PR looks generally good, but please look at the question regarding the `start` attribute which is now set in various models.

Buildings/Controls/OBC/ASHRAE/G36/AHUs/MultiZone/VAV/Economizers/Controller.mo
JayHuLBL: @mwetter If the parameter has no `start` attribute, is unassigned and disabled, the model will fail to run. For example the parameter `ashCliZon`, if the Title 24 standard is used, the parameter will be disabled. When it is unassigned and there is no `start` attribute, we can not run the model.

The `Not_Specified` should not be an option for the climate zone, as there is no warning in the `ctrl-flow` to avoid user to choose the option.

JayHuLBL:
@mwetter The `Not_Specified` was re-introduced as an option for the climate zone.

Issue #3139, opened 2022-11-03
G36 VAV MZ controller


AntoineGautier:
This issue is to track the changes related to 
- the comments at https://github.com/lbl-srg/modelica-buildings/issues/1913 that were disregarded:
  - [x] See comment on `uRelFan`: this input should be removed.
  - [x] See comment on `uSupFanSpe_actual`: this input should be removed.
  - [x] See comment on `u1SupFan` and `u1RelFan`: the description string should be corrected.
- some recent comments that further need to be considered:
  - [x] In `Buildings.Controls.OBC.ASHRAE.G36.Types.OutdoorAirSection` the `NoEconomizer` choice is not mapped to any control logic: it should be removed as the corresponding control sequence is not specified in G36.
  - [x] In `Buildings.Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.Controller` the option `Buildings.Controls.OBC.ASHRAE.G36.Types.BuildingPressureControlTypes.ReturnFanCalculatedAir` should be deleted as suggested by Hwakong on Fri, Oct 21, 2022. (Please note that the current implementation of this option yields an algebraic loop if the controller has a proportional term as the fan speed has direct feedback.)
  - [x] In `Buildings.Controls.OBC.ASHRAE.G36.Types.FreezeStat` for `Connected_to_BAS_NO` and `Connected_to_BAS_NC`: The choice of polarity should be removed as it is nowhere else present (for occupancy or window switches for instance). It is recommended to choose a NC polarity by default and to make it explicit in the documentation (for all other Boolean switches as well, such as occupancy or window switches).  In addition, the implementation should be validated as the description string for `u1FreSta` is incorrect (confusion between NC and NO): see comment on `u1FreSta` at https://github.com/lbl-srg/modelica-buildings/issues/1913. Eventually, Reece mentioned on Wed, Nov 2, 2022 that a freeze stat connected to the BAS only has an impact on the 3rd stage of freeze protection (5.16.12.3): the first 2 stages based on SAT should remain unchanged.
  - [x] FER: Add option to completely disable freeze protection (for climates that don't need it): this should be close to the actual implementation for `Hardwired_to_equipment` but an additional Boolean parameter is needed.
[...]

PR #3218, opened 2023-01-19
Issue3139 mz controller without changes to pressure independent damper option

JayHuLBL:
This address the issues in #3139, except the last two items about the pressure independent damper.


Buildings/Examples/VAVReheat/BaseClasses/VAVReheatBox.mo
```diff
@@ -72,7 +72,7 @@ model VAVReheatBox "Supply box of a VAV system with a hot water reheat coil"
     "Supply Air Volumetric Flow Rate"
     annotation (Placement(transformation(extent={{100,70},{120,90}}),
         iconTransformation(extent={{100,70},{120,90}})));
-  Buildings.Fluid.Actuators.Dampers.PressureIndependent vav(
+  Buildings.Fluid.Actuators.Dampers.Exponential vav(
```
AntoineGautier, line 75: If we don't address the issue with `have_preIndDam` in this PR then this modification must be undone.
Note that

- The zone temperature setpoint is no longer met in the examples with this change.
- `Buildings.Examples.VAVReheat.BaseClasses.Controls.RoomVAV` which is used for the ASHRAE2006 example does not include a pressure independent *control logic*, so `Dampers.PressureIndependent` is needed there.
- However, I noticed that on the master branch, `have_preIndDam` is left at its default value `false` for the Guideline36 example, despite the model using `Dampers.PressureIndependent`. This is a mistake that should be corrected (or that may become obsolete after refactoring to remove `have_preIndDam` completely). But again, not through this PR I believe.

AntoineGautier (changes requested):
See my inline comment.
In 72e208fbfc I also modified the `FreezeStat` enumeration as the previous description strings were not clear enough (as discussed yesterday).
- [x] Some reference results may still need to be updated (after undoing the change of the damper model used in the example models).

Buildings/Examples/VAVReheat/BaseClasses/VAVReheatBox.mo
JayHuLBL: @AntoineGautier Thanks for catching it. I have reverted the changes and Rerun the unit test. There is no reference results change.

AntoineGautier (approved):

JayHuLBL:
@mwetter Can you merge this PR if you are fine with the changes? I have no access to merge PR.

PR #3165, opened 2022-11-21
Issue2302 building time series enhancement

dhblum:
This is to close https://github.com/lbl-srg/modelica-buildings/issues/2302.  Still to be done:

- [x] Update release notes

Issue #2302, opened 2020-12-09
Feature enhancement of BuildingTimeSeries


AntoineGautier:
The feature branch is `issue2302_timeSeries`.

This issue will address the following:

- Provide a generic formulation suited for loads at the ETS level or computed by decoupled whole-building simulations (e.g., EnergyPlus), see below for the detailed requirements.
- Compute the scaling parameters in `BuildingTimeSeries` based on the peak loads.
- Relocation of `BuildingTimeSeries` (currently under `DHC/Loads/Examples/BaseClasses`) and proper documentation so that the model can be used as a generic model from now on.
- Allow for specifying a recirculation flow rate at minimum pump speed in `Buildings.Experimental.DHC.Loads.FlowDistribution`.
- Use signed real values for nominal heat flow rates and Delta-T.
- *Bug fix in `PartialTerminalUnit`, see https://github.com/lbl-srg/modelica-buildings/pull/2269#issuecomment-741963606: This is no more a bug as the scaling mechanism implemented in #2291 is now consistent with the former option `have_scaLoa=true`. See the documentation of `PartialTerminalUnit` for a precise description.*

## Requirements for a generic formulation of the coupling with time series

1. The refactoring must allow for using either 

    - mass flow rate and Delta-T, 
      - this should be the preferred option for a loose coupling with URBANopt-SDK EnergyPlus simulation. TODO: further specify the modeling requirements on the SDK side, cf. feedback from Christopher Mackey
      
        > Some of the ~80 HVAC templates (within the openstudio-standards gem, see https://docs.google.com/document/d/1xg20cckHsQ6s0z4QUEodQIko9SDK41ejyNezXnBfMDY/edit) have district heating/cooling while others do not. So the ability to integrate with DHC is going to be on a template-by-template basis.

[...]

PR #3222, opened 2023-01-25
Issue2912 district cooling rev2

mwetter:
This adds changes to `issue2912_districtCooling`.

It removes parameters that are related to heating but used in cooling only configurations.
It also sets a default for the air-side mass flow rate through the cooling coil in the room.

@dhblum @khinkelman Is this in principle correct? In `issue2912_districtCooling` there are various heating related parameters that I think should be removed if a model is only for cooling.

@AntoineGautier : Is there any problem regarding the modeling pattern if heating related instances and parameters are removed?

Also,
```modelica
parameter Modelica.Units.SI.HeatFlowRate QHea_flow_nominal(
    min=Modelica.Constants.eps)=Buildings.Experimental.DHC.Loads.BaseClasses.getPeakLoad(
    string="#Peak space heating load",
    filNam=Modelica.Utilities.Files.loadResource(filNam))
    "Design heating heat flow rate (>=0)"
```
is read for a cooling only model, but users don't likely have such data series if they are interested in cooling only.

Also, the assignment
https://github.com/lbl-srg/modelica-buildings/compare/issue2912_districtCooling...issue2912_districtCooling_rev2?expand=1#diff-614f52ebe519aeccf876d07b49208ffb1d061f3ce73f8d88fa13be5c799cded2L54
is confusing. I think it kind of worked because heating load data were read, but then these are used and assigned to cooling. In tropical climate, this however does not work.


AntoineGautier:
@mwetter Yes, this is annoying. However, the design heating load is needed to parameterize the instance of `Experimental.DHC.Loads.BaseClasses.SimpleRoomODE` which uses it to approximate the building lumped thermal conductance.
Note that there was an attempt through #2302 to redesign the coupling with time series, and overcome most of the limitations of the current approach. I started implementing the proposed approach in the feature branch `issue2302_timeSeries`  (see the documentation of the class `Buildings.Experimental.DHC.Loads.BaseClasses.EnergyMassFlow`). But I had to deal with some complexities that were not anticipated and did not have the opportunity to wrap this up.

Regarding the assignment of `mLoaCoo_flow_nominal` this should not be modified from the default value as explained in the documentation of `Experimental.DHC.Loads.BaseClasses.Examples.BaseClasses.BuildingTimeSeries` that was recently updated through #3165.

dhblum:
I'm just taking a more detailed look at the ``issue2912_districtCooling`` branch, as well as the ``issue2912_districtCooling_rev2``  @mwetter has started here.  In order to make it easier to read, for what follows, the prefix to all of the mentioned models is ``Buildings.Experimental.DHC.Loads.`` 

I want to point out that the model ``Cooling.BuildingTimeSeriesWithETS`` uses ``Cooling.BaseClasses.BuildingTimeSeries`` whereas the BuildingTimeSeries model that was updated in https://github.com/lbl-srg/modelica-buildings/pull/3165 is ``BaseClasses.Examples.BaseClasses.BuildingTimeSeries``, which is used by ``Combined.BuildingTimeSeriesWithETS``.  My guess (perhaps someone can confirm) is that for ``issue2912_districtCooling``, the model ``Cooling.BaseClasses.BuildingTimeSeries`` is a duplicate of an earlier version of ``BaseClasses.Examples.BaseClasses.BuildingTimeSeries``, but will not inherit the recent updates.  I'd like to make sure proposed changes are aware of that.  Perhaps ``Cooling.BuildingTimeSeriesWithETS`` should use ``BaseClasses.Examples.BaseClasses.BuildingTimeSeries``, and any proposed changes meant to be made on ``Cooling.BaseClasses.BuildingTimeSeries``, actually be made on ``BaseClasses.Examples.BaseClasses.BuildingTimeSeries``.

mwetter:
- [ ] @dhblum : On `issue2912_districtCooling`, the model `Cooling.BaseClasses.BuildingTimeSeries` is indeed a duplicate of an earlier version of `BaseClasses.Examples.BaseClasses.BuildingTimeSeries`. I will work on resolving this issue.

dhblum:
Thanks @mwetter.  There may still be opportunities for improvement of ``BaseClasses.Examples.BaseClasses.BuildingTimeSeries`` in the cooling-only context.

PR #3223, opened 2023-01-27
IBPSA merge 1682

hcasperfu:
This merges changes made in https://github.com/ibpsa/modelica-ibpsa/issues/1682.


Buildings/Resources/ReferenceResults/Dymola/Buildings_Air_Systems_SingleZone_VAV_Examples_Guideline36.txt
```diff
@@ -8,14 +8,14 @@ statistics-initialization=
 statistics-simulation=
 {
   "linear": "11, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0",
-  "nonlinear": "1",
+  "nonlinear": "2",
```
mwetter, line 11: I noted the increase in the dimension of the nonlinear system. Looking at dsmodel.mof, it seems to be due to different tearing. I don't think there is anything we can do to avoid it.

mwetter (approved):

hcasperfu:
Some models under `Buildings.ThermalZones.EnergyPlus_9_6_0.Examples.SmallOffice` now also fail with optimica. The error message "FMUException: Failed to update the events at time: xxx." is the same as observed with the fan model branch (e.g. @ 924d0ad).
I excluded these models in 7930e8e.
I am preparing a bug report email to Modelon support right now.

PR #3253, opened 2023-02-15
Issue3141 ramp up down

### LINKED ISSUE #3141: add CDL.Continuous.RampUpDown
This issue is to add new CDL block `RampUpDown` to package `CDL.Continuous`.

- when the binary `up` input is `true`, it will ramp up the analog output to maximum value (`1`). The ramp up action lasts `upDuration` time.
- when the binary `down` input is `true`, it will ramp down the analog output to minimum value (`0`). The ramp action lasts `downDuration` time.
- when both the `up` and `down` inputs are false, it stays at current value.

JayHuLBL:
This closes #3141.


JayHuLBL:

PR #3267, opened 2023-03-01
Added smooth(0, ...) to Continuous.Switch

mwetter:
This is required for a model that controls PV and battery

PR #3268, opened 2023-03-01
Added smooth(0, ...) to Continuous.Switch (maint_9.1.x)

mwetter:
This is required for a model that controls PV and battery


mwetter:
This won't be applied to the 9.1 branch, see https://github.com/lbl-srg/modelica-buildings/pull/3267#issuecomment-1450587671

PR #3272, opened 2023-03-06
Made the sci_reader.c robust for data stored with carriage return

chria:
If the CFD resource files (for example https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Resources/Data/ThermalZones/Detailed/Examples/FFD/WindowWithShade.cfd ) are stored on Windows with Windows line endings that includes the carriage return character (\r), the data will fail to load due to that the names read in the sci_reader file will contain an additional character (\r) appended to their names.

This was seen when running the example Buildings.ThermalZones.Detailed.Examples.FFD.WindowWithShade when the Buildings library was copied from a Windows machine to a Linux machine.

This pull request will make the reading of the names more robust and will handle data stored with Windows line endings.

### DIFF
--- a/Buildings/Resources/src/FastFluidDynamics/sci_reader.c
+++ b/Buildings/Resources/src/FastFluidDynamics/sci_reader.c
@@ -221,7 +221,7 @@ int read_sci_input(PARA_DATA *para, REAL **var, int **BINDEX) {
       .......................................................................*/
       fgets(string, 400, file_params);
       /* Get the length of name (The name may contain white space)*/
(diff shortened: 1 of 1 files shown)


JayHuLBL:
@mwetter The change looks good to me.

mwetter:
@JayHuLBL : If the changes look good to you can you please add it to a development branch so it runs through the CI testing and then merge it to the master branch.

JayHuLBL:
Close this PR as the change will be included in PR #3280.

PR #3273, opened 2023-03-06
Made the exchange of data more robust by making sure that the outputs are always copied

chria:
If there are multiple calls to the cfdExchangeData methods (at the same time point), this method will early return without assigning values to its output. The method will return that it succeeded but there are no values in the outputs which is problematic as there is no indication to the Modelica tool that these values should not be used.

This pull request will make this method more robust by always assigning values to the outputs (at the cost of some additional work). Instead of an early return - I added a value to indicate if data should be written to CFD or not - If there are multiple calls to the method (at the same time) the data to CFD is skipped and only the outputs are assigned.


JayHuLBL:
Close this PR as the change will be included in PR #3280.

PR #3278, opened 2023-03-07
added units to the input and output of the trigonometric functions

### LINKED ISSUE #3277: Add proper units to trigonometric blocks in CDL.Continuous
The trigonometric blocks, such as `Sin` and `Asin`, declare their inputs as `Real`. This needs to be changed so that the input of `Sin` has units of `rad` and the output of `Asin` has units of `rad`. Then they have the same behavior as the similar blocks from the Modelica Standard Library.

This also needs to be updated in the ASHRAE Standard 231P draft.

This likely then needs examples to be updated for the unit tests to work with Dymola 2023x Refresh 1 beta 1, see also https://github.com/lbl-srg/modelica-buildings/pull/3276

JayHuLBL:
This closes #3277.


mwetter (approved):

mwetter:
@JayHuLBL : This also works with Dymola 2023x refresh1 beta1.

Issue #3274, opened 2023-03-06
Logs through FastFluidDynamics and ModelicaError is (likely) problematic 


chria:
**Describe the bug**
When looking into the model Buildings.ThermalZones.Detailed.Examples.FFD.WindowWithShade I detected some issues with how the logging is handled in FastFluidDynamics and how they are propagated to a Modelica tool through the ModelicaError. 

What triggered the investigation is that I received nonsense values when a ModelicaError message was triggered. This lead me to look into how the messages was created and sent to the Modelica tool and looking at the method "modelicaError" - https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Resources/src/FastFluidDynamics/ffd.c#L231 - I have a hard time understanding how this can work. 

Looking at the method, a message is sent to it and then memory is allocated to keep this message, see https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Resources/src/FastFluidDynamics/ffd.c#L233 . Next the message that was sent to the method is stored in this newly allocated memory. However, just the next lines, this memory is then deallocated and the msg is thrown away, see https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Resources/src/FastFluidDynamics/ffd.c#L245 . 

If I understand the code (I may though misunderstand a few things) this message is supposed to be picked up for instance at https://github.com/lbl-srg/modelica-buildings/blob/e4b04ecc63eb8a9da056c9f86dec5c06b191526f/Buildings/Resources/C-Sources/cfdSendStopCommand.c#L34 but as it only lives for a fraction of time, I don't see how it works?

If I comment the line at https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Resources/src/FastFluidDynamics/ffd.c#L245 the messages are provided correctly to ModelicaError and the Modelica tool. However, this will lead to a memory leak so something else is needed I think.

chria:
I created a pull request here: https://github.com/lbl-srg/modelica-buildings/pull/3283 that will resolve it for the case that I'm looking for. I'm not sure though if it requires other changes elsewhere.

mwetter:
@chria : Thanks for the PR.

@JayHuLBL : Can you please run this through the CI testing and also check if there are other such occurrences that need to be corrected.

PR #3283, opened 2023-03-08
Moved allocation of the ModelicaError message to the constructor and destructor

chria:
As highlighted in https://github.com/lbl-srg/modelica-buildings/issues/3274 the logging to a Modelica tool is problematic as the error messages are deallocated almost instantaneously. 

In this pull request I moved the allocation of the memory for the msg to the constructor / destructor instead.


JayHuLBL:
Close this PR as it has been included in #3296. This will enable CI tests.

Issue #3288, opened 2023-03-10
Add DX coil for heating


mwetter:
This issue is to add DX coil models for heating.

The development branch is `issue3288_dxCoilHeating_renameCooledSource`, and development is ongoing at https://github.com/terrancelu92/modelica-buildings/tree/dxcoil_heating_example

mwetter:
Items for discussion
- [ ] Are coils for heating only, or switch-over?
- [ ] Evaluate how to implement multi-speed and variable speed by adapting current base models.
- [ ] Rename package from `AirCooled` to `AirSource` and `WaterCooled` to `WaterSource` (@mwetter to test change for Buildings 10.0.0)

karthikeyad-pnnl:
@mwetter Regarding the use of weather-bus to read secondary coil conditions: I confirmed that the current DX cooling component template for ctrl-flow uses a weatherbus to read the outdoor air/condenser-water temperature. Therefore, we should be able to use a weather bus input even on the coil baseclasses. Seen below: Buildings.Templates.Components.Coils.EvaporatorVariableSpeed from #1374.

![image](https://user-images.githubusercontent.com/59938817/224448666-071fdf6b-50a3-43da-9340-68fda3066ff3.png)

mwetter:
@karthikeyad-pnnl : Great, thanks for checking and confirming this compatibility with the templates.

karthikeyad-pnnl:
@mwetter We had a couple of questions about the integration of the DX heating coil:
- What is the required deadline for inclusion in Buildings library 10.0.0? Are we still planning to include this feature in it?
- We are going to assume branch [issue3288_dxCoilHeating_renameCooledSource](https://github.com/lbl-srg/modelica-buildings/tree/issue3288_dxCoilHeating_renameCooledSource) as the base branch on which we will make the PR with the proposed changes. We would like to confirm it with you.

We will be target completion of the following tasks by 03/31/23:
- [ ] Addition of partial baseclass that is common to both DX heating and cooling coil.
- [ ] Addition of separate baseclasses specifically for DX heating and cooling coil each.
- [ ] Addition of complete class definitions for DX heating and cooling coil each.
- [ ] Addition of partial curve record baseclass that is common to both cooling and heating coil.
- [ ] Addition of curve record baseclass for heating and cooling coil each.
- [ ] Addition of validation model for DX heating coil.

mwetter:
[...]

PR #3289, opened 2023-03-10
Renamed AirCooled to AirSource, and same for WaterCooled

mwetter:
This is for #3288


mwetter:
This is ready to merge. There will be additional non-backward compatible changes, such as changing the temperature input to be a weather data bus because the heating coils also need outside humidity as an input for the defrost calculation.

PR #3313, opened 2023-03-28
Issue3256 proof

### LINKED ISSUE #3256: DO/DI Proof block to verify proper equipment operation
This issue is to implement a new CDL block `Controls.OBC.CDL.Logical.Proof`, to verify if the equipment is operating properly.

The block compares two boolean inputs, `u` which is current status of the equipment and `command` which is the commanded status of the equipment. If the two inputs do not receive the same, the block provides two outputs that can trigger alarms.

The block will have two parameters, `Delay` and `Debounce`. The parameter `Delay` specifies an allowable delay between the time when the `command` becomes `true` and the time when the `u` becomes `true`. The parameter `Debounce` specifies the amount of time that the `u` must remain `true` or `false` before it is considered valid. The `Debounce` should not be greater than the `Delay`.

It will have two boolean outputs, `alarm` and `hand`. The `alarm` becomes `true` when the `command` is `true` but the `u` is `false`, indicating the the equipment is not running when it should have been turned on. The `hand` becomes `true` when the `command` is `false` but the `u` is `true`, indicating that the equipment is still running when it should have been turned off.

JayHuLBL:
This closes #3256.


Buildings/Controls/OBC/CDL/Logical/Proof.mo
```diff
@@ -0,0 +1,251 @@
...
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanOutput y1
+    "True: commanded status is true and current status is false"
+    annotation (Placement(transformation(extent={{160,110},{200,150}}),
+        iconTransformation(extent={{100,40},{140,80}})));
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanOutput y2
+    "True: commanded status is false and current status is true"
+    annotation (Placement(transformation(extent={{160,-110},{200,-70}}),
+        iconTransformation(extent={{100,-80},{140,-40}})));
+
+  Buildings.Controls.OBC.CDL.Logical.TrueDelay valTru(
```
mwetter, line 30: If there is a need to have them public, then they need to have good names rather than valTru and valTru1. I made them protected for now. This also needs update to the tests.

Buildings/Controls/OBC/CDL/Logical/Proof.mo
```diff
@@ -0,0 +1,251 @@
+within Buildings.Controls.OBC.CDL.Logical;
+block Proof "Verify two boolean inputs"
+  parameter Real validDelay(
+    final quantity="Time",
+    final unit="s")
+    "Delay to valid input. The input must remain unchanged in the time";
```
mwetter: must remain unchanged for what to happen? This is not clear.

mwetter (changes requested):
Please see my inline comments and revisions to the documentation. We can discuss tomorrow.

Buildings/Controls/OBC/CDL/Logical/Proof.mo
JayHuLBL: Leaving these two instances to be public allows in the validation to show their values. For the script itself here, there is no other special need. I agree to move it to the protected section and the tests were updated.

Buildings/Controls/OBC/CDL/Logical/Proof.mo
```diff
@@ -0,0 +1,251 @@
...
+
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanInput uMea
+    "Current status"
+    annotation (Placement(transformation(extent={{-220,110},{-180,150}}),
+        iconTransformation(extent={{-140,40},{-100,80}})));
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanInput uCom
+    "Commanded status"
+    annotation (Placement(transformation(extent={{-220,-30},{-180,10}}),
+        iconTransformation(extent={{-140,-80},{-100,-40}})));
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanOutput y1
```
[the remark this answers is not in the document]
JayHuLBL: Thanks for the suggestion. I changed the output names.

Buildings/Controls/OBC/CDL/Logical/Proof.mo
JayHuLBL: Thanks for revising it.

Buildings/Controls/OBC/CDL/Logical/Validation/Proof.mo
```diff
@@ -0,0 +1,152 @@
...
+</li>
+<li>
+The instances <code>pro1</code> and <code>pro4</code> use a different value
+for <code>feedbackDelay</code>. They test the case in which
+both the inputs change from <code>true</code> to <code>false</code>. However,
+after the input <code>u_s</code> changes, the input <code>u_m</code> remains
+<code>true</code> for a time that is longer than
+</li>
+<li>
+It also tests the case when the measured input <code>u_m</code> cannot stay stable
```
mwetter, line 121: What is "It" in this description? I revised all the other ones to start with the instance names and then describe what they do, but for this item I don't know what it means. I don't think it is `pro5` as its outputs do not remain true. Either there is a bug in the `Proof` block, or the example is not good, or the description is not about `pro5`, but then there is no explanation for `pro5`.
image

mwetter (changes requested):

Buildings/Controls/OBC/CDL/Logical/Validation/Proof.mo
JayHuLBL: Thanks for the comments. It indeed is a bug in the block. I have corrected the implementation and updated the validation model's documentation.

mwetter (approved):

PR #3316, opened 2023-03-29
Corrected ramp implementation

### LINKED ISSUE #3314: Avoid output drop when activating the output ramping
In current `CDL.Continuous.Ramp` implementation, the output might drop at the moment when the boolean input `activate` becomes `true`. See below: 

Ramp

This issue is to avoid the drop. However, according to the comment from 231p committee, it is allowed to have the output jumping up at the moment when the boolean input changes from `true` to `false`.

JayHuLBL:
This closes #3314.


JayHuLBL:

mwetter (approved):

PR #3334, opened 2023-04-11
in OBC package moved PID controller instance to public section

### LINKED ISSUE #2477: Change G36 PR1 so that PI controllers are public
This changes `Controls.OBC.ASHRAE.G36_PR1.AHUs.SingleZone.VAV.CoolingCoil` so that the PID controller is public. This allows inspecting its inputs, outputs and internal public data members which is often needed during control tuning.

JayHuLBL:
This closes #2477.


Buildings/Controls/OBC/ASHRAE/G36/ThermalZones/ControlLoops.mo
```diff
@@ -53,15 +53,20 @@ block ControlLoops "Heating and cooling control loops"
     final unit="1") "Heating control signal"
     annotation (Placement(transformation(extent={{160,-90},{200,-50}}),
       iconTransformation(extent={{100,-80},{140,-40}})));
-
-protected
   Buildings.Controls.OBC.CDL.Continuous.PIDWithReset cooCon(
```
mwetter, line 56: Shouldn't they rather be called `conCoo` and `conHea` and usually we start controllers with `con`.

Buildings/Controls/OBC/ASHRAE/G36/AHUs/SingleZone/VAV/SetPoints/CoolingCoil.mo
```diff
@@ -42,6 +42,15 @@ block CoolingCoil "Controller for cooling coil valve"
     annotation (Placement(transformation(extent={{100,-20},{140,20}}),
         iconTransformation(extent={{100,-20},{140,20}})));
 
+  Buildings.Controls.OBC.CDL.Continuous.PIDWithReset cooCoiPI(
```
mwetter, line 45: I would probably call this `conCoi` as it is not always a PI controller, and then we start controllers with `con`

Buildings/Controls/OBC/ASHRAE/G36/AHUs/MultiZone/VAV/Economizers/Subsequences/Limits/SeparateWithAFMS.mo
```diff
@@ -140,6 +140,16 @@ block SeparateWithAFMS
     annotation (Placement(transformation(extent={{220,-310},{260,-270}}),
         iconTransformation(extent={{100,-110},{140,-70}})));
 
+  Buildings.Controls.OBC.CDL.Continuous.PIDWithReset minOACon(
```
mwetter, line 143: Shouldn't this be `conMinOA` to start controllers with `con`?

mwetter (commented):
Please see comments for more uniform naming, which at this point is easy to do as they were protected before. So this should not break anything.

JayHuLBL:
@mwetter The comments have been addressed. Should we also add the changes to branch `maint_9.1.x`?

mwetter:
> @mwetter The comments have been addressed. Should we also add the changes to branch `maint_9.1.x`?

As these are changes to names of public instances, the changes are not backward compatible. Therefore they should not be applied to `maint_9.1.x`.

PR #3341, opened 2023-04-13
limited the fontSize check only in the documentation sections

### LINKED ISSUE #3340: Limit the fontSize check in info section
In `bin/verifyFiles.py`, the regular expression at https://github.com/lbl-srg/modelica-buildings/blob/f2d33a2cede64599459b484dbbc923e9506d118d/bin/verifyFiles.py#L59 is too stringent. It should only flag an error if used within the info section.

JayHuLBL:
This closes #3340.

@mwetter Should the change also be added to `modelica-ibpsa` repo?


mwetter:
@JayHuLBL : I changed fontSize= to font-size: as this is the css syntax. Not sure why we still tested for the old syntax.

However, I don't think the regular expression is correct (or the code that uses it). See commit  https://github.com/lbl-srg/modelica-buildings/commit/ba15aea90272487dcd5f4012e5049743ab860811 which should be flagged as an error.

JayHuLBL:
@mwetter I corrected the regular expression and merged the change to the branch [issue3340_fontSizeCheck_testOnly](https://github.com/lbl-srg/modelica-buildings/compare/issue3340_fontSizeCheck_testOnly). It can successfully flag the error.

However, the check cannot flag the errors in [line 195](https://github.com/lbl-srg/modelica-buildings/blob/5bebd39ce4a0dc2e9eebfc85682bdfd94826a0bb/Buildings/HeatTransfer/Conduction/SingleLayer.mo#L195) and [line 233](https://github.com/lbl-srg/modelica-buildings/blob/5bebd39ce4a0dc2e9eebfc85682bdfd94826a0bb/Buildings/HeatTransfer/Conduction/SingleLayer.mo#L233). I think the regular expressions `(Text\s*\([^\).]*)lineColor` and `(Line\s*\([^\).]*)lineThickness` in `verifyFiles.py` should be corrected as well.

mwetter:
@JayHuLBL : Yes, can you please correct these entries as well.

JayHuLBL:

PR #3378, opened 2023-05-02
Add type casts to malloc and realloc in C-Sources

rfranke:
Numerous examples fail to compile with the OpenModelica C++ runtime for this reason, see
https://libraries.openmodelica.org/branches/cpp/Buildings_latest/Buildings_latest.html

So far some allocations do cast, some don't. It would be good if all did.
This is required if compiled with C++ to avoid errors like:
```
In file included from ./OMCppBuildings_latest_Buildings.BoundaryConditions.WeatherData.Examples.ReaderTMY3Functions.cpp:7:
Buildings 10.0.0-master/Resources/C-Sources/getTimeSpan.c:27:9: error: cannot initialize a variable of type 'char *' with an rvalue of type 'void *'
  char *result = malloc(len1 + len2 + 1);
        ^        ~~~~~~~~~~~~~~~~~~~~~~~
```
or
```
In file included from ./OMCppBuildings_latest_Buildings.Utilities.Plotters.Examples.ScatterFunctions.cpp:8:
Buildings 10.0.0-master/Resources/C-Sources/plotInit.c:42:21: error: assigning to 'char **' from incompatible type 'void *'
    plotFileNames = malloc(sizeof(char*));
                    ^~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/Buildings/Resources/C-Sources/exchangeValues.c
+++ b/Buildings/Resources/C-Sources/exchangeValues.c
@@ -34,7 +34,7 @@ double exchangeValues(void* object, size_t iX, double x, size_t iY){
   /* Manage memory
    * At first call, initialize storage */
   if ( table->x == NULL ){
-    table->x= malloc( nNew * sizeof(double) );
+    table->x = (double *)malloc( nNew * sizeof(double) );
     if ( table->x == NULL )
       ModelicaError("Out of memory in storeValue.c when allocating memory for table.");
     table->n = nNew;
@@ -47,7 +47,7 @@ double exchangeValues(void* object, size_t iX, double x, size_t iY){
 
   if (iX > nTab){
     /* Assign more memory before storing the value */
-    tab2 = malloc( nTab * sizeof(double) );
+    tab2 = (double *)malloc( nTab * sizeof(double) );
     if ( tab2 == NULL )
       ModelicaError("Out of memory in storeValue.c when allocating memory for tab2.");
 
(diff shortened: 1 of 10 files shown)


mwetter:
@rfranke : Thanks for the pull request. We will integrate it into the dev branches in the next days and also move the changes upstream to the Modelica IBPSA Library.

hcasperfu:
@mwetter -
As discussed in the morning meeting, I checked whether all `malloc()` and `realloc()` are covered.
I think all of them under the following two folders now have type casts:
- `Buildings/Resources/C-Sources`
- `Buildings/Resources/src/ThermalZones/EnergyPlus_9_6_0/C-Sources/`

However if I `grep` the whole `Buildings/Resources` folder, there seem to be some stray ones.
```
Buildings/Resources$ grep -or 'malloc(' * | wc -l
157
Buildings/Resources$ grep -or '/*)malloc(' * | wc -l
59
Buildings/Resources$ grep -or '/*) malloc(' * | wc -l
86
```
where 59 + 86 = 145 < 157. (Among all occurrences of the string `'malloc('`, not all of them are preceded by `'*)'` or `'*) '`.)
There are, for example:
```
Buildings/Resources$ grep -r '= malloc(' *
src/python36/pythonInterpreter.c:	pythonPtr* ptr = malloc(sizeof(pythonPtr));
src/python36/pythonInterpreter.c:		ptrMemory->pythonPath = malloc(sizeof(char) * (lenPath + 1));
src/python36/testProgram.c:  pythonPtr* ptr = malloc(sizeof(pythonPtr));
src/python_3_8/pythonInterpreter.c:	pythonPtr* ptr = malloc(sizeof(pythonPtr));
src/python_3_8/pythonInterpreter.c:		ptrMemory->pythonPath = malloc(sizeof(char) * (lenPath + 1));
src/python_3_8/testProgram.c:  pythonPtr* ptr = malloc(sizeof(pythonPtr));
```
Do we need to deal with those as well?

mwetter:
@hcasperfu : it would be good to also correct these additional ones. I can recreate the needed binaries once the C code is updated.

hcasperfu:
Moved to PR #3383 so that I can push.

Issue #3380, opened 2023-05-03
Error while converting .epw to .mos


salomehohl:
I'm trying to convert weather data from here https://re.jrc.ec.europa.eu/pvg_tools/en/ into a .mos file. See file below.
[Karlsruhe Eggenstein.zip](https://github.com/lbl-srg/modelica-buildings/files/11381430/Karlsruhe.Eggenstein.zip)

I'm using the java commands and get an error message: "The data at line 9 column 23 in .epw file is missing. You need to fix it before the convert." 
The error does not occur while converting .epw files from other websites.

I'm using
Dymola 2022
Buildings 9.1.0 with Modelica 4.0.0
java
openjdk version "17.0.7" 2023-04-18 LTS
OpenJDK Runtime Environment Corretto-17.0.7.7.1 (build 17.0.7+7-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.7.7.1 (build 17.0.7+7-LTS, mixed mode, sharing)
on a Windows 10 laptop.

What is the matter with this file? Thank you!

mwetter:
@salomehohl : your .epw file misses the total sky cover. 
In the next version this will be reported with a clearer error message, see [669953e](https://github.com/lbl-srg/modelica-buildings/commit/669953e3d2089ae86f6282699f61247064b7dc7f)

PR #3395, opened 2023-05-10
Improved error message of weather data conversion application if data is missing

mwetter:
For #3380


mwetter:
This PR also removes the doc directory as Javadoc 11.0.18 includes js files that need to be distributed under the GNU license (commit https://github.com/lbl-srg/modelica-buildings/pull/3395/commits/669953e3d2089ae86f6282699f61247064b7dc7f)

PR #3427, opened 2023-06-07
Issue3420 update scripts to latest energy plus

### LINKED ISSUE #3420: Update run.sh scripts to latest EnergyPlus
There are various scripts that update data files produced by EnergyPlus 9.6.0 which are used to validate models. These are invoked here:
https://github.com/lbl-srg/modelica-buildings/blob/d4f1d126684da964cd44876e2eb2435a8e32bfc4/Buildings/Resources/Scripts/travis/Makefile#L18

This issue is to update these idf files to the latest version of EnergyPlus and make a PR to the master.

This change should also be done for the files on branch `issue3288_dxCoilHeating` in the directory Buildings/Resources/Data/Fluid/DXSystems/ (idf files in other directories should not be changed as we can change them by merging the master branch).

hcasperfu:
This closes #3420.
@mwetter -
Please let me know if this addressed the issue.
I updated six idf files invoked by a `run.sh` from 9.6.0 to 23.1.0. One of them caused the mos output file to change, which lead to significant change of two unit test result files under `Fluid.CHPs.Validation`.
Should I also make a separate PR to `issue3288_dxCoilHeating`?


Buildings/Resources/Data/ThermalZones/EnergyPlus_9_6_0/Examples/RefBldgSmallOffice/RefBldgSmallOfficeNew2004_Chicago.idf
mwetter, line 1: The files in `EnergyPlus_9_6_0` will need to be reverted as Spawn is using version 9.6.0. (We will need to update them sometime next fall when we update Spawn)

mwetter (changes requested):
Please see inline comment, and you will also need to update in `.travis.yml` the `EnergyPlus_VERSION` entry (and line 111 in `.travis.yml`). Otherwise travis won't use the right version.

mwetter:
The changes in the reference results are OK, they are because E+ results changed, and the accuracy of the Modelica model looks to be the same.
I updated the CI configuration in https://github.com/lbl-srg/modelica-buildings/pull/3427/commits/a765982612055c11f47ef01a5616b31e4ae8d548 as the Spawn reference results require EnergyPlus 9.6.0.

mwetter (approved):

PR #3429, opened 2023-06-09
Issue3420 update scripts to latest energy plus v2

mwetter:
This is based on https://github.com/lbl-srg/modelica-buildings/pull/3427 and the PR is made as https://github.com/lbl-srg/modelica-buildings/pull/3427 does not seem to be started on travis.


mwetter:
The issue was a misplaced comment.

PR #3436, opened 2023-06-15
Issue3425 documentation chiller EIR factors

### LINKED ISSUE #3425: chiller power consumption calculation
I'm not sure if this is a bug or a calculated feature that I don't understand.
Buildings.Fluid.Chillers.BaseClasses.PartialElectric, line144, buildings library version=9.1.0
// Compressor power.
P = - QEva_flow_ava / COP_nominal * EIRFunT * EIRFunPLR * CR;
EIRFunT: proportion of cooling COP as a function of chilled water outlet temperature and cooling water return temperature;
EIRFunPLR: proportion of cooling COP as a function of partial load factor;
I think that if the chiller's rated chilled water discharge temperature is 7 degrees Celsius and the setting at this point is 9 degrees Celsius, assuming that the COP at this point is 1.1 times the rated COP, then EIRFunT=1.1;
At this point the formula should be  P = - QEva_flow_ava / (COP_nominal * EIRFunT * EIRFunPLR) * CR;
Am I understanding you correctly?

hcasperfu:
This PR provides sources (i.e. [EnergyPlus Engineering Reference](https://energyplus.net/assets/nrel_custom/pdfs/pdfs_v22.1.0/EngineeringReference.pdf)) and details in the documentation for electric chillers.
This closes #3425.

### DIFF
--- a/Buildings/Fluid/Chillers/BaseClasses/PartialElectric.mo
+++ b/Buildings/Fluid/Chillers/BaseClasses/PartialElectric.mo
@@ -378,21 +378,43 @@ for an example with a chiller operating in heating mode.
 </p>
 <h4>Implementation</h4>
 <p>
-Models that extend from this base class need to provide
-three functions to predict capacity and power consumption:
+This implementation computes the chiller capacity and power consumption
+the same way as documented in
+<a href=\"https://energyplus.net/assets/nrel_custom/pdfs/pdfs_v22.1.0/EngineeringReference.pdf\">
+EnergyPlus v22.1.0 Engineering Reference</a>
+section 14.3.9.2.
+The chiller capacity is computed as
+</p>
+<pre>  QEva_flow_ava = QEva_flow_nominal*capFunT;</pre>
+<p>
+and the power consumption is computed as
+</p>
+<pre>  P = -QEva_flow_ava/COP_nominal*EIRFunT*EIRFunPLR*CR;</pre>
+<p>
+(See equations 14.234 and 14.240 in the referenced document.)
+</p>
+<p>
+Models that extend from this base class need to provide the following
+three functions:
 </p>
 <ul>
 <li>
-A function to predict cooling capacity. The function value needs
-to be assigned to <code>capFunT</code>.
+<code>capFunT</code> to predict cooling capacity,
 </li>
 <li>
-A function to predict the power input as a function of temperature.
-The function value needs to be assigned to <code>EIRFunT</code>.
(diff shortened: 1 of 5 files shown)


Buildings/Fluid/Chillers/BaseClasses/PartialElectric.mo
```diff
@@ -378,21 +378,43 @@ for an example with a chiller operating in heating mode.
...
+<a href=\"https://energyplus.net/assets/nrel_custom/pdfs/pdfs_v22.1.0/EngineeringReference.pdf\">
+EnergyPlus v22.1.0 Engineering Reference</a>
+section 14.3.9.2.
+The chiller capacity is computed as
+</p>
+<pre>  QEva_flow_ava = QEva_flow_nominal*capFunT</pre>
+<p>
+and the power consumption is computed as
+</p>
+<pre>  P = -QEva_flow_ava/COP_nominal*EIRFunT*EIRFunPLR*CR.</pre>
```
mwetter, line 392: Can you please also add the definition of these functions. Best may be in `Buildings.Fluid.Chillers.Data.ElectricEIR.Generic`
While referencing the EnergyPlus documentation is fine, we like to have the models to be fully documented and avoid that users need to consult other software documentation to understand the models.

mwetter (changes requested):
Please see inline comment.

Buildings/Fluid/Chillers/BaseClasses/PartialElectric.mo
hcasperfu: To clarify, do you mean adding the documentation of these two formulae to `Data.ElectricEIR.Generic` as well?

hcasperfu: Addressed in cf4d6ec. I added a table detailing how these functions are formulated.

hcasperfu: I revised the documentations further in cb19747.
I moved the table to `Buildings.Fluid.Chillers.BaseClasses.PartialElectric` which I feel is closer to a "center of traffic" that makes sense to provide information across models. Using `capFunT` as an example, now in the same document it is easy to find both "what it is" (i.e. `QEva_flow_ava = QEva_flow_nominal * capFunT` up-text) and "how it is computed" (in the table) close to each other.
Please let me know if this addressed your concern.

mwetter (approved):

PR #3442, opened 2023-06-22
Issue3441 correct nominal flow modification

### LINKED ISSUE #3441: Correct nominal flow modification in DryCoilDiscretized
First reported by @JBjoernskov.
This issue is to correct the following line: 
https://github.com/lbl-srg/modelica-buildings/blob/115ea729ea19eb45ff817b40e2276ed8cd49e1b4/Buildings/Fluid/HeatExchangers/DryCoilDiscretized.mo#L99

where it should be
`    each final m2_flow_nominal=m2_flow_nominal/nPipPar/nPipSeg,`

This correction is expected to affect the dynamics and therefore unit tests will be rerun.

hcasperfu:
This closes #3441.


Buildings/package.mo
```diff
@@ -1066,6 +1066,12 @@ that can lead to wrong simulation results):
                        <a href=\"https://github.com/lbl-srg/modelica-buildings/issues/3151\">Buildings, issue 3151</a>
     </td>
 </tr>
+<tr><td valign=\"top\">Buildings.Fluid.HeatExchangers.DryCoilDiscretized
```
mwetter, line 1069: @hcasperfu : Note that this change only needs to be added to the release notes on the `maint_9.1.x` branch, but not the master. The reason is that we will release 9.1.1 (or 9.2.0), and then copy the release notes of 9.1.1 to the release history in the master before releasing 10.0.0. Hence, this change is not really done new in 10.0.0, but rather in its earlier release 9.1.1.

mwetter (changes requested):
@hcasperfu : See the minor inline comment, otherwise it looks good for merging when the tests pass.

Buildings/package.mo
hcasperfu: Understood! I will work on the maintenance branches once the tests pass here and then move the release notes.

hcasperfu: Reverted in 234312e.

mwetter (approved):

PR #3452, opened 2023-06-28
Issue3439 spawn ptac window ac

mwetter:
This merges the PTAC window AC units to the master.


mwetter:
- [ ] PackageTerminalAirConditioner and WindowAC both have hot and cold water connections. These need to be removed if they are not used at all.

PR #3519, opened 2023-09-13
Renamed CDL.Discrete.Examples package

### LINKED ISSUE #3517: Change Example package to Validation package
This issue is to change the `OBC.CDL.Discrete.Examples` to `OBC.CDL.Discrete.Validation`.

JayHuLBL:
This closes #3517.


mwetter (changes requested):
Please also add an entry in the conversion script (and test whether the entry can be just the package name).
I also changed the html formatting to use `<code>...</code>`.

JayHuLBL:
@mwetter I tested the conversion script that uses package name as the entry. It can successfully convert the class.

mwetter (approved):

Issue #3526, opened 2023-09-18
Compatibility of G36 VAV multiple-zone controller with ctrl-flow options


AntoineGautier:
**Valve feedback position**

The controller `Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.Controller` uses the signals `uCooCoi_actual` and `uHeaCoi_actual` as inputs. (`uHeaCoi_actual` is only instantiated if `have_hotWatCoi` is `true`.)
This is not consistent with ASHRAE Guideline 36 (2021) which does not specify these signals as input points in Section 4.6.
The guideline rather specifies in Section 5.1.18 that, for modulating (analog) actuators,

> position may be assumed to be equal to ***analog signal to*** actuator.

The current implementation is therefore not compatible with the following equipment options from https://github.com/lbl-srg/ctrl-flow-dev/blob/main/server/scripts/sequence-doc/src/version/Current%20G36%20Decisions/Guideline%2036-2021%20(mappings).csv

- No cooling coil
- DX coil

The input signals `uCooCoi_actual` and `uHeaCoi_actual` should be removed from `Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.Controller` which should rather use the computed ***AO*** signal instead (with a `pre` operator if needed to avoid algebraic loops).

**Electric heating coil**

Within `Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.Controller`, the output connector `yHeaCoi` is instantiated with the condition `if have_hotWatCoi`, which makes the controller incompatible with the electric coil option from https://github.com/lbl-srg/ctrl-flow-dev/blob/main/server/scripts/sequence-doc/src/version/Current%20G36%20Decisions/Guideline%2036-2021%20(mappings).csv

**Two-position relief damper**

The controller `Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.Controller` only computes an AO signal to actuate the relief damper.
This is not compatible with configurations with a relief fan that require a two-position relief damper, i.e., a ***DO*** signal to be generated by the controller. See ASHRAE Guideline 36 (2021) Section 4.6.  

**Singular system when freezestat is hardwired to equipment**

The controller `Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.Controller` contains the following connect clause.
```
connect(frePro.yHotWatPlaReq, intSwi.u1)
```
[...]

PR #3528, opened 2023-09-19
Templates • Testing automation

### LINKED ISSUE #3520: Templates • Automate VAV template testing
This ticket is to automate the tests of VAV templates to cover all valid system configurations.
This will use the same test workflow as the one implemented in `issue3266_template_HW_plant`.
The feature branch is `issue3520_templateTesting`.

----
#### About Travis error 137

`The command make test-templates-dymola exited with 137.`

Contrary to what the documentation at https://guides.codepath.com/android/Setting-up-Travis-CI#troubleshooting suggests, the error occurs due to `Timeout (90 minutes) reached.`

As a workaround, some use a background loop that periodically invokes `echo`, see https://github.com/travis-ci/travis-ci/issues/4192#issuecomment-385265664. 
I guess the main benefit is that it is not needed to gauge the execution time. The main drawback is that a non-finishing job cannot be terminated by time out.

For now, we just check the test argument to extend the time out (to 180') when testing the templates.
https://github.com/lbl-srg/modelica-buildings/blob/issue3520_templateTesting/.travis.yml#L178-L184

----
#### About checksum verification to trigger testing

As implemented, the developer has to specify which packages to be considered for computing the checksum and comparing it to the checksum computed over the same files from the master branch. This logic is brittle: it may triggers testing when no used class is modified, or skip testing when a used class is modified.

- In the future, we can improve this by using https://openmodelica.org/doc/OpenModelicaUsersGuide/latest/scripting_api.html#getusedclassnames which, in a couple of seconds, returns the list of all classes used by a given model.

- At the moment, the main validation model stored in the library (and used by the Python script to test all supported system configurations) is still part of the standard CI tests (based on mos script). We at least have this model tested for each PR, even if the checksum test does not trigger exhaustive testing.

AntoineGautier:
This closes #3520: please read the comments about time out and checksum verification on the issue page.

The high number of changed files is due to the fact that only one validation model is kept for each system type. All supported configurations are then generated with Python scripts by modifying this unique validation model.

The main developments from this PR are within `Buildings/Resources/Scripts/travis/templates`.

This PR also includes some fixes needed to support some configurations that had not been validated before and that the new test workflow exposed as faulty. Most of these are related to #3526. The rest are due to missing parameters for the number of return and relief fans within the configuration record.

TODO before merging:

- [x] Wait for #3531 to be merged and update test script to cover all newly supported configurations.
- [x] Update documentation.
- [x] Revert 262276f60a.
- [x] Add command-line flag to indicate that the tests are triggered by checksum verification (to allow for future alternatives, such as periodic testing)


Buildings/Resources/Scripts/travis/templates/core.py
```diff
@@ -0,0 +1,487 @@
...
+        s.addPreProcessingStatement(r'openModel("../Buildings/package.mo", changeDirectory=false);')
+    if simulator == 'optimica':
+        # Set MODELICAPATH (only in child process, so this won't affect main process).
+        os.environ['MODELICAPATH'] = os.path.abspath(os.pardir)
+
+    for modif in arg[1]:
+        s.addModelModifier(modif)
+
+    s.setSolver("CVode")
+    s.setTolerance(1e-6)
```
mwetter, line 121: Is it good to override the settings here rather than use the one from the `experiment` annotation, possibly overridden from `conf.yml`?

mwetter (commented):
@AntoineGautier : the setup for the testing looks good to me. I only added one inline comment as I am not sure what the best way is for configuring solver options. (The `conf.yml` file is read by then the solver is hard coded.)

AntoineGautier:
@mwetter This is ready to merge in my opinion. Crossing fingers that this will not destroy MBL CI testing.

PR #3531, opened 2023-09-20
Multizone AHU controller: Removed measured input, fixed singular issues

### LINKED ISSUE #3526: Compatibility of G36 VAV multiple-zone controller with ctrl-flow options

JayHuLBL:
This closes #3526.
- [x] removed the measured damper position inputs
- [x] added 2-position relief damper boolean output
- [x] added electric heating coil option
- [x] fixed singular issues


AntoineGautier (changes requested):
We are getting very close!
I used this script to test ***all*** valid combinations of the controller parameters (this is based on the developments from https://github.com/lbl-srg/modelica-buildings/pull/3528).
https://github.com/AntoineGautier/modelica-buildings/blob/tmp/issue3526_connectors/Buildings/Resources/Scripts/travis/templates/VAVMultiZoneController.py
It turns out that only ***11 %*** of the tested combinations fail to simulate (with a high probability of a recurring pattern that could explain most of them). I am sending the log file to you by email so that you can inspect and debug the faulty ones.

In addition, I have the following comments.

Inside `Buildings.Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.SetPoints.PlantRequests` the input `uCooCoiSet` is not conditional. The block is still incompatible with the following options.

- No cooling coil
- DX coil

Could we introduce a parameter to specify the type of cooling coil, and adapt the block in a similar manner as with `have_hotWatCoi`?
Also, since we want to cover 3 options for the heating coil (none, HW, electric) and for the cooling coil (none, CHW, DX) I would rather introduce enumerations to describe these options, rather than increasing the number of Boolean parameters, such as `have_hotWatCoi` and `have_eleHeaCoi`. Note also that the implementation currently does not avoid `have_hotWatCoi=true` and `have_eleHeaCoi=true`, which an enumeration would avoid.

AntoineGautier:
Hi Jianjun

Thanks for the quick feedback and corrections for issue #3526.
Following my review at
https://github.com/lbl-srg/modelica-buildings/pull/3531#pullrequestreview-1639511502,
I
am sending you the log file.

Best
--
Antoine

JayHuLBL:
@AntoineGautier I added the new enumeration `Buildings.Controls.OBC.ASHRAE.G36.Types.Coil` that lists coil types. However, it has less items than the one in Template folder `Buildings.Templates.Components.Types.Coil`.

The new enumeration has been applied in MultiZone and single zone VAV controller, and also the terminal unit controllers. Due to the change, I believe the template models also need to update.

AntoineGautier:
@JayHuLBL Noted. I'll refactor the templates via a commit to the branch `issue3526_connectors` early next week.

Buildings/Controls/OBC/ASHRAE/G36/AHUs/MultiZone/VAV/Controller.mo
```diff
@@ -16,36 +16,37 @@ block Controller "Multizone VAV air handling unit controller"
...
+  parameter Buildings.Controls.OBC.ASHRAE.G36.Types.Coil cooCoi=Buildings.Controls.OBC.ASHRAE.G36.Types.Coil.WaterBasedCooling
+    "Cooling coil type"
     annotation (__cdl(ValueInReference=false),
                 Dialog(group="System and building parameters"));
-  parameter Boolean have_eleHeaCoi=false
-    "True: the AHU has electric heating coil"
+  parameter Buildings.Controls.OBC.ASHRAE.G36.Types.Coil heaCoi=Buildings.Controls.OBC.ASHRAE.G36.Types.Coil.WaterBasedHeating
+    "Heating coil type"
     annotation (__cdl(ValueInReference=false),
                 Dialog(group="System and building parameters"));
```
AntoineGautier, line 48: @JayHuLBL I think 2 distinct enumerations should be introduced: one for heating coils, another one for cooling coils.
The current implementation does not prevent from selecting `cooCoi=Buildings.Controls.OBC.ASHRAE.G36.Types.Coil.ElectricHeating` for example (the simulation of `Buildings.Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.Validation.Controller` actually succeeds with this class modification).
Let me know if this is possible. (I'll wait before refactoring the VAV template.)

JayHuLBL:
@AntoineGautier I was thinking about having two distinct enumeration, but realized that in `Template` folder, it also uses one enumeration. Will the same issue exist for the template models?
However, it would be relatively easy to make the change to have two enumuerations. Please let me know your thought about the one in `Template` folder.

AntoineGautier:
The use of the enumeration differs in the Templates package. The enumeration itself is not exposed to the user, but only used "in the background", as a means of type introspection, for example:
```
within Buildings.Templates.Components.Coils;
model ElectricHeating "Modulating electric heating coil"
  extends Buildings.Templates.Components.Interfaces.PartialCoil(
    final typ=Buildings.Templates.Components.Types.Coil.ElectricHeating,
``` 
At the top-level, we use replaceable components with an explicit choices annotation, for example:
```
replaceable Buildings.Templates.Components.Coils.WaterBasedHeating coiHeaPre(...)
    constrainedby Buildings.Templates.Components.Interfaces.PartialCoil(...)
    "Heating coil in preheat position"
    annotation (
    choices(
      choice(
        redeclare replaceable Buildings.Templates.Components.Coils.None coiHeaPre
        "No coil"),
      choice(
        redeclare replaceable Buildings.Templates.Components.Coils.WaterBasedHeating coiHeaPre(
          redeclare final package MediumHeaWat=MediumHeaWat,
          redeclare final Buildings.Templates.Components.Valves.TwoWayModulating val)
        "Hot water coil with two-way valve"),
      choice(
        redeclare replaceable Buildings.Templates.Components.Coils.ElectricHeating coiHeaPre
        "Modulating electric heating coil")),
```
Having enumerations that differ between the packages Controls and Templates is not an issue. (Of course, if we can use the same, this is better, but this is not a requirement.)

JayHuLBL:
@AntoineGautier I just changed the one enumeration to two enumerations for heating and cooling coil.

AntoineGautier:
I updated the templates in my last commit.

However, the model `Buildings.Controls.OBC.ASHRAE.G36.AHUs.MultiZone.VAV.Validation.Controller` fails to compile with the following class modification.
```
conAHU(heaCoi=Buildings.Controls.OBC.ASHRAE.G36.Types.HeatingCoil.Electric)
```

The model `Buildings.Controls.OBC.ASHRAE.G36.TerminalUnits.Reheat.Controller` enables selecting `heaCoi=Buildings.Controls.OBC.ASHRAE.G36.Types.HeatingCoil.None`. This should not be permitted as then it is a cooling-only box that is modeled with `Buildings.Controls.OBC.ASHRAE.G36.TerminalUnits.CoolingOnly`. 
If `choices` annotations are allowed in CDL, you can restrict the choices exposed in the parameter dialog by adding such annotation to the declaration of `parameter Buildings.Controls.OBC.ASHRAE.G36.Types.HeatingCoil heaCoi`. As this is only a UI feature, an assert statement guarding against `heaCoi==Buildings.Controls.OBC.ASHRAE.G36.Types.HeatingCoil.None` is also needed.

AntoineGautier (approved):

AntoineGautier:
@JayHuLBL This is ready to merge I believe.

JayHuLBL:
@mwetter Please let us know if it is ready to merge.

JayHuLBL:
@mwetter The tests have passed and it is ready for your review.

PR #3532, opened 2023-09-21
Issue3484 dat generation update

### LINKED ISSUE #3484: Python script update for DAT file generation
Issue to track updates to Python script for generating DAT files, that creates new argument for start time.

Current version of script only allows user to define end time, which results in very large DAT files if it needs to cover time periods in later part of simulation year.

JayHuLBL:
This closes #3484.


Buildings/Resources/Scripts/EnergyPlus/energyplus_csv_to_mos.py
```diff
@@ -1,13 +1,14 @@
 #!/usr/bin/env python3
 
-def energyplus_csv_to_mos(output_list, dat_file_name, step_size, final_time):
+def energyplus_csv_to_mos(output_list, dat_file_name, step_size, final_time, start_time:int = 0):
```
mwetter, line 3: Is there a reason you use typing for one argument (`start_time:int`) but don't use typing anywhere else in the whole file? Please use consistent style in your code.

mwetter (changes requested):

JayHuLBL:
@mwetter Can it be merged?

PR #3554, opened 2023-10-09
Issue3538 air2 air recovery

### LINKED ISSUE #3538: Air-to-Air Heat recovery
This issue is to add a new component model to represent sensible and latent air-to-air heat exchangers which are commonly used for the exhaust or relief air heat recovery.
The EnergyPlus component <code>HeatExchanger:AirToAir:SensibleAndLatent</code> will be leveraged for the development.

JayHuLBL:
This closes #3538.


mwetter:
@JayHuLBL : What are the next steps, and who needs to act on them?

JayHuLBL:
@mwetter The PR will be ready for your review when CI test passed.

One questions however is about the base class `Buildings.Fluid.HeatExchangers.AirToAirHeatRecovery.BaseClasses.HeatExchagerWithInputEffectiveness`. The class is similar as `Buildings.Fluid.MassExchangers.ConstantEffectiveness`, with the difference that the effectiveness `espS` and `espL` are inputs rather than parameter. Is it a good choice to duplicate the class and change the effectiveness to be inputs (as implemented here.)?
Or making the changes in `Buildings.Fluid.MassExchangers.ConstantEffectiveness` so that the effectiveness could be conditional inputs. But in this case, the class name `ConstantEffectiveness` will need to be changed.

mwetter:
I think `HeatExchagerWithInputEffectiveness` is fine as I don't see an elegant way to change the effectiveness to be a conditional input, and remove the parameter if the input is enabled.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,165 @@
...
+    "Partial load (75%) latent heat exchanger effectiveness at the cooling mode";
+  parameter Modelica.Units.SI.Efficiency epsS_heat_nominal(max=1) = 0.8
+    "Nominal sensible heat exchanger effectiveness at the heating mode";
+  parameter Modelica.Units.SI.Efficiency epsL_heat_nominal(max=1) = 0.8
+    "Nominal latent heat exchanger effectiveness at the heating mode";
+  parameter Modelica.Units.SI.Efficiency epsS_heat_partload(max=1) = 0.75
+    "Partial load (75%) sensible heat exchanger effectiveness at the heating mode";
+  parameter Modelica.Units.SI.Efficiency epsL_heat_partload(max=1) = 0.75
+    "Partial load (75%) latent heat exchanger effectiveness at the heating mode";
+  parameter Modelica.Units.SI.VolumeFlowRate v_flow_sup_nominal(min = 100*Modelica.Constants.eps)
```
mwetter, line 22: In our naming convention, and in all the control implementations, we call this `VSup_flow_nominal`. Please check and change variable  names for consistency.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,165 @@
...
+      vRat-0.75)/0.25);
+    epsL :=y*(epsL_cool_partload + (epsL_cool_nominal - epsL_cool_partload)*(
+      vRat-0.75)/0.25);
+  else
+    // heating mode
+    epsS :=y*(epsS_heat_partload + (epsS_heat_nominal - epsS_heat_partload)*(
+      vRat-0.75)/0.25);
+    epsL :=y*(epsL_heat_partload + (epsL_heat_nominal - epsL_heat_partload)*(
+      vRat-0.75)/0.25);
+  end if;
```
mwetter, line 70: This must be differentiable.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,165 @@
...
+equation
+  // check if the extrapolation goes too far
+  assert(v_flow_Sup - 2*v_flow_Exh < 0 or v_flow_Exh - 2*v_flow_Sup < 0,
+    "Unbalanced air flow ratio",
+    level=AssertionLevel.warning);
+  // calculate the average volumetric air flow and flow rate ratio
+  vRat =  (v_flow_Sup + v_flow_Exh)/2/v_flow_sup_nominal;
+  // check if the extrapolation goes too far
+  assert(vRat > 0.5 and vRat < 1.3,
+    "Operatiing flow rate outside full accuracy range",
```
mwetter, line 98: Check code for typos before review, and end sentences with a period.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,165 @@
...
+</p>
+<p>
+It first calculates the average volumetric air flow rate through the heat exchanger by:
+</p>
+<pre>
+  v_ave = (v_sup + v_exh)/2,
+  vRat = v_ave/v_sup_nom,
+</pre>
+<p>
+where <code>v_ave</code> is the average volumetric air flow rate,
```
mwetter, line 120: See how we otherwise refer to volume or mass flow rates. (Moreover, if you were to use mass flow rate, then the sensor would be simpler, and the result would in my opinion not change much if at all).

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,165 @@
...
+<code>epsL_75</code> are parameters.
+Depending on the cooling or heating mode, their values are different.
+In this model, if the supply air temperature is larger than the exhaust air
+temperature, the exchanger is considered to operate under the cooling mode;
+Otherwise, it is considered to operate under a heating mode.
+</p>
+<P>
+<b>Note:</b> The average volumetric air flow rate should be between 50% and 130% of the nominal supply air flow rate.
+In addition, the ratio of the supply air flow rate to the exhaust air flow rate should be between 0.5 and 2.
+</P>
```
mwetter, line 153: Explain why it should be that way and what happens if it is not in that range.

Moreover, as fans are typically turned down to 10% to 30% of the design flow rate, I would think this conditions is very often violated. Is there a reason we need it? How would the model be used in typical HVAC systems that have variable speed fans?

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/HeatExchagerWithInputEffectiveness.mo
```diff
@@ -0,0 +1,185 @@
...
+equation
+  // Definitions for effectiveness model
+  X_w_in1 = Modelica.Fluid.Utilities.regStep(m1_flow,
+                  state_a1_inflow.X[i1_w],
+                  state_b1_inflow.X[i1_w], m1_flow_small);
+  X_w_in2 = Modelica.Fluid.Utilities.regStep(m2_flow,
+                  state_a2_inflow.X[i2_w],
+                  state_b2_inflow.X[i2_w], m2_flow_small);
+
+  // mass exchange
```
mwetter, line 73: Review capitalization and punctuation.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/HeatExchagerWithInputEffectiveness.mo
```diff
@@ -0,0 +1,185 @@
...
+January 28, 2010, by Michael Wetter:<br/>
+Added regularization near zero flow.
+</li>
+<li>
+October 21, 2008, by Michael Wetter:<br/>
+First implementation, based on
+<a href=\"modelica://Buildings.Fluid.HeatExchangers.ConstantEffectiveness\">
+Buildings.Fluid.HeatExchangers.ConstantEffectiveness</a>.
+</li>
+</ul>
```
mwetter, line 183: That is a new model which did not start in 2008. PLEASE review your code before it is sent for review.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Validation/HeatExchagerWithInputEffectiveness.mo
mwetter, line 1: First, don't put components right on the edge as they may not be properly displayed or exported.
Second, no need to turn around `epsS` and then cross lines. Arrange your components to avoid clutter.
Third,there is no need for `PSin_1`, just to set a constant pressure, and the same for `POut`.

Let's review the models, make sure they are not full of clutter and lines that don't need to cross and then submit for review.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/Examples/BypassDamper.mo
```diff
@@ -0,0 +1,153 @@
...
+    startTime=240,
+    height=100,
+    offset=1E5 - 110)
+    "Pressure of the supply air"
+    annotation (Placement(transformation(extent={{40,60},{60,80}})));
+  Buildings.Fluid.HeatExchangers.AirToAirHeatRecovery.Wheel whe(
+    redeclare package Medium1 = Medium1,
+    redeclare package Medium2 = Medium2,
+    m1_flow(start=5),
+    m2_flow(start=5),
```
mwetter, line 71: Why are there start attribute set?

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/Examples/BypassDamper.mo
```diff
@@ -0,0 +1,153 @@
...
+    m1_flow_nominal=5,
+    m2_flow_nominal=5,
+    dp1_nominal=100,
+    dp2_nominal=100,
```
mwetter, line 75: Is that realistic for a wheel? I think more common are around 300 Pa.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/Examples/BypassDamper.mo
```diff
@@ -0,0 +1,153 @@
...
+    show_T=true,
+    conTyp=Buildings.Fluid.HeatExchangers.AirToAirHeatRecovery.Types.RecoveryControl.BypassAir,
+    epsL_cool_nominal=0.7,
+    epsL_cool_partload=0.6,
+    epsL_heat_nominal=0.7,
+    epsL_heat_partload=0.6) "Wheel"
+    annotation (Placement(transformation(extent={{6,-4},{26,16}})));
+  Modelica.Blocks.Sources.Ramp DamPos(
```
mwetter, line 83: Naming convention.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/Examples/BypassDamper.mo
```diff
@@ -0,0 +1,153 @@
...
+    height=1.0,
+    duration=60,
+    offset=0,
+    startTime=60) "Damper position"
+    annotation (Placement(transformation(extent={{-80,-40},{-60,-20}})));
+  inner Modelica.Fluid.System system
+    "Ambient environment"
```
mwetter, line 90: Why is this needed?

mwetter (changes requested):
I did a quick review and it is clear that this model is not yet ready for a final review. Please see comments and review first the code; afterwards I am happy to do a last review.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,156 @@
...
+  // switch between cooling and heating modes based on the difference between the supply air temperature and the exhaust air temperature.
+  epsSenPL = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsSenCooPL, epsSenHeaPL, 1e-5);
+  epsSen_nominal = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsSenCoo_nominal, epsSenHea_nominal, 1e-5);
+  epsLatPL = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsLatCooPL, epsLatHeaPL, 1e-5);
+  epsLat_nominal = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsLatCoo_nominal, epsLatHea_nominal, 1e-5);
+  // calculate effectiveness.
+  epsSen =wheSpe*(epsSenPL + (epsSen_nominal - epsSenPL)*(rat - 0.75)/0.25);
+  epsLat =wheSpe*(epsLatPL + (epsLat_nominal - epsLatPL)*(rat - 0.75)/0.25);
+  assert(epsSen > 0 and epsSen < 1,
+    "Insensed value for the sensible heat exchanger effectiveness",
```
JayHuLBL, line 85: what does the `Insensed` mean? Can you improve this message like `*** Error: abcde`. In the message, you may add information of how to avoid the error.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,155 @@
...
+    final displayUnit="degC")
+    "Exhaust air temperature
+    " annotation (Placement(transformation(extent={{-140,-100},{-100,-60}})));
+  Modelica.Blocks.Interfaces.RealInput VSup_flow(final unit="m3/s")
+    "Supply air volumetric flow rate"
+    annotation (Placement(transformation(extent={{-140,60},{-100,100}})));
+  Modelica.Blocks.Interfaces.RealInput VExh_flow( final unit="m3/s")
+    "Exhaust air volumetric flow rate"
+    annotation (Placement(transformation(extent={{-140,20},{-100,60}})));
+  Modelica.Blocks.Interfaces.RealInput uSpe(final unit="1") "Wheel speed"
```
JayHuLBL, line 43: Should be wheel speed have contributes (`min=0, max=1`)?

JayHuLBL (commented):
@SenHuang19 Please see the inline comments.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
SenHuang19: Thanks for your comments, the messages are updated.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
SenHuang19: I added <code>max=1</code>. However, since there may be minimum speed ratio requirement, I didn't add <code>min=1</code>.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/WheelWithBypassDamper.mo
```diff
@@ -0,0 +1,306 @@
...
+  Modelica.Blocks.Sources.RealExpression TExh(
+    final y=Medium2.temperature(
+      Medium2.setState_phX(
+        p=port_a2.p,
+        h=inStream(port_a2.h_outflow),
+        X=inStream(port_a2.Xi_outflow))))
+    "Exhaust air temperature"
+    annotation (Placement(transformation(extent={{-160,-50},{-140,-30}})));
+  Modelica.Blocks.Sources.RealExpression PEle(
+    final y=P_nominal)
```
JayHuLBL, line 132: Is it correct? With this calculation, the power consumption will be constant and always equal to the nominal value.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/WheelWithVariableSpeed.mo
```diff
@@ -0,0 +1,244 @@
...
+  Modelica.Blocks.Sources.RealExpression TExh(
+    final y=Medium2.temperature(
+      Medium2.setState_phX(
+        p=port_a2.p,
+        h=inStream(port_a2.h_outflow),
+        X=inStream(port_a2.Xi_outflow))))
+    "Exhaust air temperature"
+    annotation (Placement(transformation(extent={{-90,10},{-70,30}})));
+  Modelica.Blocks.Sources.RealExpression PEle(
+    final y=P_nominal)
```
JayHuLBL, line 104: Is it correct? With this calculation, the power consumption will always equal to the nominal value.

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/Examples/BypassDamper.mo
JayHuLBL: @SenHuang19 In class `Examples.WheelWithBypassDamper` and `Examples.WheelWithVariableSpeed`, the `dp1_nominal` and `dp2_nominal` are still set to be 100. Is it typical?

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/WheelWithBypassDamper.mo
```diff
@@ -0,0 +1,347 @@
...
+    "Electric power consumption"
+    annotation (Placement(transformation(extent={{100,-10},{120,10}})));
+  Buildings.Fluid.HeatExchangers.AirToAirHeatRecovery.BaseClasses.HeatExchangerWithInputEffectiveness
+    hex(
+    redeclare package Medium1 = Medium1,
+    redeclare package Medium2 = Medium2,
+    final m1_flow_nominal=m1_flow_nominal,
+    final m2_flow_nominal=m2_flow_nominal,
+    final show_T=true,
+    final dp1_nominal=0,
```
JayHuLBL, line 67: @SenHuang19 The `dp1_nominal` and `dp2_nominal` are set to be 0 and the pressure drops are set to be on the by pass dampers (and they are 125 Pa by default.)
In the `WheelWithVariableSpeed`, the `dp1_nominal` and `dp2_nominal` are set to be 125 Pa directly on the heat exchanger. Why should we have this difference?

SenHuang19: 125 Pa is a typical pressure drop of a wheel based on [ref](https://c03.apogee.net/mvc/home/hes/land/el?utilityname=duquesnelight&spc=UD&id=1116#:~:text=Heat%20wheels%20have%20a%20relatively,be%20reduced%20in%20some%20cases.) and I set it on the by pass dampers to provide a high damper authority ratio, which can help to achieve a good control performance.
In the <code>WheelWithVariableSpeed</code>, a similar need doesn't exist so I just set the pressure drop to the wheel directly for simplification.
Please let me know if any of them need to be modified.

JayHuLBL:

mwetter:
@JayHuLBL : I substantially refactored the model to avoid code duplication, and to prepare for adding sensible only wheels, e.g., heat recovery instead of enthalpy recovery wheels. I also made the following changes
- corrected dp_nominal, as 100 Pa is too low. This is now 500 Pa.
- Set pressure drop of wheel, it was ignoring `dp{1,2}_nominal`
- Removed some `noEvent` in assertions.

We should also create a package `Sensible` that has models only configurable for sensible wheels. In these models, the latent calculations need to be avoided (for efficiency) which will require refactoring `Effectiveness.mo`.

See also some inline comment for further changes.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,163 @@
+within Buildings.Fluid.HeatExchangers.ThermalWheels.Latent.BaseClasses;
+model Effectiveness
+  "Model for calculating the heat exchange effectiveness"
+  extends Modelica.Blocks.Icons.Block;
+  parameter Modelica.Units.SI.Efficiency epsSenCoo_nominal(final max=1)
+    "Nominal sensible heat exchanger effectiveness at the cooling mode";
+  parameter Modelica.Units.SI.Efficiency epsLatCoo_nominal(final max=1)
+    "Nominal latent heat exchanger effectiveness at the cooling mode";
+  parameter Modelica.Units.SI.Efficiency epsSenCooPL(final max=1)
+    "Part load (75%) sensible heat exchanger effectiveness at the cooling mode";
```
mwetter, line 10: It is nowhere explained what the 75% relate too. Mass flow rate, or rotational speed?

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,163 @@
...
+  Real epsSen_nominal
+    "Nominal sensible heat exchanger effectiveness used for calculation";
+  Real epsLatPL
+    "Part load (75%) latent heat exchanger effectiveness used for calculation";
+  Real epsLat_nominal
+    "Nominal latent heat exchanger effectiveness used for calculation";
+
+equation
+  // Check if the air flows are too unbalanced
+  assert(noEvent(VSup_flow - 2*VExh_flow < 0) or noEvent(VExh_flow - 2*VSup_flow < 0),
```
mwetter, line 70: Remove the `noEvent` in all assertions.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,163 @@
...
+    "*** Warning in " + getInstanceName() + ": The ratio of the supply flow rate to the exhaust flow rate should be in the range of [0.5, 2].",
+    level=AssertionLevel.warning);
+  // Calculate the average volumetric air flow and flow rate ratio.
+  rat = (VSup_flow + VExh_flow)/2/VSup_flow_nominal;
+  // Check if the extrapolation goes too far
+  assert(noEvent(rat > 0.5) and noEvent(rat < 1.3),
+    "*** Warning in " + getInstanceName() + ": The ratio of the operating flow rate to the nominal supply flow rate should be in the range of [0.5, 1.3].",
```
mwetter, line 77: We never write `*** Warning in`, just use `In `. The warning is written by the assertion.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,163 @@
...
+  epsSenPL = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsSenCooPL, epsSenHeaPL, 1e-5);
+  epsSen_nominal = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsSenCoo_nominal, epsSenHea_nominal, 1e-5);
+  epsLatPL = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsLatCooPL, epsLatHeaPL, 1e-5);
+  epsLat_nominal = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsLatCoo_nominal, epsLatHea_nominal, 1e-5);
+  // Calculate effectiveness
+  epsSen =uSpe*(epsSenPL + (epsSen_nominal - epsSenPL)*(rat - 0.75)/0.25);
+  epsLat =uSpe*(epsLatPL + (epsLat_nominal - epsLatPL)*(rat - 0.75)/0.25);
+  assert(noEvent(epsSen >= 0) and noEvent(epsSen < 1),
+    "In " + getInstanceName() + ": The sensible heat exchange effectiveness epsSen = " + String(epsSen) + ". It should be in the range of [0, 1].
+    Check if the part load (75%) or nominal sensible heat exchanger effectiveness is too high or too low.",
```
mwetter, line 89: It is not clear what the 75% relate to.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/PartialWheel.mo
```diff
@@ -0,0 +1,231 @@
...
+  parameter Modelica.Units.SI.Efficiency epsLatHeaPL(
+    final max=1) = 0.75
+    "Part load (75%) latent heat exchanger effectiveness at the heating mode"
+    annotation (Dialog(group="Part load effectiveness"));
+
+  Modelica.Blocks.Interfaces.RealOutput P(
+    final unit="W")
+    "Electric power consumption"
+    annotation (Placement(transformation(extent={{100,-10},{120,10}})));
+  Buildings.Fluid.HeatExchangers.ThermalWheels.Latent.BaseClasses.HeatExchangerWithInputEffectiveness hex(
```
mwetter, line 64: These should be protected -- no need to expose all these result variables.
But then, at the top-level, add a variable `epsLat` and `epsSen` to see what the actual efficiency is.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/PartialWheel.mo
```diff
@@ -0,0 +1,231 @@
...
+    redeclare package Medium1 = Medium1,
+    redeclare package Medium2 = Medium2,
+    final m1_flow_nominal=m1_flow_nominal,
+    final m2_flow_nominal=m2_flow_nominal,
+    final show_T=true,
```
mwetter, line 69: Why would you use `show_T=true` at such low a level?

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/PartialWheel.mo
```diff
@@ -0,0 +1,231 @@
...
+  Buildings.Fluid.HeatExchangers.ThermalWheels.Latent.BaseClasses.Effectiveness effCal(
+    final epsSenCoo_nominal=epsSenCoo_nominal,
+    final epsLatCoo_nominal=epsLatCoo_nominal,
+    final epsSenCooPL=epsSenCooPL,
+    final epsLatCooPL=epsLatCooPL,
+    final epsSenHea_nominal=epsSenHea_nominal,
+    final epsLatHea_nominal=epsLatHea_nominal,
+    final epsSenHeaPL=epsSenHeaPL,
+    final epsLatHeaPL=epsLatHeaPL,
+    final VSup_flow_nominal=m1_flow_nominal/1.293)
```
mwetter, line 83: Compute the density from the medium properties, as a parameter, and use that value, as we also do for `cp_default`

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/PartialWheel.mo
```diff
@@ -0,0 +1,231 @@
...
+  Modelica.Blocks.Sources.RealExpression VSup_flow(
+    final y=hex.port_a1.m_flow/
+        Medium1.density(state=Medium1.setState_phX(
+        p=hex.port_a1.p,
+        h=hex.port_a1.h_outflow,
+        X=hex.port_a1.Xi_outflow)))
+    "Supply air volume flow rate"
+    annotation (Placement(transformation(extent={{-160,30},{-140,50}})));
+  Modelica.Blocks.Sources.RealExpression VExh_flow(
+    final y=hex.port_a2.m_flow/
```
mwetter, line 130: Add units to these expressions.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BypassDampers.mo
```diff
@@ -0,0 +1,167 @@
...
+          lineColor={28,108,200},
+          fillColor={255,255,255},
+          fillPattern=FillPattern.None),
+        Line(points={{-58,64},{-58,96},{64,96},{64,64}}, color={28,108,200}),
+        Line(points={{-58,-64},{-58,-96},{64,-96},{64,-64}}, color={28,108,200})}),
+          Diagram(
+        coordinateSystem(preserveAspectRatio=true, extent={{-180,-100},{100,180}})),
+Documentation(info="<html>
+<p>
+Model of a generic, sensible and latent air-to-air heat recovery wheel, which consists of
```
mwetter, line 122: What does `generic` mean?

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BypassDampers.mo
```diff
@@ -0,0 +1,167 @@
...
+  extends Buildings.Fluid.HeatExchangers.ThermalWheels.Latent.BaseClasses.PartialWheel;
+  Modelica.Blocks.Interfaces.RealInput uBypDamPos(
+    final unit="1",
+    final min=0,
+    final max=1)
+    "Bypass damper position"
+    annotation (Placement(transformation(extent={{-222,120},{-182,160}}),
+      iconTransformation(extent={{-140,-20},{-100,20}})));
+  Modelica.Blocks.Interfaces.BooleanInput opeSig
+    "True when the wheel is operating"
```
mwetter, line 13: Rename this to `uRot` so it is clear it is a control input for the rotor.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BypassDampers.mo
```diff
@@ -0,0 +1,167 @@
...
+a heat exchanger and two dampers to bypass the supply and exhaust airflow. 
+</p>
+<p>
+This model does not require geometric data. The performance is defined by specifying the
+part load (75%) and nominal sensible and latent effectiveness in both heating and cooling conditions.
+This operation of the wheel is configured as follows.
+</p>
+<ul>
+<li>
+If the operating signal <code>opeSig=true</code>,
```
mwetter, line 132: Should be changed to `uRot`

Buildings/package.mo
```diff
@@ -229,6 +229,17 @@ The following <b style=\"color:blue\">new components</b> have been added
 to <b style=\"color:blue\">existing</b> libraries:
 </p>
 <table class=\"releaseTable\" summary=\"summary\" border=\"1\" cellspacing=\"0\" cellpadding=\"2\" style=\"border-collapse:collapse;\">
+
+<tr><td colspan=\"2\"><b>Buildings.Fluid.HeatExchangers</b>
+    </td>
+</tr>
+<tr><td valign=\"top\">Buildings.Fluid.HeatExchangers.ThermalWheels.Latent
+    </td>
+    <td valign=\"top\">Package of models for air-to-air heat recovery devices.<br/>
```
mwetter, line 238: Change it to thermal wheels, as a plate heat exchanger is also an air-to-air heat recovery device.

mwetter (changes requested):
See inline comments.

mwetter:
Todo
- [ ] Add package `Sensible` for sensible only wheels.

SenHuang19:
> Todo
> 
> * [ ]  Add package `Sensible` for sensible only wheels.

@mwetter thank you so much for your modification and comments. I will work on this. 
One quick question: should we make <code>sensibleWheel</code> and <code>enthalpyWheel</code> as enum so that the two types of wheels can share <code>BypassDampers</code> and <code>SpeedControlled</code>

mwetter:
I would make a separate package called `Sensible` (similar to the new package `Latent`) as this way, users see right away what kind of rotary heat exchangers are available.

JayHuLBL:

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/Effectiveness.mo
```diff
@@ -0,0 +1,163 @@
...
+  rat = (VSup_flow + VExh_flow)/2/VSup_flow_nominal;
+  // Check if the extrapolation goes too far
+  assert(rat > 0.5 and rat < 1.3,
+    "In " + getInstanceName() + ": The ratio of the operating flow rate to the nominal supply flow rate should be in the range of [0.5, 1.3].",
+    level=AssertionLevel.warning);
+  // Switch between cooling and heating modes based on the difference between the supply air temperature and the exhaust air temperature
+  epsSenPL = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsSenCooPL, epsSenHeaPL, 1e-5);
+  epsSen_nominal = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsSenCoo_nominal, epsSenHea_nominal, 1e-5);
+  epsLatPL = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsLatCooPL, epsLatHeaPL, 1e-5);
+  epsLat_nominal = Buildings.Utilities.Math.Functions.regStep(TSup-TExh, epsLatCoo_nominal, epsLatHea_nominal, 1e-5);
```
mwetter, line 83: These 4 calls to `regStep` should be changed: There should only be one call to `regStep` as the first argument is always the same. 
In many configurations, the parameters for the efficiency are the same, in which case no `regStep` (or weighting of the two values) is needed.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/SpeedControlled.mo
```diff
@@ -0,0 +1,89 @@
+within Buildings.Fluid.HeatExchangers.ThermalWheels.Latent;
+model SpeedControlled
+  "Enthalpy recovery wheel with a variable speed drive"
+  extends
+    Buildings.Fluid.HeatExchangers.ThermalWheels.Latent.BaseClasses.PartialWheel;
+  parameter Real a[:] = {1}
+    "Coefficients for power consumption curve for rotor. The sum of the elements must be equal to 1"
+    annotation (Dialog(group="Efficiency"));
```
mwetter, line 8: Consider adding a table instead of using a polynomial, and then use cubic spline.
This would avoid users having to do a polynomial fit.
Perhaps default motor curves could be proposed as are for fans.

JayHuLBL:
@SenHuang19 Would you please address the comments?

Buildings/Fluid/HeatExchangers/AirToAirHeatRecovery/BaseClasses/Effectiveness.mo
mwetter: I removed the assertion as it would trigger every time the fan switches off.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/Effectiveness.mo
mwetter: I changed these calls.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BaseClasses/PartialWheel.mo
```diff
@@ -0,0 +1,261 @@
...
+    final y(final unit="m3/s")=hex.port_a1.m_flow/
+        Medium1.density(state=Medium1.setState_phX(
+        p=hex.port_a1.p,
+        h=hex.port_a1.h_outflow,
+        X=hex.port_a1.Xi_outflow)))
+    "Supply air volume flow rate"
+    annotation (Placement(transformation(extent={{-160,30},{-140,50}})));
+  Modelica.Blocks.Sources.RealExpression VExh_flow(
+    final y(final unit="m3/s")=hex.port_a2.m_flow/
+        Medium2.density(state=Medium2.setState_phX(
```
mwetter, line 129: All efficiency calculations must be based on mass flow rate.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BypassDampers.mo
```diff
@@ -0,0 +1,208 @@
...
+    redeclare package Medium = Medium1,
+    final m_flow_nominal=m1_flow_nominal,
+    final dpDamper_nominal=dp1_nominal)
+    "Supply air damper"
+    annotation (Placement(transformation(
+        extent={{-10,-10},{10,10}},rotation=0,origin={-50,40})));
+  Buildings.Fluid.Actuators.Dampers.Exponential damExh(
+    redeclare package Medium = Medium2,
+    final m_flow_nominal=m2_flow_nominal,
+    final dpDamper_nominal=dp2_nominal)
```
mwetter, line 33: It is unreasonable -- and wrong -- to set the damper pressure drops to the same value as the heat exchanger pressure drop, which is by default 500 Pa.

mwetter (changes requested):

SenHuang19:
> @SenHuang19 Would you please address the comments?

Sure, will do.

SenHuang19:
- [x] Update the revision history for https://github.com/lbl-srg/modelica-buildings/pull/3554#discussion_r1373884726
- [x] Specify the definition of 75% part load for https://github.com/lbl-srg/modelica-buildings/pull/3554#discussion_r1443589910
- [x] Replace the volumetric flow rates with mass flow rates to for https://github.com/lbl-srg/modelica-buildings/pull/3554#discussion_r1578599965
- [x] Replace the RealExpression blocks with mass flow meters
- [x] Only one medium object is needed
- [x] Wrong pressure values for the damper for https://github.com/lbl-srg/modelica-buildings/pull/3554#discussion_r1578601156
- [x] Change the way to define the wheel energy efficiency for https://github.com/lbl-srg/modelica-buildings/pull/3554#discussion_r1577886929
- [x] Add realistic examples

SenHuang19:
Change the way to define the wheel energy efficiency for https://github.com/lbl-srg/modelica-buildings/pull/3554#discussion_r1577886929

- [ ] Define the user input efficiency so that the power at the part load should be lower than the nominal value
- [ ] Check if the relationship between the speed ratio and the work is actually linear

SenHuang19:
Add realistic examples
- [  ]  Add a control loop to the wheel
- [  ]  Making the mass flow rate of the air through the wheel equals 0 for some times during the simulation period

SenHuang19:
Based on the discussion on 6/12/2024.
- [ ]  Modify the icons of the models to improve the readability
- [ ]  Change the <code>Buildings.Fluid.HeatExchangers.BaseClasses.VariableSpeedWheel</code> into <code>Buildings.Fluid.HeatExchangers.BaseClasses.VariableSpeedThermalWheel</code>
- [ ]  Modify the icons of the models to improve the readability

mwetter:
@SenHuang19 : Note that it should be plural not singular, e.g., `VariableSpeedThermalWheels` as the models are also in the package `ThermalWheels` (plural).

Also, the first and last bullet are the same. What  needs to be corrected is the layout of the connection lines so that they don't lay on top of other lines, and are aligned with the grid.

SenHuang19:
> @SenHuang19 : Note that it should be plural not singular, e.g., `VariableSpeedThermalWheels` as the models are also in the package `ThermalWheels` (plural).
> 
> Also, the first and last bullet are the same. What needs to be corrected is the layout of the connection lines so that they don't lay on top of other lines, and are aligned with the grid.

@mwetter thank you so much for your clarification. I have updated the task list.

SenHuang19:
Updates are made in #3909, which is ready for Jianjun's review.

Buildings/Fluid/HeatExchangers/ThermalWheels/Latent/BypassDampers.mo
JayHuLBL: @SenHuang19 Has it been addressed?

SenHuang19: I believe so. The damper pressure is now a separate parameter, set to 20 Pa by default.

JayHuLBL:

mwetter (approved):

mwetter (changes requested):
@JayHuLBL : This still needs significant revisions, namely:

- [ ] The examples use `sta_a` and/or `sta_b`, but these are not defined at zero flow rate. No example should use these quantities. (We removed them in other models due to verification issues across tools.)

- [ ] The bigger change due to non-sound physics is as follows: EnergyPlus uses different efficiencies for heating and cooling because EnergyPlus uses a heating and cooling design flow rate. This Modelica model also uses different efficiencies, but it only has one design flow rate. From a heat transfer point of view, **for the sensible wheel** if the design flow rates are the same, there is no physical justification to use different efficiencies. For the **latent wheel**, however, cooling also has latent exchange so the sensible cooling efficiency may be different from the (sensible) heating efficiency. 

  Moreover, I don't understand what a "latent heating efficiency" is. Therefore, `epsLatHea*` needs to be removed (or clearly explained what it is).

- [ ] Another change is that all efficiency parameters should be in a data record. That is, for the wheels with bypass dampers, the thermal efficiencies for the wheel, and for the wheel with variable speed, the thermal efficiencies plus the speed correction and motor characterization need to be in one data record. This will allow building up a `Data` package with different wheel characterizations.

  Currently, the users sees the `Data` package, but these data only have part of the thermal characterization, and only for some wheel (the ones with variable speed). It would be much clearer if all data is in the record as we do for other components.

SenHuang19:
> @JayHuLBL : This still needs significant revisions, namely:
> 
> * [ ]  The examples use `sta_a` and/or `sta_b`, but these are not defined at zero flow rate. No example should use these quantities. (We removed them in other models due to verification issues across tools.)
> * [ ]  The bigger change due to non-sound physics is as follows: EnergyPlus uses different efficiencies for heating and cooling because EnergyPlus uses a heating and cooling design flow rate. This Modelica model also uses different efficiencies, but it only has one design flow rate. From a heat transfer point of view, **for the sensible wheel** if the design flow rates are the same, there is no physical justification to use different efficiencies. For the **latent wheel**, however, cooling also has latent exchange so the sensible cooling efficiency may be different from the (sensible) heating efficiency.
>   Moreover, I don't understand what a "latent heating efficiency" is. Therefore, `epsLatHea*` needs to be removed (or clearly explained what it is).
> * [ ]  Another change is that all efficiency parameters should be in a data record. That is, for the wheels with bypass dampers, the thermal efficiencies for the wheel, and for the wheel with variable speed, the thermal efficiencies plus the speed correction and motor characterization need to be in one data record. This will allow building up a `Data` package with different wheel characterizations.
>   Currently, the users sees the `Data` package, but these data only have part of the thermal characterization, and only for some wheel (the ones with variable speed). It would be much clearer if all data is in the record as we do for other components.

@mwetter Thank you so much for the edits and comments. I’m working on addressing the points above and will submit a PR for @JayHuLBL’s review shortly.

mwetter:
@SenHuang19 : Looking at https://www.drirotors.com/wp-content/uploads/pdfs/HRW-MS-200-Tech.-Specification_1.pdf I think it makes sense to also put a design mass flow rate and pressure drop in the data record. Then the wheel can be fully characterized by a data record.

JayHuLBL:
@SenHuang19 Is it ready for review, or you are still working on it?

SenHuang19:
> @SenHuang19 Is it ready for review, or you are still working on it?

It is ready for review.

SenHuang19:
> @SenHuang19 : Looking at https://www.drirotors.com/wp-content/uploads/pdfs/HRW-MS-200-Tech.-Specification_1.pdf I think it makes sense to also put a design mass flow rate and pressure drop in the data record. Then the wheel can be fully characterized by a data record.

Thank you for the instructions. I have added the design mass flow rate and the design pressure drop to the data record.

mwetter:
@JayHuLBL to review changes.

Buildings/Fluid/HeatExchangers/ThermalWheels/Data/Characteristics/MotorEfficiency.mo
```diff
@@ -0,0 +1,36 @@
...
+  annotation (
+  defaultComponentPrefixes = "parameter",
+  defaultComponentName = "per",
+  Documentation(info="<html>
+<p>
+This model describes wheel speed ratio <code>uSpe</code> versus
+the motor percent full-load efficiency (see
+<a href=\"modelica://Buildings.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve\">
+Buildings.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve</a>).
+It is based on
```
JayHuLBL, line 22: The development is not based on the model. However, you may instruct user to check the `Buildings.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve` of how it is used.

SenHuang19: @JayHuLBL 
Thanks for your comment. What I was trying to convey is that this model is essentially the same as <code>Buildings.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve</code>, except for the parameter definitions.
Based on your feedback, I’m now wondering if this model is actually necessary, as it might cause confusion.
Do you think I should remove it?

JayHuLBL: @SenHuang19 We may keep it. However, I don't think the two models are same. We may not need to mention `Buildings.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve` here.
Also don't need to refer it in `Buildings.Fluid.HeatExchangers.ThermalWheels.Data.Generic`.

JayHuLBL:

mwetter:
@SenHuang19 : This will be merged when all the CI tests pass.

PR #3609, opened 2023-12-15
Issue2884 pid autotuning

SenHuang19:
@JayHuLBL Please take a look at the PR based on your comments.

### DIFF
--- a/Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/Controller.mo
+++ b/Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/Controller.mo
@@ -49,7 +49,7 @@ protected
     final k=-yLow)
     "Lower value for the output"
     annotation (Placement(transformation(extent={{-80,20},{-60,40}})));
-Buildings.Controls.OBC.CDL.Reals.Subtract conErr
+  Buildings.Controls.OBC.CDL.Reals.Subtract conErr
     "Control error (set point - measurement)"
     annotation (Placement(transformation(extent={{40,0},{60,20}})));
   Buildings.Controls.OBC.CDL.Reals.Switch swi1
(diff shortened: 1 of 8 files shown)


Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/Controller.mo
```diff
@@ -164,38 +155,42 @@ equation
...
 </ul>
 </li>
 <li>
-if <code>yErr &lt; -deaBan</code> and <code>trigger</code> is <code>true</code>,
+if <code>yDif &lt; -deaBan</code> and <code>trigger</code> is <code>true</code>,
 then <code>y = yHig</code>, <code>yOn = true</code>,
 </li>
 <li>
-else if <code>yErr &gt; deaBan</code> and <code>trigger</code> is <code>true</code>,
+else if <code>yDif &gt; deaBan</code> and <code>trigger</code> is <code>true</code>,
```
JayHuLBL, line 193: Note that in current implementation, if `yDiff` (the input to instance `hys`) > `deaBan`, then, `hys.y = true`, `yOn=true` and `y=yHig`. This is opposite to the documentation here.
Which one is right?

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/Controller.mo
```diff
@@ -50,23 +50,23 @@ protected
     "Lower value for the output"
     annotation (Placement(transformation(extent={{-80,20},{-60,40}})));
   Buildings.Controls.OBC.CDL.Reals.Subtract conErr
-    "Control error (set point - measurement)"
+    "Control error (measurement - set point)"
```
JayHuLBL, line 53: In `CDL.Reals.PID` and other PID controller variants, the control error is (set point - measurement). See [here](https://github.com/lbl-srg/modelica-buildings/blob/b282060ebdb814b4b1c2ecde566843d50ce05493/Buildings/Controls/OBC/CDL/Reals/PID.mo#L56).
I see that the output `yErr` from this block becomes input to the block `PIDWithAutotuning.SystemIdentification.FirstOrderTimedelayed.ControlProcessModel`, so for this block, is it expecting `set point - measurement` or `measurement - setpoint`?

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/Controller.mo
SenHuang19: Sorry for confusion and actually I confused myself with the signs. I have updated the model doc.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/Controller.mo
SenHuang19: I removed the <code>yErr</code> and define a new variable called <code>yDiff</code>. <code>yDiff</code> = <code>yErr</code> if <code>reverseActing</code> is true; otherwise <code>yDiff = -yErr</code>.

PR #3614, opened 2023-12-18
Improved wetCoil example

### LINKED ISSUE #3607: Control in wetCoil example cannot meet setpoint temperature
In example model `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControl`, it has the result as:
WetCoilCounterFlowPControl

It indicates that the control cannot drive the valve to allow the measured temperature to meet the set point. It is due to the low capacity. To fix it, we can
- reduce the airside nominal flow `m2_flow_nominal`, and
- reduce the supply water temperature specified in the instance `TWat`.

JayHuLBL:
This closes #3607.

- [x] reduced airflow rate and reduced the water temperature,
- [x] updated .mos script so improve the plots,
- [x] updated reference.


JayHuLBL:

PR #3619, opened 2023-12-28
Issue2884 pid autotuning

SenHuang19:


Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -120,7 +128,7 @@ block FirstOrderAMIGO
     "Check if an autotuning is ongoing while a new autotuning request is received"
     annotation (Placement(transformation(extent={{120,-72},{140,-52}})));
   Buildings.Controls.OBC.CDL.Utilities.Assert assMes(
-    message="An autotuning is ongoing and an autotuning request is ignored.")
+    message="*** Warning in " + getInstanceName() + "An autotuning is ongoing and an autotuning request is ignored.")
```
JayHuLBL, line 131: We should not use the Modelica function `getInstanceName()` in OBC package. Please change the warning message as:
`Warning: An autotuning is ongoing and the new autotuning request is ignored.`

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -148,6 +148,13 @@ protected
       else Buildings.Controls.OBC.CDL.Types.SimpleController.PID
     "Type of controller";
 
+initial equation
```
JayHuLBL, line 151: We cannot have the `initial equation` section in OBC package.
Please use the assert block instead.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
```diff
@@ -18,6 +18,14 @@ block FirstOrderAMIGO
   parameter Real r(
     min=100*Buildings.Controls.OBC.CDL.Constants.eps)=1
     "Typical range of control error, used for scaling the control error";
+  parameter Real yHig = 1
+    "Higher value for the relay output";
+  parameter Real yLow = 0.1
+    "Lower value for the relay output";
+  parameter Real deaBan = 0.1
+    "Deadband for holding the output value";
```
JayHuLBL, line 26: Could you please give more instruction of how to set the `yHig`, `yLow`, `deaBan` and `yRef` value? I see in the example `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPControlAutoTuning`, you set different values, especially the `deaBan=1/298.15` and `yRef=0.5`.

Could these parameters be any value, as long as satisfying the assert condition?

Buildings/Fluid/HeatExchangers/Examples/WetCoilCounterFlowPControlAutoTuning.mo
```diff
@@ -104,11 +109,21 @@ model WetCoilCounterFlowPControlAutoTuning
     "Reset signal"
     annotation (Placement(transformation(extent={{-80,160},{-60,180}})));
   Buildings.Controls.OBC.CDL.Logical.Sources.Pulse autTunSig(
```
JayHuLBL, line 111: In a real control scenario, is there any guideline of when should the controller be tuned? When we are saying auto-tuning, it gives the impression that the controller will automatically tuning the index when certain condition is (or is not) satisfied.

This example triggered the tuning process just based on time, which it may start earlier, or later in reality.

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
SenHuang19: fixed

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
SenHuang19: fixed

Buildings/Controls/OBC/Utilities/PIDWithAutotuning/FirstOrderAMIGO.mo
SenHuang19: The reason to change<code>yHig</code>, <code>yLow</code>, and <code>yRef</code> is to avoid <i>log(0)</i> for the following equation:

image

I believe <code>deaBan</code> is more critical for the performance and I added one sentence to provide a general suggestion.
Please do let me know if it is appropriate.

Buildings/Fluid/HeatExchangers/Examples/WetCoilCounterFlowPControlAutoTuning.mo
SenHuang19: This is a very good question. The reference says "the system is at equilibrium at the working point (u0, y0) before
the relay experiment is started" but I am not sure how to define the equilibrium. 
Can you please let me know if you have any suggestion?

For the studied example, if we make the tuning starts at the very beginning, the result don't change much.

Issue #3631, opened 2024-01-09
Remove unused file from DHC package


dhblum:
This issue will remove the unused file from https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Experimental/DHC/EnergyTransferStations/Combined/Subsystems/Validation/HeatPump.mos.

PR #3632, opened 2024-01-09
Issue 3631 remove file

dhblum:
This is for https://github.com/lbl-srg/modelica-buildings/issues/3631.


dhblum:
@AntoineGautier This all that's needed right?

AntoineGautier:
Yes, that's it.

Issue #3604, opened 2023-12-13
refactor Fluid.SolarCollectors


mwetter:
This issue is to refactor the models in `SolarCollectors` as described by @jelgerjansen during the Modelica IBPSA Working Group meeting.

mwetter:
@jelgerjansen: Please fork from the branch `issue3604_solarCollectors`, which copied all existing models to `Buildings.Obsolete`, using the latest master branch. Please make your changes then in `Buildings.Fluid.SolarCollectors` so that we can merge them back.

jelgerjansen:
Hi @mwetter, you can find my fork here: https://github.com/jelgerjansen/modelica-buildings/tree/issue3604_solarCollectors.

I plan first to do some minor changes, comprising:
- [ ] splitting up data records of ASHRAE93 and EN12975 collectors.
- [ ] updating the documentation of the existing models.
- [ ] allowing to directly define the heat capacity of the collector without fluid in the data record, since the Solar Keymark and SPF databases report the heat capacity (replace `mDry` by `C`).

When this is finished, I will create a pull request to check if all models still work. Next, I plan to do the following:
- [ ] Consider moving from using values `B0` and `B1` (which is not provided in the datasheets of SRCC) to calculate the IAM curve to using tabular data (which is provided in most data sheets these days) combined with spline interpolation. An extra added value is the correctness of the IAM calculation (the EnergyPlus documentation states that the formula using `B0` and `B1` is only valid for incidence angles below 60°). The unit tests will indicate whether or not this results in large differences in collector performance.
- [ ] Include the option to define rectangular arrays of collectors consisting of a number of 'rows' (parallel configuration) and 'columns' (serial configuration) and update the code accordingly.

Furthermore, could you please comment on the following questions:
- The EnergyPlus collector model equations (which are used in the Modelica model) were derived for flat-plate collectors only. Furthermore, the model assumes a unidirectional incidence angle modifier (IAM), but an evacuated tube collector has a bi-directional IAM (longitudinal and transversal). Should we remove the data records for concentrating and evacuated tube collectors (and update the documentation accordingly)?
- The heat capacity of the collector is currently calculated by multiplying the dry mass of the collector with the specific heat capacity of copper (385 J/kg/K). Do we want to add an enumeration in the data record to provide either `mDry` or `C` (similar to `ATyp` for defining the area), or do we just replace `mDry` by `C` in the data record and mention the assumption of `mDry*385` in the documentation?

[...]

PR #3639, opened 2024-01-17
Issue3604 solar collectors

jelgerjansen:
This is for (the first part of) #3604

This PR includes the refactor of the solar collector models. The current changes can be summarised as follows:

- A refactor of the data record
  - Split up the existing data record into an ASHRAE93 and EN12975 record, both extending from a data record containing some general parameters (like area)
  - Remove obsolete inputs: `ATyp`, `dT_nominal`, `G_nominal`
  - Include a parameter `CTyp` and `C` which allows the user to calculate the heat capacity of the collector using the dry or total heat capacity (incl. fluid) when this data is available
  - Add a parameter `eta0` for the EN record, because this value might differ from `y_intercept` which was used before (and is used by the ASHRAE standard)
  - Rename parameters to comply with the data records of SRCC, Solar keygen, and SPF: `b0`, `b1`, `a1`, and `a2` instead of `B0`, `B1`, `C1`, and `C2`
  - Clarify the meaning of `IAMDiff` (IAM at 50°)
  - Update existing data records such that they comply with this new data record format (I didn’t update the values yet)
  - As a consequence of splitting up the data, I had to make the data record replaceable in `PartialSolarCollector.mo` and redeclare it in `ASHRAE93.mo` and `EN12975.mo`
- Remove everything related to concentrating solar collectors (model equations were derived for flat plate solar collectors)
- Remove the constraint that sets the minimum number of discretization segments (used to be 3), but left the default equal to 3
- Remove obsolete calculation of `QLos_nominal` and `QUse_nominal`
- Change the calculation of the representative volume based on the choice of `CTyp` in `PartialSolarCollector.mo`
- Remove obsolete parameter declarations and update parameter names in `ASHRAE93.mo` and `EN12975.mo`
- Update the Users Guide (e.g. adding warnings regarding pressure drops depending on the medium used and IAM values for evacuated tube collectors)
- Update the documentation (including references to equations in the EnergyPlus documentation) and revision history for all models (revision history still needs a specific date)
- Update unit tests scripts and to comply with the new variable names
- Update the position of models and connections in the examples, validation and controls package models (more structured).
- Update reference results (no differences, except name of some variables)

Still to be done before merging:

- [ ] Specify date in revision history

Still to be done is this or a future PR:

- Update the data records using the latest data sheets and sources
- Revise the implementation of the incidence angle modifier (currently only valid approximation for angles <60°)
- Allow defining rectangular solar collector fields with a number of collectors in series and parallel


mwetter:
@jelgerjansen : I will close this as the changes are addressed via the two PRs that I just merged, https://github.com/lbl-srg/modelica-buildings/pull/3667 followed by https://github.com/lbl-srg/modelica-buildings/pull/3641

Please use `issue3604_solarCollectors` for further work. (You may want to do a test PR as you start the development to make sure only real changes show up.)

PR #3675, opened 2024-03-01
Updated optimica to 1.48.2

mwetter:
This closes #3674

PR #3677, opened 2024-03-02
Updated optimica to 1.48.2 - revised conf.yml

mwetter:
This includes https://github.com/lbl-srg/modelica-buildings/pull/3675 and enables models for Optimica that were excluded in `conf.yml`.


mwetter:
Closing this PR as it is replaced by https://github.com/lbl-srg/modelica-buildings/pull/3686

PR #3678, opened 2024-03-04
Corrected displayUnit in SkyClearness and G36

mwetter:
This corrects a wrong `displayUnit` attribute.


mwetter:
@AntoineGautier : This fails because of the check-sum for the templates (https://app.travis-ci.com/github/lbl-srg/modelica-buildings/jobs/618818896). 

Based on the documentation in `checkandrun.sh` I run it as follows, but obtained an error:
```bash
Buildings$ ../Buildings/Resources/Scripts/travis/templates/checkandrun.sh --local --tool Optimica
No --checksum flag is passed, so no test is performed. This is considered improper usage.
```
Can you please let me know how to invoke regeneration of the checksums, and update the help in `checkandrun.sh`.

AntoineGautier:
In https://github.com/lbl-srg/modelica-buildings/pull/3678/commits/1da730406b3c635086dc162e68cb7989128d3149 I added a `--skip` option to disable all simulations and simply update the checksums. I also updated the checksums in this commit.

Before that, in order to update the checksums, you also needed to run all simulations locally with the following command:
`./Resources/Scripts/travis/templates/checkandrun.sh --checksum --local [--tool SIMULATOR]`

Now, you can skip simulations (locally) and rely on Travis for that. To simply update the checksums, you can execute:
`./Resources/Scripts/travis/templates/checkandrun.sh --checksum --local --skip`

Note: The option `--checksum` means "use the checksum comparison to trigger testing". It is a mandatory option because it is currently the only method available. It comes from your suggestion to prepare for an alternative method to trigger testing in the future, e.g., periodic testing once every month.

PR #3691, opened 2024-03-08
Issue2884 pid autotuning - revision

mwetter:
This merges revisions for the autotuning controller.

Todo:
- [ ] Revise latest changes
- [ ] Check why `gai1` and `gai2` are used in `WetCoilCounterFlowPControlTuning`, and why it sets `deaBan=1/298.15` which is very small (and absolute temperatures do not matter for a temperature *difference*.
- [ ] Check why some models use `deaBan` of 0.5 while other use 0.1. Why not use a consistent value?
- [ ] Allow control output outside of 0...1 to make controller replaceable with standard PID controller.


mwetter:
@SenHuang19 @JayHuLBL : Do you have any explanation for the 2nd and 3rd item above? Please don't change the code yet as I implemented `r` in adba520

SenHuang19:
> @SenHuang19 @JayHuLBL : Do you have any explanation for the 2nd and 3rd item above? Please don't change the code yet as I implemented `r` in [adba520](https://github.com/lbl-srg/modelica-buildings/commit/adba5202aa0d0a4ca1bbac26c4dcc4fa020df6be)

Item 2 is for normalizing the measurement and the setpoint.
For item 3, <code>deaBan</code> seems to be case dependent, please let me know if we need to add some guidelines for users to find an appropriate value.

mwetter:
@SenHuang19 : What is the rationale for such a normalization as the *range* of the controlled variable has nothing to do with 298.15. The degC scale is just an offset. If the error is of the order of 10 (say) then we should set r=10 in `con`. That is what `con.r` is for. Is there another reason I am missing or this a mistake in the model?

image

Maybe that is the reason for `con.deaBan` to be `1/298.15`. A deadband of 0.003 Kelvin does not make sense.

mwetter:
Regarding `deaBan` I understand that it is case dependent, but is there any rational for building a controller that has sub-blocks, and then be inconsistent in the default value? Users may wonder or believe that there is a rationale for it. Is there?

SenHuang19:
> @SenHuang19 : What is the rationale for such a normalization as the _range_ of the controlled variable has nothing to do with 298.15. The degC scale is just an offset. If the error is of the order of 10 (say) then we should set r=10 in `con`. That is what `con.r` is for. Is there another reason I am missing or this a mistake in the model?
> 
> image
> 
> Maybe that is the reason for `con.deaBan` to be `1/298.15`. A deadband of 0.003 Kelvin does not make sense.

Thanks a lot for the insight!
I didn't notice the existence of <code>r</code>.
I will update the example based on your instruction.

> Regarding `deaBan` I understand that it is case dependent, but is there any rational for building a controller that has sub-blocks, and then be inconsistent in the default value? Users may wonder or believe that there is a rationale for it. Is there?

No, there is no such rationale.
Can I address this issue by removing the default value?

mwetter:
@SenHuang19 : Thanks for the clarification. As I wrote in https://github.com/lbl-srg/modelica-buildings/pull/3691#issuecomment-1985766979 do NOT update the code as I am in the middle of making changes. Otherwise our code will conflict.

PR #3696, opened 2024-03-11
Add Modelica Tough coupling user guide

### LINKED ISSUE #1495: interface integrating Modelica and TOUGH
This task intends to create interface between Modelica and TOUGH.

JayHuLBL:
This closes #1495.


JayHuLBL:
@EttoreZ Would you please take another look at the changes?

EttoreZ:
Hey @JayHuLBL just two quick comments:

- In the doStep description of the UserGuide. I would comment/remove this line: ` os.system("./writeincon < writeincon.inp")` and uncomment this line ` # write_incon()`, and also remove this line `os.system(" ./readsave < readsave.inp > out.txt")` and uncomment this line `# readsave()`. This is because the Python script now does not rely on the Fortran utility programs anymore.
- The second comment is also related to that. In the diagram where you show the workflow of the Python script, `writeincon` and `readsave` are cited as utility programs. I would change that to Python functions.
Apart from those, everything else looks good to me.

JayHuLBL:
@EttoreZ Thanks for the comments. I just addressed it. Please double-check.

EttoreZ:
@JayHuLBL  the changes look good to me.

JayHuLBL:

mwetter (changes requested):
Please do one more revision.

Buildings/Fluid/Geothermal/Borefields/TOUGHResponse/UsersGuide.mo
```diff
@@ -0,0 +1,299 @@
...
+Buildings.Utilities.IO.Python_3_8.Real_Real</a>) is true, Modelica calls the
+TOUGH simulation through the ground response model as described in the above section.
+The Python module invokes heat flow rates from the ground <code>QBor_flow</code>,
+the initial borehole wall temperature <code>TBorWal_start</code>, the ambient
+air temperature <code>TOut</code> and the current simulation time. 
+</p>
+<ol>
+<li>
+In the first invocation of Python, this object is not yet initialized. Python
+therefore takes the initial temperature from the Modelica to initialize the object.
```
mwetter: It is not clear what "this object" refers to. Do you mean the ground temperature? Are they initialized uniformly, or does TOUGH initialize the ground temperature through its input files?

Buildings/Fluid/Geothermal/Borefields/TOUGH/UsersGuide.mo
```diff
@@ -0,0 +1,299 @@
...
+<ul>
+<li>
+<code>writeincon.inp</code>: the file contains the initial borehole wall temperature and 
+the heat flow rate. The initial borehole wall temperature will be updated with the
+borehole wall temperature stored in the state. The heat flow rate is the heat flow rate
+measured in the Modelica model.
+</li>
+<li>
+<code>INFILE</code>: the start and stop TOUGH simulation time will be updated.
+The start time is the one stored in the state. The stop time is the current
```
mwetter: I don't understand that bullet. `INFILE` is a file, but then the description says time will be updated. Explain this clearer.

Buildings/Resources/Python-Sources/TOUGH/readsave
mwetter, line 1: This is a binary file. We need to post the source code and the built file.
Also, this implies that the coupling only works on Linux 64 bit, and there are various dependencies.
Moreover, `readsave` and`writeincon` depend on different versions of `libgfortran` (and I don't think version 4 is available on Ubuntu 24.04.
What are these files doing and why are they only in compiled form?

Buildings/Resources/Images/Fluid/Geothermal/Borefields/TOUGHResponse/workFlow.svg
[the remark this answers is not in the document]
mwetter: @JayHuLBL : this is still not addressed.

Buildings/Fluid/Geothermal/Borefields/TOUGH/BaseClasses/PartialBorefieldWithTough.mo
```diff
@@ -0,0 +1,280 @@
...
+between each pipe and the borehole wall.
+</p>
+<p>
+The thermal interaction between the borehole wall and the surrounding soil
+is modeled using
+<a href=\"modelica://Buildings.Fluid.Geothermal.Borefields.BaseClasses.HeatTransfer.GroundTemperatureResponse\">
+Buildings.Fluid.Geothermal.Borefields.BaseClasses.HeatTransfer.GroundTemperatureResponse</a>,
+which uses a cell-shifting load aggregation technique to calculate the borehole wall
+temperature after calculating and/or read (from a previous calculation) the borefield's
+thermal response factor.
```
mwetter: What does cell-shifting load aggregation and thermal response factor has to do with TOUGH?

mwetter (changes requested):
This is not ready for review.

I started a review but stopped. You need to check it more carefully and revise the writing one more time for clarity.

Also, we can't post binaries with such dependencies that will inadvertently break as Ubuntu is being updated.

Also, earlier comment is still unaddressed.

Buildings/Resources/Python-Sources/TOUGH/readsave
JayHuLBL: The function is now being replaced by the Python function.

Buildings/Fluid/Geothermal/Borefields/TOUGH/BaseClasses/GroundResponse.mo
```diff
@@ -0,0 +1,162 @@
...
+    final k={nSeg,nTouSeg,nInt}) "Total number of grids"
+    annotation (Placement(transformation(extent={{-80,60},{-60,80}})));
+  Buildings.Controls.OBC.CDL.Reals.Sources.Constant borHei(final k=hBor)
+    "Total height of the borehole"
+    annotation (Placement(transformation(extent={{-40,-90},{-20,-70}})));
+
+equation
+  // Delete the TOUGH temporary working folder
+  // Note that the working folder path is specified in the Python function.
+  when {terminal()} then
```
JayHuLBL: The `terminal()` function is not supported by Optimica 1.66.

The function is also being used in [Buildings.DHC.Loads.HotWater.ThermostaticMixingValve](https://github.com/lbl-srg/modelica-buildings/blob/dc18485a811428470bad3288af1371c771173d82/Buildings/DHC/Loads/HotWater/ThermostaticMixingValve.mo#L236). Running the model `Buildings.DHC.Loads.HotWater.Examples.StorageTankWithExternalHeatExchanger` with Optimica 1.66, we can see
```
In component theMixVal:
    The terminal() operator is not supported, and is currently evaluated to false
```

JayHuLBL: The coupled simulation requires creating the temporary folder for TOUGH simulation. If the temporary folder cannot be removed after the simulation in CI test, the test will fail. Thus, the `terminal()` is used in the validation model, for deleting the temp folder at the end of the simulation.

The coupled simulation itself does not require the `terminal()` operator.

JayHuLBL:
@mwetter It will be ready for your review after the CI tests passed.

Buildings/Fluid/Geothermal/Borefields/TOUGH/BaseClasses/Validation/GroundResponse.mo
```diff
@@ -0,0 +1,75 @@
...
+  Buildings.Controls.OBC.CDL.Reals.Sources.Constant outTem(
+    final k=273.15 + 20)
+    "Outdoor temperature"
+    annotation (Placement(transformation(extent={{-60,-60},{-40,-40}})));
+
+equation
+  // Delete the TOUGH temporary working folder
+  // Note that the working folder path is specified in the Python function.
+  when {initial(), terminal()} then
+    Modelica.Utilities.Files.remove("Resources/Python-Sources/tmp-tou-work");
```
mwetter: This needs to be changed as the library folder may be read only if the library is properly installed (such as in `/usr/local`.
Instead, in Tough, get the system dependent tmp folder, and use that instead.

Buildings/Fluid/Geothermal/Borefields/TOUGH/BaseClasses/GroundResponse.mo
```diff
@@ -0,0 +1,156 @@
...
+    final flag=0,
+    final passPythonObject=true)
+    "Python interface model to call TOUGH simulator"
+    annotation (Placement(transformation(extent={{40,-10},{60,10}})));
+  Modelica.Blocks.Routing.Multiplex mul(
+    final n=2*nSeg + 6)
+    "Multiplex"
+    annotation (Placement(transformation(extent={{0,-10},{20,10}})));
+  Modelica.Blocks.Sources.ContinuousClock clock
+    "Current time"
```
mwetter: This won't work if the simulation does not start at t=0. See other models where we use the model time. If  Tough internally needs to start at t=0, then it would be better to put this into the Python code as this block does not output the "Current time".

JayHuLBL:
- [x] added new icons for the modelica-tough coupling class
- [x] enabled openmodelica simulation
- [x] moved the TOUGH files to data folder
- [x] changed the TOUGH working folder to a temp folder
- [x] used model time block

JayHuLBL:

Buildings/Fluid/Geothermal/Borefields/TOUGH/BaseClasses/Validation/GroundResponse.mo
mwetter: This is not correctly addressed: Currently files are written to cwd/tmp-tou-wor where cwd is the current working directory. If there are two borefields, this won’t work as they use the same directory. Instead, use the python mechanism to get a unique name for a temporary directory, and prepend to it tmp-tou-wor. See for example BuildingsPy for how temporary directories are created.

This change also requires you to change the line 
```modelica
Modelica.Utilities.Files.remove(”tmp-tou-wor”);
```

When you change GroundResponse.py, also remove code that is commented out if that code is no longer needed (or if you have a reason for keeping it, add a comment to the file that explains why it is still there).

Buildings/Fluid/Geothermal/Borefields/TOUGH/BaseClasses/PartialBorefield.mo
```diff
@@ -0,0 +1,216 @@
...
+    annotation(Evaluate=true, Dialog(tab = "Dynamics", group="Equations"));
+
+  // Initialization
+  parameter Medium.AbsolutePressure p_start = Medium.p_default
+    "Start value of pressure"
+    annotation(Dialog(tab = "Initialization"));
+
+  // General parameters of borefield
+  parameter Buildings.Fluid.Geothermal.Borefields.Data.Borefield.Template borFieDat "Borefield data"
+    annotation (choicesAllMatching=true,Placement(transformation(extent={{-80,-80},{-60,-60}})));
```
mwetter, line 36: This model uses the instance `borFieDat`, which requires borehole coordinates and soil properties. Neither of them are used in this model. Therefore, this data record should not be used here as users may be confused about having to enter data that end up not being used, and also because TOUGH allows non-uniform soil data which contradicts the soil data entry in borFieDat.
You will need to use a data record that has only those data that are needed by this model

Buildings/Fluid/Geothermal/Borefields/TOUGH/Examples/Borefield.mo
mwetter, line 1: TOUGH.Examples.Borefields need to be renamed to TOUGH.Examples.Borefield as it is only one borefield.

mwetter (changes requested):
The file `Buildings/UsersGuide/package.mo` lists all users’ guides of the library. You also need to add the TOUGH users’ guide to that table.

Please also see other inline comments.

Buildings/Fluid/Geothermal/Borefields/TOUGH/BaseClasses/Validation/GroundResponse.mo
JayHuLBL: By extending the class `Buildings.Utilities.IO.Python_3_12.Real_Real` to the package and add the string parameter for naming TOUGH working folder would not work, as the `Buildings.Utilities.IO.Python_3_12.Real_Real` hard coded `nStrWri=0, strWri={""}` as below:
```Modelica
    yR = Buildings.Utilities.IO.Python_3_12.Functions.exchange(
      moduleName=moduleName,
      functionName=functionName,
      dblWri=uRWri,
      intWri={0},
      nDblWri=nDblWri,
      nDblRea=nDblRea,
      nIntWri=0,
      nIntRea=0,
      nStrWri=0,
      strWri={""},
      pytObj=pytObj,
      passPythonObject = passPythonObject);
    // Store current value of integral
```
It would be necessary to duplicate the `Real_Real`?

Or, just use the Python mechanism to get the unique name for the temporary directory. The temporary directory will be used for the TOUGH simulation sampling moment, and thus it cannot be deleted by the Python code (and also Modelica). Not sure if the not-deleted temporary directory will cause CI test fail.

mwetter: We should delete the temporary directory after the simulations. You may need to copy `Real_Real` to have an implementation that allows sending a string parameter.

JayHuLBL:
- [x] renamed example model
- [x] added user guide link to the table of the UserGuide
- [x] corrected the model to allow sampling at begin of any simulation start time
- [x] added borefield data record
- [x] duplicated the python interface model and propagated the parameters for wring string to python

Buildings/Fluid/Geothermal/Borefields/TOUGH/UsersGuide.mo
```diff
@@ -0,0 +1,340 @@
...
+    # to the TOUGH boundary mesh point
+    Q_toTough = mesh_to_mesh(toughLayers, modelicaLayers, Q_stored, 'Q_Mo2To')
+
+    # Update TOUGH input files for each TOUGH call:
+    #   -- update the INFILE to specify beginning and ending TOUGH simulation time
+    #   -- update the GENER for specifying the heat flow boundary condition
+    write_incon()
+
+    # Conduct one step TOUGH simulation
+    os.system(\"/.../tough3-install/bin/tough3-eos1\")
```
mwetter: There is no such code in the python file. 
Update the user guide to reflect the actual implementation.

mwetter (changes requested):
@JayHuLBL : Once the UsersGuide.mo is revised, this is ready to merge.

mwetter (approved):

mwetter:
There seems to be a race condition. On my Linux computer, if I run omc in the docker with the command
```bash
for i in {1..20}; do echo "=== Test $i"; ../bin/runUnitTests.py -s Buildings.Fluid.Geothermal.Borefields.TOUGH.BaseClasses.Validation -d -b -t openmodelica --skip-verification || break; done
```
and BuildingsPy 5.2.0, the test fails usually in the first 5 evaluations.

If run outside of the docker, the test works for all 20 runs.

Further tests show that the reason is not `BuildingsPy`:
A test failed in `/tmp/tmp-Buildings-0-nmt8avn4`. Next, running, with omc from the docker, the command
```bash
cd /tmp/tmp-Buildings-0-nmt8avn4
$ omc Buildings.Fluid.Geothermal.Borefields.TOUGH.BaseClasses.Validation.GroundResponse_simulate.mos 
Using docker lbnlblum/ubuntu-2404-omc:1.26.3-1
[/mnt/shared/tmp-Buildings-0-nmt8avn4/Buildings.Fluid.Geothermal.Borefields.TOUGH.BaseClasses.Validation.GroundResponse_simulate.mos:1:17-1:230:writable] Warning: Lexer treating \ as \\, since \[ is not a valid Modelica escape sequence.

... [multiple outputs omitted]

[/mnt/shared/tmp-Buildings-0-nmt8avn4/Buildings.Fluid.Geothermal.Borefields.TOUGH.BaseClasses.Validation.GroundResponse_simulate.mos:1:17-1:585:writable] Warning: Lexer treating \ as \\, since \] is not a valid Modelica escape sequence.

LOG_STDOUT        | warning | simulation_input_xml.c: override variable name not found in model: startTime
LOG_STDOUT        | warning | simulation_input_xml.c: override variable name not found in model: stopTime
LOG_STDOUT        | warning | simulation_input_xml.c: override variable name not found in model: stepSize
LOG_STDOUT        | warning | simulation_input_xml.c: override variable name not found in model: tolerance
LOG_STDOUT        | warning | simulation_input_xml.c: override variable name not found in model: solver
LOG_STDOUT        | warning | simulation_input_xml.c: override variable name not found in model: outputFormat
LOG_STDOUT        | warning | simulation_input_xml.c: override variable name not found in model: variableFilter
LOG_SUCCESS       | info    | The initialization finished successfully without homotopy method.

Limited backtrace at point of segmentation fault
/lib/x86_64-linux-gnu/libc.so.6(+0x45330)[0x7613e76c1330]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x326b38)[0x7613e8fabb38]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x1aef3b)[0x7613e8e33f3b]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x32777c)[0x7613e8fac77c]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x329474)[0x7613e8fae474]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x2cf62f)[0x7613e8f5462f]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(_PyEval_EvalFrameDefault+0x43e4)[0x7613e8da28e4]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x182d9e)[0x7613e8e07d9e]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(PyObject_CallMethodObjArgs+0x101)[0x7613e8e07fe1]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x3e7951)[0x7613e906c951]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x3e86ab)[0x7613e906d6ab]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x3e965b)[0x7613e906e65b]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(+0x3e9f3c)[0x7613e906ef3c]
/lib/x86_64-linux-gnu/libpython3.12.so.1.0(_PyEval_EvalFrameDefault+0x4a5d)[0x7613e8da2f5d]
/mnt/shared/tmp-Buildings-0-nmt8avn4/Buildings/Resources/Library/linux64/libModelicaBuildingsPython_3_12.so(pythonExchangeValuesNoModelica+0xdbc)[0x7613e953158d]
Segmentation fault (core dumped)
139
```
However, sometimes, that command finishes without an error, producing
`LOG_SUCCESS       | info    | The simulation finished successfully.`

JayHuLBL:
The diagnostic with LLM is below:
```
Root cause of the intermittent segmentation fault

The crash happens inside the embedded‑CPython wrapper that all `TOUGH`‑based borefield models use to call Python:

`Resources/src/Utilities/IO/Python_3_12/pythonInterpreter.c` → `pythonExchangeValuesNoModelica()`

Two problems in this file combine to create a race condition that only shows up under certain OMC builds (such as the one baked into the `lbnlblum/ubuntu-2404-omc` Docker image):

1. **The Python interpreter is (re‑)initialized on every single call, not just once.**
   Unlike the Python 3.8 version of this file (`Resources/src/Utilities/IO/Python_3_8/pythonInterpreter.c`, which correctly guards initialization with `if (!Py_IsInitialized()) Py_Initialize();`), the Python 3.12 version unconditionally calls:
   ```c
   PyConfig_InitPythonConfig(&config);
   ...
   status = Py_InitializeFromConfig(&config);
   ...
   if (!Py_IsInitialized())      // always true here, so this check is dead code
       Py_Initialize();
   ```
   `Py_InitializeFromConfig()` is executed on *every* invocation of `doStep` (i.e., on every `sampleTrigger` event), even though the interpreter is already running. Repeated re‑initialization with a `PyConfig` (as opposed to the simple, idempotent `Py_Initialize()`) is not something CPython guarantees to be safe, and it manipulates global interpreter/runtime state (including `sys.path`, via the `strtok()` call performed in‑place on the persistently‑stored `ptrMemory->pythonPath` buffer) on every call.

2. **No GIL protection around the CPython C‑API calls.**
   The function calls `PyImport_Import`, `PyObject_GetAttrString`, `PyObject_CallObject`, etc. directly, without ever calling `PyGILState_Ensure()` / `PyGILState_Release()`. This implicitly assumes the function is only ever entered from a single native thread at a time.

The OpenModelica runtime, however, does not guarantee that external functions embedded in `when` clauses / the initialization system are only ever evaluated from one thread. Depending on how the runtime was built (this differs between the OMC binary bundled in the `lbnlblum/ubuntu-2404-omc` Docker image and a natively-installed OMC), the initialization/event system can evaluate the residual equations that contain the external call from more than one thread (e.g., for parallel/finite‑difference Jacobian evaluation during initialization, which is exactly where the log shows the crash occurring — right at/after "The initialization finished successfully"). When two threads enter `pythonExchangeValuesNoModelica` at (nearly) the same time:
- both may try to run `Py_InitializeFromConfig`/mutate `sys.path` concurrently, and/or
- both execute CPython bytecode via `PyObject_CallObject`/`_PyEval_EvalFrameDefault` concurrently without holding the GIL.

CPython's interpreter state and object allocator (`obmalloc`) are not thread‑safe without the GIL, so concurrent, unsynchronized access corrupts interpreter memory. The corruption doesn't crash deterministically — it depends on thread scheduling/timing.
```

JayHuLBL:
@mwetter It's ready to merge.

mwetter:
@JayHuLBL : Please also compile the binary on windows. The it is ready to merge.

mwetter (approved):

PR #3719, opened 2024-03-21
Rename DHC models and update release notes

EttoreZ:
This PR addresses issues #3694 and #3712.

Issue #3606, opened 2023-12-14
Update icons in Buildings.Experimental.DHC.EnergyTransferStations.Combined.Subsystems


dhblum:
This issue is to make the icons of Buildings.Experimental.DHC.EnergyTransferStations.Combined.Subsystems more descriptive of the subsystems they represent.

dhblum:
Closed by https://github.com/lbl-srg/modelica-buildings/pull/3734.

PR #3734, opened 2024-03-27
Issue3606 district icons

dhblum:
This is for #3606.  It adds/updates icons for ``Buildings.Experimental.DHC.EnergyTransferStations``.

It includes changes in https://github.com/lbl-srg/modelica-buildings/pull/3719 and should be merged after that one.

@mwetter This is ready for review, tests should pass.  Should icon updates be added to release notes?  If not, it's also then ready for merge if review and tests pass.  Otherwise, that still needs to be updated.


mwetter:
@dhblum : Thanks. Icon updates don't need to be added to the release notes.

mwetter (changes requested):
Can you check the `Direct` and `Indirect` icons in `Cooling` and `Heating`. It looks like in Dymola lines are not removed:
image

Also, the connector icons for `Q` and `Q_flow` are overlapping.

dhblum:
Thanks for noticing, will correct the Q_flow and Q.

For the lines, it has to do with district system type assignment and boolean checking in the visibility annotation.  Will look into further to correct.

dhblum:
@mwetter Your correction has been addressed in https://github.com/lbl-srg/modelica-buildings/pull/3734/commits/cce112d2f3a20978d6157baf527d950bfb07f760.

dhblum:
Ok but don't merge yet, I found similar issues elsewhere.  Will correct.

dhblum:
Ok made additional corrections in https://github.com/lbl-srg/modelica-buildings/pull/3734/commits/eed7c361794d06aa0fba728693a594a53d04a377.  Ok to merge if review and tests pass.

mwetter (approved):

PR #3755, opened 2024-03-29
Issue3492 direct evaporative cooler

### LINKED ISSUE #3492: Evaporative cooler models for SPAWN
This issue is to track the merging of the following three components:
- [ ] Direct evaporative cooler
- [ ] Indirect dry evaporative cooler
- [ ] Indirect wet evaporative cooler

JayHuLBL:
This closes #3492.


Buildings/Fluid/Humidifiers/EvaporativeCoolers/Baseclasses/DirectCalculations.mo
```diff
@@ -0,0 +1,154 @@
...
+  replaceable package Medium = Modelica.Media.Interfaces.PartialMedium
+    "Medium";
+
+  parameter Modelica.Units.SI.Area padAre
+    "Area of the rigid media evaporative pad";
+
+  parameter Modelica.Units.SI.Length dep
+    "Depth of the rigid media evaporative pad";
+
+  parameter Real effCoe[11] = {0.792714, 0.958569, -0.25193, -1.03215, 0.0262659,
```
JayHuLBL, line 14: Can the user change the `effCoe[:]`? Is it from the E+ reference? If it shall not be changed, make it to be protected.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/Baseclasses/DirectCalculations.mo
```diff
@@ -0,0 +1,154 @@
...
+    final unit="Pa",
+    displayUnit="Pa",
+    final quantity="AbsolutePressure")
+    "Pressure"
+    annotation (Placement(transformation(origin={-120,-60},extent={{-20,-20},{20,20}}),
+      iconTransformation(origin={-120,-60}, extent={{-20,-20},{20,20}})));
+
+  Buildings.Controls.OBC.CDL.Interfaces.RealOutput dmWat_flow(
+    final unit="kg/s",
+    displayUnit="kg/s",
```
JayHuLBL, line 57: remove `displayUnit`

JayHuLBL:
@karthikeyad-pnnl See below comments:

1. Avoid using the temperature sensor to measure the instrem fluid temperature as it will introduce additional states, as below the instance `senTem`, `senTemWetBul`:
Direct
You may use the `instream()` function, see for example in model `Buildings.Fluid.HeatExchangers.DryCoilCounterFlow`.

2. Also, it's better to use the mass flow rate sensor, rather than volume flow rate sensor.

3. Check if we can extent the base class `Buildings.Fluid.Interfaces.TwoPortHeatMassExchanger`, as it's being used in `Buildings.Fluid.Humidifiers.Humidifier_u`.

JayHuLBL:
@karthikeyad-pnnl The reference results have been updated. I wonder if it is still a draft PR? Are you working on any other issues for this PR?

JayHuLBL:
@karthikeyad-pnnl Can this PR become formal PR?

karthikeyad-pnnl:
@JayHuLBL Yes I think it's ready for review and can be changed to formal PR. We are not making any additional updates from our end. Thanks!

JayHuLBL:
@mwetter Would you please let us know if the new model `Buildings.Fluid.Humidifiers.EvaporativeCoolers.Baseclasses.Xi_TDryBulTWetBul` should be in the package `Buildings.Utilities.Psychrometrics`?

mwetter:
@JayHuLBL : Thanks for checking. Yes the block should be moved. And its output `Xi[]` should actually be `X_w` similar than in `Buildings.Utilities.Psychrometrics.X_pW` because the implementation assigns only one element. Hence the block would be invalid if `Medium.nXi > 1`.

JayHuLBL:
@karthikeyad-pnnl This PR looks good to me. Can you double-check if everything looks good to you? Otherwise, we can ask Michael to review.

karthikeyad-pnnl:
@JayHuLBL Thank you! I will get back to you shortly after review

Buildings/Fluid/Humidifiers/EvaporativeCoolers/Direct.mo
mwetter, line 1: This component requires a control input to modulate the evaporative cooling.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/IndirectDry.mo
mwetter, line 1: This component requires a control input to modulate the evaporative cooling.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/IndirectWet.mo
mwetter, line 1: This component requires a control input to modulate the evaporative cooling.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/IndirectDry.mo
```diff
@@ -0,0 +1,136 @@
+within Buildings.Fluid.Humidifiers.EvaporativeCoolers;
+model IndirectDry
+  "Indirect dry evaporative cooler"
+  extends Buildings.Fluid.Interfaces.PartialFourPortParallel;
```
mwetter: *All* other components in the library extend from `PartialFourPort`. Use this also here, otherwise we have one component for which the user need for no obvious reason connect the fluid ports differently.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/IndirectWet.mo
```diff
@@ -0,0 +1,246 @@
+within Buildings.Fluid.Humidifiers.EvaporativeCoolers;
+model IndirectWet "Indirect wet evaporative cooler"
+  extends Buildings.Fluid.Interfaces.PartialFourPortParallel;
```
mwetter: *All* other components in the library extend from `PartialFourPort`. Use this also here, otherwise we have one component for which the user need for no obvious reason connect the fluid ports differently.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/IndirectDry.mo
```diff
@@ -0,0 +1,136 @@
...
+          -56},{60,90},{120,90}}, color={0,0,127}));
+  connect(dirEvaCoo.port_b, hex.port_a2) annotation (Line(points={{10,-60},{20,-60},
+          {20,4},{10,4}}, color={0,127,255}));
+annotation (defaultComponentName="indDryEva",
+Documentation(info="<html>
+<p>
+Model for an indirect dry evaporative cooler.
+</p>
+<p>
+This model contains two components, a direct evaporative cooler
```
mwetter: Add a schematic diagram that shows the arrangement of the heat exchanger and humidifier. This will help users immediately understanding what the component does, and how the fluid ports are to be connected.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/IndirectWet.mo
```diff
@@ -0,0 +1,246 @@
...
+is cooled within the heat exchanger by an evaporative effect when water is directly
+sprayed on to the secondary surface of the heat exchanger to saturate the secondary
+air-flow. Therefore, it is assumed that the secondary air is completely saturated,
+and typically vented away after use.
+</p>
+<p>
+This model consists of the following components:
+</p>
+<ul>
+<li>
```
mwetter: Add a schematic diagram that shows the arrangement of the heat exchanger and humidifier. This will help users immediately understanding what the component does, and how the fluid ports are to be connected.

Such information is more useful than listing what sensors the model uses to feed data to its internal calculations.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/IndirectWet.mo
```diff
@@ -0,0 +1,246 @@
...
+  connect(volPri.ports[2], port_b1) annotation (Line(points={{71,30},{72,30},{
+          72,20},{92,20},{92,60},{100,60}}, color={0,127,255}));
+
+annotation (defaultComponentName = "indWetEva",
+Documentation(info="<html>
+<p>
+Model for an indirect wet evaporative cooler. In this component, the primary airflow
+is cooled within the heat exchanger by an evaporative effect when water is directly
+sprayed on to the secondary surface of the heat exchanger to saturate the secondary
+air-flow. Therefore, it is assumed that the secondary air is completely saturated,
```
mwetter: No dash in airflow. Use air flow rate, rather than air-flow.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/IndirectWet.mo
```diff
@@ -0,0 +1,246 @@
...
+Buildings.HeatTransfer.Sources.PrescribedTemperature</a>) and mixing volume 
+(<a href=\"modelica://Buildings.Fluid.MixingVolumes.MixingVolume\">
+Buildings.Fluid.MixingVolumes.MixingVolume</a>) for enforcing calculated outlet 
+air temperature.
+</li>
+</ul>
+<p>
+<b>Note:</b> The model works correctly only when the ports a1 and a2 are used as inlet ports, 
+and ports b1 and b2 are used as outlet ports, for the primary and secondary flow 
+respectively. Also, the secondary air outlet conditions are not validated,
```
mwetter: The secondary outlet is not only not validated, but their temperature and humidity is wrong.
You need to finish the model to make sure it outputs the right properties. Relying on users reading the note is not robust; and this would be the first model I know of in the library that produces incorrect output.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/Baseclasses/Direct.mo
```diff
@@ -0,0 +1,156 @@
...
+  Real eff(
+    final unit="1")
+    "Evaporative humidifier efficiency";
+
+  Buildings.Controls.OBC.CDL.Interfaces.RealInput V_flow(
+    final unit="m3/s",
+    final quantity = "VolumeFlowRate")
+    "Air volume flow rate"
+    annotation (Placement(transformation(origin={-120,-20},extent={{-20,-20},{20,20}}),
+      iconTransformation(origin={-120,-20}, extent={{-20,-20},{20,20}})));
```
mwetter: Change to using `m_flow` as input. This implementation is *very* convoluted:

First, in the sensor, density must be calculated to convert mass to volume flow rate. Then, here density must be computed to convert it back to mass flow rate. Moreover, the media must be propagated just to access the density function, which in fact uses default temperature and species, differently from the density computed in the volume flow meter.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/Baseclasses/Direct.mo
```diff
@@ -0,0 +1,156 @@
...
+  parameter Modelica.Units.SI.Density rho_default=Medium.density(sta_default)
+    "Density, used to compute fluid volume";
+
+  parameter Real effCoe[11] = {0.792714, 0.958569, -0.25193, -1.03215, 0.0262659,
+                               0.914869, -1.48241, -0.018992, 1.13137, 0.0327622,
+                               -0.145384}
+    "Coefficients for evaporative medium efficiency calculation";
+
+equation
+  vel =abs(V_flow)/padAre;
```
mwetter: We use regularization for this to make the function $C^1$.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/Baseclasses/IndirectWet.mo
```diff
@@ -0,0 +1,117 @@
...
+    "Dry bulb temperature of the outlet air"
+    annotation (Placement(transformation(origin={120,0}, extent={{-20,-20},{20,20}}),
+      iconTransformation(origin={140,0}, extent={{-20,-20},{20,20}})));
+
+  Real eff(
+    unit="1")
+    "Actual efficiency of component";
+
+equation
+  eff = max((maxEff - floRat*abs(VPri_flow)/abs(VSec_flow)),0);
```
mwetter: We use regularization for this to make the function $C^1$.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/Baseclasses/Direct.mo
```diff
@@ -0,0 +1,156 @@
...
+    effCoe[8]*(dep*vel<sup>3</sup>) + effCoe[9]*(dep<sup>3</sup>*vel) + effCoe[10]*(vel<sup>3</sup>*dep<sup>2</sup>) +
+    effCoe[11]*(dep<sup>3</sup>*vel<sup>2</sup>)
+</p>
+<p>
+where <code>effCoe[:]</code> is the evaporative efficiency coefficients for the
+CelDek rigid media pad used in evaporative coolers. It is currently protected from
+modification by the user, but can be modified for other materials within this class
+by advanced users. <code>dep</code> is the depth of the evaporative media, and
+<code>vel</code> is the velocity of the fluid media which is calculated from the volume flowrate
+<code>V_flow</code> and evaporative media cross-sectional area <code>padAre</code>
```
mwetter: What???
Not even a "advanced user" can modify this as it is `protected` and the language does not allow changing protected assignment. And this is for a specific pad. This makes me wonder if this is a generic model that can be parameterized for a specific equipment, or whether it just happened to be for one specific equipment, in which case exposing `dep` and `padAre` would not make sense.

You need to finish the implementation and documenation.
1. The coefficient need to be explained. How are they defined? Add it to the documentation of the model, including the top-level model. Are they invariant if `dep` or `padAre` are changed?
2. Propagate the device specific parameters `dep`, `padAre` and `coe` into a data record, to be put into `EvaporativeCoolers.Data`

Buildings/Fluid/Humidifiers/EvaporativeCoolers/Baseclasses/Direct.mo
```diff
@@ -0,0 +1,156 @@
...
+Block that calculates the water vapor mass flow rate addition in the 
+direct evaporative cooler component. The calculations are based on the direct 
+evaporative cooler model in the Engineering Reference document from EnergyPlus v23.1.0.
+</p>
+<p>
+The effectiveness of the evaporative media <code>eff</code> is calculated using 
+the curve
+</p>
+<p align=\"center\" style=\"font-style:italic;\">
+  eff = effCoe[1] + effCoe[2]*(dep) + effCoe[3]*(vel)  + effCoe[4]*(dep<sup>2</sup>) +
```
mwetter: Use proper notation:
1. no need to put the `dep` and `vel` into brackets, these are just factors.
2. Either use parameter names and Courier font, or use mathematical symbols and italics font.
3. This would be much more readable as a sum notation.

Also, explain where the user can get these data.

Buildings/Fluid/Humidifiers/EvaporativeCoolers/Baseclasses/Direct.mo
```diff
@@ -0,0 +1,156 @@
...
+                               0.914869, -1.48241, -0.018992, 1.13137, 0.0327622,
+                               -0.145384}
+    "Coefficients for evaporative medium efficiency calculation";
+
+equation
+  vel =abs(V_flow)/padAre;
+  eff = effCoe[1] + effCoe[2]*(dep) + effCoe[3]*(vel)  + effCoe[4]*(dep^2) +
+    effCoe[5]*(vel^2) + effCoe[6]*(dep*vel)  + effCoe[7]*(vel*dep^2) +
+    effCoe[8]*(dep*vel^3) + effCoe[9]*(dep^3*vel) + effCoe[10]*(vel^3*dep^2) +
+    effCoe[11]*(dep^3*vel^2);
```
mwetter: This implementation is very expensive, computing the 2nd and 3rd power of `vel` and `dep` multiple times. Do it efficiently.

mwetter (changes requested):
I don't see how to realistically use these components as the evaporative cooling is always operating. I expect users use these components as part of an air handler unit, in which case the evaporative cooling would not be operating if the system is in heating mode. Therefore, we need to add control input that allows modulating the humidification.

Moreover, the comment about advanced users being able to change the components of the `Direct` model does not make sense. It also makes me wonder if this is a generic component or whether it is for a particular product, yet it exposes the depth and area, which makes me think it can be changed. However, it is not clear if changing them invalidates the default coefficients, which are not defined and are `protected`.

Also, the consumed water must be an output to allow assessing water use.

So, this PR is far from being finished and need more work and substantial better review.

karthikeyad-pnnl:
> I don't see how to realistically use these components as the evaporative cooling is always operating. I expect users use these components as part of an air handler unit, in which case the evaporative cooling would not be operating if the system is in heating mode. Therefore, we need to add control input that allows modulating the humidification.
> 
> Moreover, the comment about advanced users being able to change the components of the `Direct` model does not make sense. It also makes me wonder if this is a generic component or whether it is for a particular product, yet it exposes the depth and area, which makes me think it can be changed. However, it is not clear if changing them invalidates the default coefficients, which are not defined and are `protected`.
> 
> Also, the consumed water must be an output to allow assessing water use.
> 
> So, this PR is far from being finished and need more work and substantial better review.

@mwetter Thank you for the feedback. I am adding some additional contextual information and a list of improvements I will make.

- The original intended use of this model was as zonal cooling equipment that cycles when the zone temperature exceeds the cooling setpoint. In addition to updating the documentation, I think it may make more sense to add a typical use-case in a zonal unit within the `Buildings.Fluid.ZoneEquipment` package.
I will also search for any specialized equipment that manage the cooling output for use in air-handlers, in which case it may make sense to incorporate the variable cooling output feature into the model. If I dont find any, I will still add an example of how the component can be integrated into an AHU to achieve supply temperature setpoint (variable cooling).
- Based on the EnergyPlus documentation, the performance coefficients represent a commonly used evaporative media called a CelDek pad, which can be procured in multiple sizes. I will look for any performance coefficients that might be available for other media, and update the code to propagate the parameters from a data record.

**List of actions**
- [ ] Add model for use in zonal unit and a suitable example for the zonal unit.
- [ ] Identify any special equipment for use in air handling units. If none are found, add example of how the component can be used to achieve supply temperature setpoint (variable cooling).
- [ ] Propagate the performance coefficients through data records. Moreover, now we have
```
eff = effCoe[1] + effCoe[2]*(dep) + effCoe[3]*(vel)  + effCoe[4]*(dep^2) +
    effCoe[5]*(vel^2) + effCoe[6]*(dep*vel)  + effCoe[7]*(vel*dep^2) +
    effCoe[8]*(dep*vel^3) + effCoe[9]*(dep^3*vel) + effCoe[10]*(vel^3*dep^2) +
    effCoe[11]*(dep^3*vel^2)
```
but `dep` is a parameter. There should be a package `Data` with performance data, and the efficiency in the model should then simply be a cubic polynomial (assuming this is the functional form that best meets Weiping's performance data).
- [ ] Update the models with secondary vented air-stream to incorporate correct changes in mixed-air states for the secondary stream.
- [ ] Address all other comments.
- [ ] Add a boolean input `on` to enable or disable the evaporative cooling
- [ ] Check if flow resistance is indeed proportional to m_flow^2. If not, use other model that allows coefficient different from 2. Otherwise use `PressureDropPartiallyTurbulent` from development branch `issue4443_dataCenterLiquidCooling`
- [ ] Rename package `EvaporativeCoolers` to `EvaporativePads` (because the `SprayAirWasher` is also an adiabatic cooler).

PR #3676, opened 2024-03-01
Issue3618 air filter

JayHuLBL:
This closes #3618.

PR #3807, opened 2024-04-18
Issue3618 air filter

SenHuang19:
This is to address the [comments](https://github.com/lbl-srg/modelica-buildings/pull/3676#issuecomment-2048609110).


SenHuang19:
@JayHuLBL Can you please take a look at it?

PR #3814, opened 2024-04-23
Issue2884 pid autotuning

SenHuang19:
this is to address https://github.com/lbl-srg/modelica-buildings/pull/2957#issuecomment-2062593673


SenHuang19:
@JayHuLBL can you please review it?

PR #3842, opened 2024-05-22
Set start attribute to avoid warnings

mwetter:
This resolves some conflicting start values.
Todo:
- [x] Port to `IBPSA` library.


mwetter:
@hcasperfu : The failing fan models work with the nightly build of OpenModelica but not with the docker version. Please put them on the exclude list.
https://app.travis-ci.com/github/lbl-srg/modelica-buildings/jobs/623419428

Issue #3808, opened 2024-04-18
Templates • FE to the AWHP plant template


AntoineGautier:
This ticket is to develop the following features for the template `Buildings.Templates.Plants.HeatPumps.AirToWater`.

- [x] Failsafe staging conditions for HP
- [x] Failsafe staging conditions for headered pumps with ∆p control
- [x] Optional sidestream HRC (need to update chiller component with `ElectricReformulatedEIR` and status emulator)
- [x] Local ∆p reset block in template control section outside of plant controller
- [x] Buffer tank integration: 3-port in common leg, 2-port in primary supply/return
    - 2-port buffer tanks have been added and can be integrated either in the primary supply (default for HW loop) or in the primary return (default for CHW loop): see additional comments on buffer tank integration on p. 29/30 of the document https://docs.google.com/document/d/1V6-akpnWRZ6MdZZYm0RrsMXIRgiD6Es_/edit?usp=sharing&ouid=100056680517301634140&rtpof=true&sd=true
    - The option for a 3-port buffer tank in the common leg is not included based on Brandon's comment:
      
      > I presume the primary benefit of the 3-port option is that the BT is always actively mixed/engaged, but we get the same thing with the two port approach and a cheaper tank.
      
      This integration point is also incompatible with primary-only distributions, adding complexity to the template without providing significant user benefits.
- [x] Modular HP system with primary-only distribution
    - The primary-only distribution has been added.
    - ⚠︎ Primary-only configurations exhibit non-linear solver failures, see #3759.
    - The modular HP system consists of identical modules that are staged by onboard controls rather than BAS controls. To accurately represent this modular system, a new component should be created to integrate the staging logic currently managed by the plant controller. However, there is limited value in this development, as users can already model an equivalent system using the existing template, which centralizes all control logic within the plant controller. This means that the practical outcomes in terms of system performance and efficiency can already be analyzed using the current template.
- [x] Automatic computation of pump speed or balancing valve ∆p_nom to meet design flow
    - ⚠︎ OCT fails to solve for the pump speed in a configuration with separate dedicated primary HW and CHW pumps, see #3759.
- [x] Test scripts similar to the ones in `Buildings/Resources/Scripts/travis/templates`

If possible before target date:
- [ ] Pump towards option (should be the default if no high pressure concern inside HP HX)
[...]

Issue #3759, opened 2024-04-01
Templates • Improve OCT support of heat pump plant template


AntoineGautier:
The purpose of this ticket is to investigate OCT initialization and convergence issues observed with the new heat pump plant template developed in `issue3605_template_HP_plant`.

- OCT version: 1.48.2 running on Ubuntu
- Simulation models that cause issues:
    -  `Buildings.Templates.Plants.HeatPumps.Components.Validation.ValvesIsolation`: configuration with separate dedicated primary CHW pumps
    - `Buildings.Templates.Plants.HeatPumps.Validation.AirToWater`: most plant configurations

AntoineGautier:
## `Templates.Plants.Controls.HeatPumps.Validation.AirToWater` with Dymola

Primary-only configurations exhibit non-linear solver failures. 
In most configurations, the tool recovers from these failures and the simulation succeeds. 

The simulation still fails for the following configurations. 
```mo
pla(
    typDis_select1=Buildings.Templates.Plants.HeatPumps.Types.Distribution.Variable1Only,
    typArrPumPri=Buildings.Templates.Components.Types.PumpArrangement.Dedicated,
    have_pumChiWatPriDed_select=false,
    linearized=false
```
```mo
pla(
    typDis_select1=Buildings.Templates.Plants.HeatPumps.Types.Distribution.Variable1Only,
    typArrPumPri=Buildings.Templates.Components.Types.PumpArrangement.Headered,
    linearized=false
```

The simulation fails with both DASSL and CVODE. Linearizing the flow/∆p equations in the valve components solves the issue. However, since the HP flow resistance is lumped with the isolation valve, there is an impact on the pump head and pump power. 

[EDIT] Even with linearized valves, the following configuration in a97b6ce68d cannot be simulated with Dymola.
```
modelInstance="Buildings.Templates.Plants.HeatPumps.Validation.AirToWaterReversibleHeatRecovery(
  pla(typDis_select1=Buildings.Templates.Plants.HeatPumps.Types.Distribution.Variable1Only),
  pla(typArrPumPri_select=Buildings.Templates.Components.Types.PumpArrangement.Dedicated),
  pla(have_pumPriDedComHp_select=true),
  pla(ctl(have_senDpHeaWatRemWir=false)),
  pla(typ=Buildings.Templates.Plants.Controls.Types.PlantHeatPump.ReversibleHeatRecovery))";
```

PR #3847, opened 2024-06-01
Add new features to the AWHP plant template

AntoineGautier:
This addresses #3808.

With this PR:

1. Exhaustive testing of all template configurations is now exclusively carried out with Dymola. This change is necessary because some configurations of the HP plant template fail to simulate with OCT (tracked under #3759), and the current test workflow (`make test-templates-*` in `.travis.yml`) does not support excluding specific configurations for a given simulator. 
However, the main validation model is still tested with all tools (`make test-*  PACKAGE=\"Buildings.Templates\"`), which ensures that the templates are at least validated in their default configuration.
2. Only ***30 %*** of all supported configurations are tested — for HP plants, but also VAV systems. This is because a single Travis job cannot last more than 120 min (https://docs.travis-ci.com/user/customizing-the-build#build-timeouts), and setting `TRAVIS_WAIT=180` or higher does not extend this limit (see for example the result of a failed test from this PR: https://api.travis-ci.com/v3/job/622428039/log.txt).
    - To prepare for an increasing number of templates in the future, we should think about having a Travis test for each package under `Templates`, or switch to a more flexible CI tool like Gitlab on a local server.


AntoineGautier:
This is ready for review.

AntoineGautier:

@karthikeyad-pnnl Consider checking out this branch for the development of other HP plant templates. It contains a few additions and corrections to the basic control blocks and components used to create the AWHP template currently in the master branch.

mwetter (changes requested):
@AntoineGautier : Can you please also update the schematic drawing that is shown in the info section of Buildings.Templates.Plants.HeatPumps.AirToWater

AntoineGautier:
@mwetter I've updated the schematics in 853181b. Please let me know if there are any other changes to consider. Otherwise, I'll trigger the CI tests and we can then merge this branch.

mwetter (approved):
@AntoineGautier : Thanks for the changes. This is ready to be merged.

Issue #3815, opened 2024-04-25
Incorrect computation of the load in EquationFitReversible and PartialElectric


AntoineGautier:
**Describe the bug**

Simulating `Fluid.HeatPumps.Examples.EquationFitReversible_CoolingClosedLoop` with `Q_flow.k=-4200` (instead of `Q_flow.k=4200` as on the master branch) shows that the heat pump component operates in heating mode (with a negative `heaPum.PLR`) despite being controlled in cooling mode (`heaPum.uMod=-1`).
This is because the model computes the load as:
```mo
  Modelica.Blocks.Sources.RealExpression Q_flow_set(
    final y= if (uMod == 0)
      then 
        0
      else 
        m1_flow*(hSet - inStream(port_a1.h_outflow)))
    "Required heat flow rate to meet set point"
```
without taking into account that the required heat flow rate should be zero if the setpoint is already met.

The model `Fluid.Chillers.BaseClasses.PartialElectric` has the same issue in heating control mode where `QCon_flow_set` should be bounded by `-Q_flow_small` as `QEva_flow_set`:
```mo
    // Cooling capacity required to chill water to setpoint
    QEva_flow_set = Buildings.Utilities.Math.Functions.smoothMin(
      x1=if coo_internal then m2_flow * (hSet - inStream(port_a2.h_outflow))
         else P * etaMotor - QCon_flow_set,
      x2=Q_flow_small,
      deltaX=-Q_flow_small/100);
    // Heating capacity required to heat up condenser water to setpoint
    QCon_flow_set = if coo_internal then QCon_flow
                    else m1_flow * (hSet - inStream(port_a1.h_outflow));
```

**Additional context**

Having the HP entering CHW temperature below setpoint can typically occur when a sidestream heat recovery chiller is used.

AntoineGautier:
Closed with https://github.com/lbl-srg/modelica-buildings/pull/3849.

Issue #3811, opened 2024-04-20
Incorrect parameter declaration in Fluid.Chillers.Data.ElectricReformulatedEIR.Generic


AntoineGautier:
The following declaration lacks the `parameter` keyword.
https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Fluid/Chillers/Data/ElectricReformulatedEIR.mo#L15

AntoineGautier:
Closed with https://github.com/lbl-srg/modelica-buildings/pull/3849.

PR #3849, opened 2024-06-04
Fix load computation in chiller and heat pump models

AntoineGautier:
This addresses #3815 and #3811.


mwetter (approved):

mwetter:
@AntoineGautier : This looks good. I think this should also be applied to the `maint_11.x` and perhaps `maint_10.0.x` branches?

PR #3879, opened 2024-06-12
Maint 11.x Fix load computation in chiller and heat pump models

AntoineGautier:
This addresses #3815.


mwetter (changes requested):
@AntoineGautier : looks good except that there should be an entry in the top-level release notes

mwetter (approved):

PR #3880, opened 2024-06-12
Maint 10.0.x Fix load computation in chiller and heat pump models

AntoineGautier:
This addresses #3815.


mwetter (approved):

mwetter (changes requested):
@AntoineGautier : looks good except that there should be an entry in the top-level release notes

AntoineGautier:
@mwetter Should it be under "Version 10.1.1" or "Version 10.0.0"?

mwetter:
> @mwetter Should it be under "Version 10.1.1" or "Version 10.0.0"?

It needs to be under 10.1.1 as this will be the version to be released.

mwetter (approved):

Issue #3923, opened 2024-07-01
Simplify Templates.Components.Controls.StatusEmulator


mwetter:
This model is currently implemented as follows:
![image](https://github.com/lbl-srg/modelica-buildings/assets/1329080/e7780347-9788-4141-bdbf-ecf87c9a0526)

with parameters
![image](https://github.com/lbl-srg/modelica-buildings/assets/1329080/96db6f0a-bdcc-4845-905c-78b4200c6347)

This introduces two differential equations and a state event detection. The model could be done much simpler if it were to use a timer that starts when the boolean input switches to false, and after `yLim*riseTime` has passed, the output switches to `true`.

@AntoineGautier : What do you think about this simplification?

AntoineGautier:
Not sure I understand exactly the refactoring with a timer. 
Assume the current configuration with the filter results in an actual delay time of about 2 s between the command `y1` and the returned status `y1_actual` (similar to the results from `Buildings.Templates.Components.Controls.Validation.StatusEmulator`):
- Time 0:05 - Boolean input `y1` switches to false: timer starts
- Time 0:07 - Delay time elapsed: `y1_actual` switches to true
- Time 0:10 - Boolean input `y1` switches to true: timer output and `y1_actual` switch to false?
 
However, the expected sequence of events is as follows (as in `Buildings.Templates.Components.Controls.Validation.StatusEmulator`):
- Time 0:05 - Boolean input `y1` switches to false: `y1_actual` remains true
- Time 0:07 - Delay time elapsed: `y1_actual` switches to false
- Time 0:10 - Boolean input `y1` switches to true: `y1_actual` remains false
- Time 0:12 - Delay time elapsed: `y1_actual` switches to true

In the branch `issue3923_statusEmulator` I proposed two alternative implementations using either the `delay` operator (`StatusEmulatorDelay`), or a state graph (`StatusEmulatorGraph`). 
Below are the validation model statistics for these variants compared to the original implementation using Dymola with the DASSL solver.
Based on these results, I believe StatusEmulatorDelay presents a viable alternative to the current implementation.

@mwetter, I would appreciate your feedback on these alternatives. If you identify any opportunities for further optimization, please feel free to commit additional changes to this branch.

<img width="625" alt="image" src="https://github.com/lbl-srg/modelica-buildings/assets/10965994/703a1dd8-0df2-4187-a3f1-e4310cfbd523">

mwetter:
[...]

PR #3925, opened 2024-07-04
Improve status emulator

AntoineGautier:
This addresses #3923.


mwetter:
Todo:
- [ ] Update checksum (and some other reference results may have to be updated due to this change).

mwetter:
@AntoineGautier : Thanks for the refactoring and update. Glad this is closed now.

Issue #3926, opened 2024-07-04
Improve templates test workflow


AntoineGautier:
Currently, the script `Buildings/Resources/Scripts/travis/templates/checkandrun.sh` checks that the last committed checksum (on HEAD) matches the checksum computed from the committed files.
This test is part of the travis job `make test-templates-*`. 
This ticket is to rather perform this quick test with a GH action so that the Travis tests are not run if the test fails.

AntoineGautier:
Closed with https://github.com/lbl-srg/modelica-buildings/pull/3927.

PR #3927, opened 2024-07-04
Improve checksum test

AntoineGautier:
This is for #3926.


AntoineGautier:
The CI tests for ffe4764 show that if the checksum test — performed as a GH action — fails, the Travis tests are not run.
The CI tests for 7b07a7b show that if the checksum test — performed as a GH action — succeeds, the Travis tests are run as they previously were.

Issue #3966, opened 2024-08-26
Incorrect output of TrueFalseHold when using a unit impulse input signal


AntoineGautier:
The issue appears with the following model proposed by @JayHuLBL (which yields the same results with Dymola and OCT).

<img width="555" alt="Screenshot 2024-08-20 at 4 02 22 PM" src="https://github.com/user-attachments/assets/010c78dd-45b6-4ace-9c19-57569781e134">

<img width="826" alt="Screenshot 2024-08-20 at 4 05 08 PM" src="https://github.com/user-attachments/assets/7dd38c47-8b37-476b-b476-1a4246535f94">

The issue is due to the use of the `change` operator in the implementation of `TrueFalseHold` to detect any change (from true to false, or from false to true) of input or output. 
However, in the case of a Dirac delta function, the `change` operator only fires a single event when the signal switches from false to true — but not when the signal switches back from true to false.

- The workaround for `TrueFalseHold` requires detecting the opposite transition with `edge(not_u)` and `edge(not_y)`. This is implemented in #3967. 
- I wonder whether the behavior of  the `change` operator in the case of a unit impulse signal is consistent with the MLS, which states:
  > `change` Is expanded into "(v<>pre(v))".

  > `pre` Returns the "left limit" y(tpre) of variable y(t) at a time instant t. At an event instant, y(tpre) is the value of y after the last event iteration at time instant t. 

  I conclude from these definitions that `change` should also fire at the second event iteration when the signal switches back from true to false. I will open a ticket at DS to confirm this.

  [EDIT] Peter Meisrimel from Modelon provided a detailed explanation whereas to why the `change` operator does not fire twice: see the model `ChangeOperator_x` below. 

  > The first rising edge is due to "not pre(sampleTrigger.y) and sampleTrigger.y", however the falling edge of "sampleTrigger.y" causes "change(sampleTrigger.y)" to remain True, due to which the when expression does not trigger a second time.

```mo
model ChangeOperator_x
discrete Integer y;
.Modelica.Blocks.Sources.SampleTrigger sampleTrigger(period=1, startTime=0.5);
Boolean b;
equation
b = change(sampleTrigger.y);
when b then
y = pre(y) + 1;
end when;
end ChangeOperator_x;

Event iter 0:
sampleTrigger.y: 0 -> 1
b = (sampleTrigger.y <> pre(sampleTrigger.y) = True <> False): False -> True
// (b and not pre(b) = True and not False = True)
[...]

PR #3967, opened 2024-08-26
Fix TrueFalseHold with unit impulse input signal

AntoineGautier:
This addresses #3966.


AntoineGautier:
## Unit tests failing

### `Buildings.ThermalZones.EnergyPlus_9_6_0.Examples.SmallOffice`

This is due to the following sequence within `Buildings.Controls.OBC.ASHRAE.G36.ZoneGroups.OperationMode`.

image

The following plot compares the results between the master branch `//1` and the feature branch `//2`.

image

We can verify that the feature branch yields the expected results.

AntoineGautier:
There is still an issue with the implementation in 72c45d3 that is visible in the plots for `Buildings.ThermalZones.EnergyPlus_9_6_0.Examples.SmallOffice` in https://github.com/lbl-srg/modelica-buildings/pull/3967#issuecomment-2310093495.
The condition `time >= pre(entryTimeFalse) + falseHoldDuration and time >= pre(entryTimeTrue) + trueHoldDuration` does not fire the when clause as expected based on https://specification.modelica.org/maint/3.6/equations.html#when-equations:

> The equations within a when-equation are activated only at the instant when the scalar expression or any of the elements of the vector expression becomes true.

This seems related to the CVODE solver. The plot below compares the trajectories of the same variables between DASSL (diagram 1) and CVODE (diagram 2).

image

As a workaround for CVODE, the vector expression is split into two separate when clauses.

AntoineGautier:
@JayHuLBL This is ready for your review. 
(Another PR more complicated than anticipated: 2 tickets open at DS, 1 at Modelon)

JayHuLBL (approved):
@AntoineGautier It looks good. Thanks for the fix.

AntoineGautier:
@mwetter This is ready to merge.

mwetter (approved):

mwetter:
Thanks, that was indeed a hard one...

PR #3971, opened 2024-08-27
Corrected media choices

JayHuLBL:


Buildings/UsersGuide/ReleaseNotes/Version_12_0_0.mo
mwetter, line 1: @JayHuLBL : The revision notes for version 12 won't need to be updated. We plan to release an update to version 11 first, and then release 12, so the revision notes only need to be added on the `maint_11.x` branch. Prior to the 12 release, I will copy these notes to version 12.

mwetter (changes requested):
@JayHuLBL : Please see notes regarding the revision notes, and I made some minor change to the IBPSA development branch that can be merged down.

JayHuLBL:
@mwetter I merged the change from IBPSA branch, and removed the version 12 release note.

mwetter (approved):

PR #3976, opened 2024-08-29
Ibpsa1912 avoid get peak function unless asked

hcasperfu:
This PR is to sync https://github.com/ibpsa/modelica-ibpsa/pull/1923 from IBPSA.


Buildings/Fluid/Movers/Preconfigured/UsersGuide.mo
```diff
@@ -42,6 +40,9 @@ Buildings.Fluid.Movers.BaseClasses.Characteristics.motorEfficiencyCurve</a>
 for details.
 </li>
 </ul>
+<h4>References</h4>
+<a href=\"https://simulationresearch.lbl.gov/modelica/releases/v10.0.0/help/Buildings_Fluid_HydronicConfigurations_UsersGuide.html#Buildings.Fluid.HydronicConfigurations.UsersGuide.ModelParameters\">
```
mwetter, line 44: This link should be to the Modelica URL (e.g., `href=\"modelica://Buildings....`). This way, it is always pointing to the documentation of the current version.

mwetter (changes requested):
Please see inline comment, otherwise it is ready to merge.

Buildings/Fluid/Movers/Preconfigured/UsersGuide.mo
hcasperfu: I used an external link because the referenced documentation is not available in IBPSA. Otherwise when syncing back to IBPSA the merge script would generate an invalid link. Is there a different workaround for this?

hcasperfu: In [171f866](https://github.com/lbl-srg/modelica-buildings/pull/3976/commits/171f86686fdf72fd6cd1f3443f29d5d76be577c2), I copied some texts and the figure from `HydronicConfiguration` documentation. This should be enough to resolve the confusion as to where this default curve comes from without solely referencing to an external source.
Creating a default curve would require further refactoring of the data record, because this curve is scaled on the nominal pressure and the nominal flow rate, and they are not known to the data record.

mwetter (approved):

PR #3977, opened 2024-08-29
Issue3964 add start time

### LINKED ISSUE #3964: Some models fails verification due to the stopTime is smaller than the startTime
The following models:
- `Buildings.Air.Systems.SingleZone.VAV.Examples.OptimalStart.ConventionalWinter`
- `Buildings.Air.Systems.SingleZone.VAV.Examples.OptimalStart.Guideline36Winter`

fail the verification vs the Modelica reference files because the `StopTime=604800`, written inside the model, is smaller than the `StartTime=6393600`, inherited from the extended model.

In the reference files the simulation time starts from zero, maybe that files have been produced with different settings.
In any case, the better way to avoid this kind of problems is to locally specify the experiment parameters.

Keeping @casella in the loop

JayHuLBL:
This closes #3964.


mwetter (approved):
@JayHuLBL : The change look good, please also merge to the maintenance branches as these are also tested by OpenModelica.

Issue #3978, opened 2024-08-30
Dymola 2025x


mwetter:
This issue is to test the library with Dymola 2025x.

mwetter:
Todo
- [ ] Check warning: 
      
     Pedantic translation of [Buildings.Electrical.AC.ThreePhasesBalanced.Lines.Examples.ACLine] 
     (modelica://Buildings.Electrical.AC.ThreePhasesBalanced.Lines.Examples.ACLine)

   The record constructor [Buildings.Electrical.Transmission.LowVoltageCables.Cu50] 
    (modelica://Buildings.Electrical.Transmission.LowVoltageCables.Cu50) has a replaceable function lineResistance.
    A redeclaration of that function will not be propagated to uses of the record constructor.

PR #3983, opened 2024-08-31
Correct compliance with MSL (maint 10)

mwetter:
This is for #3978


hcasperfu:
The branch test passed. The PR test failed because one of the jobs errored. It should pass if the job was rerun.

PR #3991, opened 2024-09-04
Issue3989 pump control

### LINKED ISSUE #3989: Pumps work against closed valves
In class `Buildings.Applications.DataCenters.ChillerCooled.Examples.NonIntegratedPrimarySecondaryEconomizer`, we have the results as below:

WatersideEconomizer

Between 1200 seconds and 1260 seconds, the primary pump is running at full speed (`priPum.u[1]=1`), while the chillers and the WSE are OFF (`chiWSE.on[1]=false`, `chiWSE.on[2]=false` and `chiWSE.on[3]=false`) and their isolation valves are closed. Thus the primary pump working pressure is at `-2E11 Pa`.

The cause is that in the instance `chiStaCon` the `tWai = 60 seconds` and in the instance `priPumCon` the `tWai = 0 seconds`.

We can address the issue by adding an `on` boolean input (from the instance `plaOn.y`) to the instance `priPumCon` and the instance `CWPumCon`, to avoid the pump works against the closed valves.

JayHuLBL:
This closes #3989.


Buildings/Applications/DataCenters/ChillerCooled/Examples/BaseClasses/PartialDataCenter.mo
```diff
@@ -551,6 +552,8 @@ equation
           {-126,226},{-126,248},{-122,248}},color={0,0,127}));
   connect(plaOn.y, varSpeCon.on) annotation (Line(points={{-138,240},{-136,240},
           {-136,210},{-190,210},{-190,4},{-170,4}}, color={255,0,255}));
+  connect(plaOn.y, CWPumCon.on) annotation (Line(points={{-138,240},{-136,240},
```
mwetter, line 555: Please update the revision notes of all changed classes.

mwetter (changes requested):
This is ready to merge if you update the revision notes of all changed classes.

mwetter (approved):

PR #4037, opened 2024-11-06
Flag for checking if the Co-Simulation has started or not

chria:
In the constructor of the external object `Buildings.ThermalZones.Detailed.BaseClasses.CFDThread`, there is a call to the underlying C method "`cfdcosim`". Furthermore, in the destructor of the same, there is a call to the underlying C method "`cfdSendStopCommand`". There is an issue in the latter as it currently assumes that the Co-Simulation has started when the destructor is called, i.e. that `Buildings.ThermalZones.Detailed.BaseClasses.cfdStartCosimulation` has been invoked.

This is not necessarily the case as in some unusal situations it might be so that the destructor is called without there having been a call to the `cfdStartCosimulation`. If this happens, the destructor will get stuck and eventually throw an error.

To handle these situations more gracefully I'd propose to simply add a flag in the underlying C-struct that indicates if the co-simulation has been started or not and that this is taken into account in the destructor.

### DIFF
--- a/Buildings/Resources/C-Sources/cfdSendStopCommand.c
+++ b/Buildings/Resources/C-Sources/cfdSendStopCommand.c
@@ -29,7 +29,7 @@ void cfdSendStopCommand(void *thread) {
   cosim->para->flag = 0;
 
   /* Wait for the feedback from FFD*/
- while(cosim->para->flag==0 && i<imax) {
+ while(cosim->started==1 && cosim->para->flag==0 && i<imax) {
     if(cosim->para->ffdError==1) {
       ModelicaError(cosim->ffd->msg);
     }
@@ -43,7 +43,7 @@ void cfdSendStopCommand(void *thread) {
     if(cosim->para->ffdError==1) {
       ModelicaError(cosim->ffd->msg);
     }
-    else {
+    else if (cosim->started==1) {
       ModelicaMessage("Successfully stopped the FFD simulation.\n");
     }
   }
--- a/Buildings/Resources/C-Sources/cfdStartCosimulation.c
+++ b/Buildings/Resources/C-Sources/cfdStartCosimulation.c
@@ -230,6 +230,7 @@ int cfdStartCosimulation(const char *cfdFilNam, const char **name, const double 
   | Implicitly launch DLL module.
   ****************************************************************************/
   ffd_dll(cosim);
+  cosim->started = 1;
 
   return 0;
 } /* End of cfdStartCosimulation()*/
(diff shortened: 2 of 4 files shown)


mwetter:
@chria : Thanks for the code fixes.

@JayHuLBL : Can you please address this, build the library and port it to the maintenance branches (`maint_9.1.x` and higher).

JayHuLBL:
The CI tests stop running due to the error:
```
The command "echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin" failed and exited with 1 during .
```
@mwetter Is it the travis issue?

JayHuLBL:
I see the reason. I will create a new PR to trigger the tests.

JayHuLBL:
When compiling the code to build the library (`gcc version 11.4.0`), it will give error:
```
gcc -shared -fPIC -Wl,-soname,libffd.so -o libffd.so advection.o boundary.o chen_zero_equ_model.o cosimulation.o data_writer.o diffusion.o ffd.o ffd_data_reader.o ffd_dll.o geometry.o initialization.o interpolation.o parameter_reader.o projection.o sci_reader.o solver.o solver_gs.o solver_tdma.o timing.o utility.o -lpthread -lc
/usr/bin/ld: boundary.o:(.bss+0x0): multiple definition of `msg'; advection.o:(.bss+0x0): first defined here
/usr/bin/ld: boundary.o:(.bss+0x3e8): multiple definition of `file1'; advection.o:(.bss+0x3e8): first defined here
/usr/bin/ld: boundary.o:(.bss+0x3f0): multiple definition of `file_old_ffd'; advection.o:(.bss+0x3f0): first defined here
/usr/bin/ld: boundary.o:(.bss+0x3f8): multiple definition of `file_params'; advection.o:(.bss+0x3f8): first defined here
...
```
This is due to: https://gcc.gnu.org/gcc-10/porting_to.html#common, as below:

----
A common mistake in C is omitting extern when declaring a global variable in a header file. If the header is included by several files it results in multiple definitions of the same variable. In previous GCC versions this error is ignored. GCC 10 defaults to -fno-common, which means a linker error will now be reported. To fix this, use extern in header files when declaring global variables, and ensure each global is defined in exactly one C file. If tentative definitions of particular variables need to be placed in a common block, __attribute__((__common__)) can be used to force that behavior even in code compiled without -fcommon. As a workaround, legacy C code where all tentative definitions should be placed into a common block can be compiled with -fcommon.


      int x;  // tentative definition - avoid in header files

      extern int y;  // correct declaration in a header file
 ----

To fix it, we need add `extern` keyword when declaring global variables, i.e. `msg`, `file_params`, `file1`, `file_old_ffd`, `file_log`, `file_para`, in the head files.

JayHuLBL:
Close this PR as it has been replaced by #4039.

PR #4059, opened 2024-11-23
Remove time events in ETS model with heat recovery chiller

### LINKED ISSUE #4058: Reduce time events in DHC.ETS.Combined.ChillerOnly (and related models)
Models such as `Buildings.DHC.ETS.Combined.Examples.ChillerOnly` sample every 60 seconds to keep a boolean signal constant for a while and through that avoid short-cycling.
This sampling is done with a `FirstOrderHold`. If it were done with a `TrueFalseHold`, then the number of time events would be significantly reduced.

For example, for `Buildings.DHC.ETS.Combined.Examples.ChillerOnly`, the current implementation has in Dymola 8000 time events and takes 1.4 seconds to simulate, and with the proposed implementation, it has 150 time events and takes 0.6 seconds to simulate.

mwetter:
This closes #4058
It also removes a duplicate signal connection.


mwetter:
@AntoineGautier : Can you please have a brief look into this refactoring. It replaces a `FirstOrderHold` with 60 s sampling time with a `TrueFalseHold` and a `Pre` at its input. The `TrueFalseHold` avoids the frequent time events and the `Pre` is needed for OpenModelica to avoid a nonlinear equation in a `when` construct during the initialization.

AntoineGautier:
The change looks good to me.
In addition, I suggest increasing the minimum cycle time from 60 s to 5 min. This has minimal impact on the supply temperatures and avoids cycling the two HX pumps every 60 s, which seems a little demanding.
I committed this modification via the last two commits. The comparison results can be downloaded from here:
https://drive.google.com/file/d/1uEiDJpI6ymrWTNUKxW2WFFKu6d5cGJZS/view?usp=drive_link
Extract archive and run `python -c "import buildingspy.development.regressiontest as t; t.Tester(tool=\"dymola\").report()"` to visualize.

mwetter:
@AntoineGautier : Thanks, increasing it to 5 min is good.

PR #4075, opened 2024-12-10
Made test case robust against rounding errors

### LINKED ISSUE #4074: Buildings.Templates.Plants.Controls.HeatRecoveryChillers.Validation.EnableAndModeControl fails verification
This is to make the test case robust against numerical noise. See
https://github.com/OpenModelica/OpenModelica/issues/12726

mwetter:
This closes #4074


AntoineGautier (approved):
Yes, that's better than before.

PR #4104, opened 2025-01-29
Changed BuildingsPy to recent development version

mwetter:
This updates BuildingsPy to the lastest stable version


mwetter:
This fails because of the reason documented in https://github.com/OpenModelica/OpenModelica/issues/13446#issuecomment-2625378180

Issue #4105, opened 2025-01-31
Update OCT image to 1.55.11


mwetter:
This issue is to update the OCT image to version 1.55.11

PR #4108, opened 2025-01-31
Updated OCT image to 1.55.11

mwetter:
This is for #4105


mwetter:
This has the same CI test timeout as reported in https://github.com/lbl-srg/modelica-buildings/issues/4128

PR #4130, opened 2025-03-05
kappa calculation moved from initial equation to binding equation

AndreaBartolini:
this fixes OpenModelica Issue #12677


casella:
Here is the link to the issue: OpenModelica/OpenModelica#12677. I'll post an extensive comment there in a short while.

casella:
See https://github.com/OpenModelica/OpenModelica/issues/12677#issuecomment-2701725063

casella:
Fixes OpenModelica/OpenModelica#12677. A small drawback of this solution is that the aggregationCellTimes() function is called twice with the same argument. I think OMC should be smart enough to figure out that they are the same function call and factor it out. Even if it doesn't, I understand the overhead is marginal, since the extra computation is performed only once during initialization.

mwetter:
Thanks for the change. I will process it from here and merge it to the master branch.

PR #4195, opened 2025-04-17
Model fixes for staging process issues, and functioning example model

karthikeyad-pnnl:


AntoineGautier:
@karthikeyad-pnnl Thanks for the quick feedback.
The model `Buildings.Examples.BoilerPlants.ClosedLoopTest` from commit 3244420 cannot be simulated on my end.
Simulation fails after 906 s (out of 172800 s from the experiment annotation) with all solvers I've tested. 
Is there some specific settings you are using to simulate this model?
Right now it does not simulate enough to validate the event sequencing at plant enable. 
image

In the HW plant template the boilers and primary pumps remain disabled and the boiler isolation valves remain closed although `pla.ctl.ctlLooPri.staSetCon.ySta` transitions to 1.

image

karthikeyad-pnnl:
@AntoineGautier Huh yeah that sounds weird. I am just using the "simulate and plot" command. Let me know if that doesn't work either on your end.

Also, can you please point me to your working branch for the template? I will try it out on my end and let you know if that behaves differently on my system. Thanks!
image

AntoineGautier:
@karthikeyad-pnnl Thanks! This is so weird. I've never experienced something similar. 

- Are we talking about the same model `Buildings.Examples.BoilerPlants.ClosedLoopTest` from commit 3244420?
- Which version of Dymola do you use? I run Version 2025x, 2024-12-20 on Ubuntu 22.04. I will test on Windows and on another Linux machine as well.

You can check out the template model in the branch `issue3266_template_HW_plant`. 
The validation model with the controller is `Buildings.Templates.Plants.Boilers.HotWater.Validation.BoilerPlant`.

AntoineGautier:
Adding a recap of email iterations here for reference.

Convergence issues of `Buildings.Examples.BoilerPlants.ClosedLoopTest` from commit https://github.com/lbl-srg/modelica-buildings/commit/32444209f44dff09050a03d6d74f682e77309cf3 can be reproduced using Dymola 2025x.

The reason why the HW plant template validation model `Buildings.Templates.Plants.Boilers.HotWater.Validation.BoilerPlant` from 264119bd89 shows no equipment enable at plant enable is because of the staging matrix specified as `sta=[0,0; 1,0; 1,1]`. Changing to `sta=[1,0; 1,1]` allows enabling equipment but still reveals some issues with the event sequencing:  the isolation valve is open and then closed, the primary pump is enabled and then disabled and the boiler remains enabled after the valve is closed and the pump is disabled: see plot below.

image

There is a question about the stage 0 of the staging matrix with no equipment enabled. This complies with G36 but does not seem to be handled by the implemented staging logic yet.

image

PR #4221, opened 2025-05-04
Delete package Buildings.Fluid.CHPs.DistrictCHP.Obsolete and ensure a…

ZhanweiHe:
1) Delete the package `Buildings.Fluid.CHPs.DistrictCHP.Obsolete` 
2) Ensure all instances have descriptive comments


JayHuLBL:
@ZhanweiHe The change is not sufficient.

`Instance` does not mean the class. The comments added in this PR are comments for the class.

We need to have comment for all the instances. Like in https://github.com/lbl-srg/modelica-buildings/blob/f4850bad625e33f1ee57347bb13e8698567ee777/Buildings/Fluid/CHPs/DistrictCHP/BottomingCycle.mo#L193, the instance `pump` does not have comment.

Issue #4227, opened 2025-05-07
Replace direct fluid pass-through with fixed resistance in Templates.Components.Actuators


AntoineGautier:
This ticket is to replace the class `Templates.Components.Routing.PassThroughFluid` with `Fluid.FixedResistances.PressureDrop` for the valve and damper models within `Templates.Components.Actuators`.

This will allow using the same component in configurations where no modulating valve or damper is used, but a balancing valve or damper is needed (which can be parameterized with `dpFixed_nominal`).

**Development branch**: `issue4227_fixedResistanceActuators`

mwetter:
@AntoineGautier : I think this can be closed?

PR #4228, opened 2025-05-07
Replace fluid pass-through with fixed resistance in actuators

AntoineGautier:
This is for #4227.

Note that with this PR, the parameter `m_flow_nominal` needs to be assigned a non-zero value (via the parameter record) even if the damper or valve type is set to `None`. Previously, this parameter could be left unassigned (since it was declared with a `start` attribute and a `Dialog(enable=false)` annotation).

This is not seen as an issue for the template development because a default value can always be provided as in:
 https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Templates/AirHandlersFans/Data/VAVMultiZone.mo#L38-L45
or:
https://github.com/lbl-srg/modelica-buildings/pull/4228/commits/30ec283df23ab940014212ded5db481590b67913#diff-acd47242c40cc65fce0b368f54ca2b640e80b9ca5774dab11288ae2bd81cb934

In the equipment schedule page of ctrl-flow (to be developed), the whole parameter record is disabled if the type is `None`, which should prevent requesting a value for `m_flow_nominal`.


mwetter:
@AntoineGautier : Is this ready to merge or more work needed from your end? I plan to bump the version number to 12.1.0, so merging this before will avoid a merge conflict.

AntoineGautier:
Yes, ready to merge on my end.

PR #4232, opened 2025-05-09
Issue 4199 Fix nominal mass flow in boreholes

### LINKED ISSUE #4199: ZonedBorefields assign internally the wrong value for m_flow_nominal, leading to wrong pressure drop
Simulating on master (a15d21dd3aff4ebae800f90c062a15cb38dddcc4) the model `Buildings.Fluid.Geothermal.ZonedBorefields.Validation.TransientBoreholeDynamics` gives the mass flow rates `TransientBoreholeDynamics[end].borHol.borHol[1].intHex[1].preDro1.m_flow_nominal = 40 kg/s` and
`TransientBoreholeDynamics[end].borHol.borHol[2].intHex[1].preDro1.m_flow_nominal = 44 kg/s`

The design flow rate for each zone is

Image
and the number of boreholes per zone is

Image

Hence, the design flow rate should be for
- zone 1: 0.5*80 = 40 kg/s
- zone 2: 0.3*148= 44.4 kg/s

This is assigned correctly. However, this design flow rate is also used for the pressure drop elements:

Image

This leads to the wrong pressure drop calculation:

Image

The problem seem to be at the level of `borHol` as this instance receives the flow rate after `masFloDiv`.

I have not checked if this also affects other borefield models or only the zoned borefields.

MassimoCimmino:
This fixes #4199.

### DIFF
--- a/Buildings/Fluid/Geothermal/ZonedBorefields/BaseClasses/PartialStorage.mo
+++ b/Buildings/Fluid/Geothermal/ZonedBorefields/BaseClasses/PartialStorage.mo
@@ -77,7 +77,7 @@ partial model PartialStorage
     redeclare each final package Medium = Medium,
     final borFieDat=zonDat,
     each final nSeg=nSeg,
-    final m_flow_nominal=borFieDat.conDat.mZon_flow_nominal,
+    final m_flow_nominal=borFieDat.conDat.mBor_flow_nominal,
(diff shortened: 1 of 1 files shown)


mwetter:
@MassimoCimmino : Thank you for the correction. I will do some more testing and then merge it to the master and maintenance 12 branch.

Issue #4269, opened 2025-06-05
Models Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.xxx fail verification due to critical threshold


AndreaBartolini:
The following models fail the verification vs the Dymola reference file:
- `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.DirectActingPIDWithFirstOrderAMIGO`
- `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.DirectActingPIWithFirstOrderAMIGO`
- `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.ReverseActingPIDWithFirstOrderAMIGO`
- `Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.ReverseActingPIWithFirstOrderAMIGO`

the complete analysis of the failure is available here: https://github.com/OpenModelica/OpenModelica/issues/13978

The critical point is that there is a threshold that should evaluate the condition `0 > 1e-15` that is nine orders of magntude lower than the tolerance required to the solver (`1e-6`).

Dymola, in some way, manages the condition, omc doesn't. As consequence the output is different in Dymola and omc. If the tolerance is reduced to `1e-12` then the transients generated by Dymola and omc overlap.

In my opinion there is a very critical design in the model logic (further details are at the link above), that probably should be revised in order to avoid to use thresholds to compare signals that differ for a quantity that is orders of magnitude lower than the tolerance required to the solver.

Keep @casella in the loop.

mwetter:
@SenHuang19 : Can you please address this and make a PR. Thanks.

PR #4270, opened 2025-06-06
issue4269_AutoTunPID_verification

SenHuang19:
This PR addresses https://github.com/lbl-srg/modelica-buildings/issues/4269


SenHuang19:
> @SenHuang19 Do we need to have the constant? I think the purpose of the constant is for check if the `tOn` or `tOff` is greater than 0.
> 
> I think we can change the `Buildings.Controls.OBC.CDL.Reals.Greater` to `Buildings.Controls.OBC.CDL.Reals.GreaterThreshold` and use the threshold as 0, since the input `tOn` and `tOff` are output from the triggered sampling block.

@JayHuLBL with `Buildings.Controls.OBC.CDL.Reals.GreaterThreshold` , we can't set the threshold to 0, as the OM issue persists under default solver settings. Therefore, a constant is still required when using this block.

PR #4271, opened 2025-06-06
Issue4269 pid autotune verification

### LINKED ISSUE #4269: Models Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Validation.xxx fail verification due to critical threshold

JayHuLBL:
This closes #4269.

### DIFF
--- a/Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/BaseClasses/TuningMonitor.mo
+++ b/Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/BaseClasses/TuningMonitor.mo
@@ -1,8 +1,7 @@
 within Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.BaseClasses;
 block TuningMonitor "Monitor the tuning process"
-  constant Modelica.Units.SI.Time minHorLen = 1E-6
-    "Minimum value for horizon length, used to guard against rounding errors, 
-    must be greater or equal to the solver tolerance";
+  constant Modelica.Units.SI.Time minHorLen = 1E-5
+    "Minimum value for horizon length, used to guard against rounding errors, must not be smaller than the solver tolerance";
   Buildings.Controls.OBC.CDL.Interfaces.RealInput tOn(
     final quantity="Time",
     final unit="s",


Buildings/Controls/OBC/Utilities/PIDWithAutotuning/Relay/BaseClasses/TuningMonitor.mo
```diff
@@ -1,7 +1,7 @@
 within Buildings.Controls.OBC.Utilities.PIDWithAutotuning.Relay.BaseClasses;
 block TuningMonitor "Monitor the tuning process"
-  constant Modelica.Units.SI.Time minHorLen = Buildings.Controls.OBC.CDL.Constants.eps
-    "Minimum value for horizon length, used to guard against rounding errors";
+  constant Modelica.Units.SI.Time minHorLen = 1E-5
```
mwetter: @SenHuang19 : I don't understand this code and its comment: First, a minimum horizon length of 1E-5 seconds is unreasonable as this is incredible fast, faster than (HVAC) controllers are clocked. Second, it says "...must not be smaller than the solver tolerance". What solver is this? This is used I believe to detect some event. Eventually it is used to compare `min(tOn, tOff) > minHorLen` and `max(tOn, tOff) > minHorLen` which I don't know how this relates to anything that has to do with "the solver" . Also, `tOn` and `tOff` change only when `on` changes or the `trigger` changes, which makes the relation to a "solver tolerance" even more dubious. Can you please properly explain in the model that this is, and why the suggested value is used.

mwetter (changes requested):
@SenHuang19 : Please see the inline comment.

mwetter (approved):

PR #4301, opened 2025-07-22
Issue4278 function inlining

### LINKED ISSUE #4278: Dymola inlining of functions with protected variables
Dymola has a flag `Advanced.Translation.Log.FailureToInline=true;` to report functions with `annotation(Inline=true)` that could not be inlined.
There is one easy way to make some functions possible to be inlined, and the trick is to compute the value of protected variables directly in the declaration section.

Consider this example:
- Buildings.Utilities.Math.Examples.QuadraticLinear
- Buildings.Utilities.Math.QuadraticLinear
- [Buildings.Utilities.Math.Functions.quadraticLinear](https://github.com/lbl-srg/modelica-buildings/blob/master/Buildings/Utilities/Math/Functions/quadraticLinear.mo)

By rewriting it as below, it can be inlined:

```modelica
function quadraticLinear
  "Function that is quadratic in first argument and linear in second argument"
  extends Modelica.Icons.Function;
 input Real a[6] "Coefficients";
 input Real x1 "Independent variable for quadratic part";
 input Real x2 "Independent variable for linear part";
 output Real y "Result";
protected 
 Real x1Sq=x1*x1;
algorithm 
  y := smooth(999, a[1] + a[2]*x1 + a[3]*x1Sq + (a[4] + a[5]*x1 + a[6]*x1Sq)*x2);

  annotation (Inline=true, Documentation(info="<html>
This function computes
<p align=\"center\" style=\"font-style:italic;\">
  y =   a<sub>1</sub> + a<sub>2</sub>  x<sub>1</sub>
        + a<sub>3</sub> x<sub>1</sub><sup>2</sup>
        + (a<sub>4</sub> + a<sub>5</sub>  x<sub>1</sub>
        + a<sub>6</sub> x<sub>1</sub><sup>2</sup>)  x<sub>2</sub>
</p>
</html>", revisions="<html>
<ul>
<li>
February 29, 2009 by Michael Wetter:<br/>
First implementation.
</li>
</ul>
</html>"));
end quadraticLinear;

```

That variant is also committed here, in case you consider it a useful change: https://github.com/thorade/modelica-buildings/blob/feature/issue4278_inlineProtected/Buildings/Utilities/Math/Functions/quadraticLinear.mo

For this single function, it probably does not make any difference, but doing it for several functions might give a small performance improvement.

Just a note: See discussions in this old ticket:
https://github.com/ibpsa/modelica-ibpsa/issues/300#issuecomment-127042618

hcasperfu:
This closes #4278.

The following functions are now inlined:
- `Buildings.Utilities.Math.Functions.bicubic`
- `Buildings.Utilities.Math.Functions.quadraticLinear`
- `Buildings.Utilities.Math.Functions.quinticHermite`
- `Buildings.Utilities.Math.Functions.smoothHeaviside`
- `Buildings.Utilities.Math.Functions.smoothLimit`

@mwetter -
I will make the same changes to IBPSA once you approve this PR.


Buildings/Utilities/Math/Functions/bicubic.mo
```diff
@@ -6,20 +6,18 @@ function bicubic "Bicubic function"
...
-  x1Sq :=x1*x1;
-  x2Sq :=x2*x2;
-  y := a[1] + a[2] * x1 + a[3] * x1^2
-            + a[4] * x2 + a[5] * x2^2
-            + a[6] * x1 * x2
-            + a[7] * x1Sq * x1
-            + a[8] * x2Sq * x2
-            + a[9] * x1Sq * x2
-            + a[10] * x1 * x2Sq;
+  y := smooth(999, a[1] + a[2] * x1 + a[3] * x1^2
```
mwetter: @hcasperfu : I don't think that removing `smoothOrder` and adding `smooth(...)` is required. In fact, `smoothOrder` applies to the whole function and not only to one assignment in it.
Please revert the change from `smoothOrder` to `smooth(...)` unless it is indeed needed to inline the function.

mwetter (changes requested):
Please see inline comment.

Buildings/Utilities/Math/Functions/bicubic.mo
hcasperfu: Reverted in 7cf89b2.

PR #4334, opened 2025-09-05
Issue2885 fan coil unit controls 060325

karthikeyad-pnnl:
@JayHuLBL I am submitting this PR with a simplified closed-loop example. The integration time for this new model is down to <4 seconds. Can you please check if this passes the CI tests? I was able to get the Openmodelica unit test to pass when I ran it on my local setup. In the mean time, I will clean the models up and submit an additional PR for it. Thanks!


karthikeyad-pnnl:
@JayHuLBL I had a chat about it with Michael at the last meeting he made. I think he would like us to eliminate the vectorization and create separate instances for the fan coil unit and controller instead. The good news is that the new setup runs in about 4 seconds. I also got the whole Examples package to run the openmodelica unit test without resulting in any translation timeout errors, so I think it should pass the CI tests now. I have also addressed a few of Michael's other comments in the changes I have pushed. The only pending update is adding a schematic and updating the fan class, but I would like some feedback on whether the unit test passes in the meantime if possible. Thanks!

PR #4352, opened 2025-09-18
Updated autosizing implementation

mwetter:
@dhblum This implements the following changes

- Updated json file and C code to remove the non-needed parameters `autosize` and `runSimulationForSizingPeriods`.
- Changed the parameter value for time from `T` to `t` (as we use `T` typically for temperature)
- Changed order of specifying and getting parameters for thermal zone and for system (so that the order is the same for heating and for cooling)
- Update all examples and validations. They all check successfully, but don't simulate as it requires the new binaries for EnergyPlus.
- Updated all validations in `Buildings.ThermalZones.EnergyPlus_24_2_0.BaseClasses.Validation` and the associated FMUs in `Resources/src/ThermalZones/EnergyPlus_24_2_0/FMUs` 
- Corrected old annotation `fillColor` for `Text` annotations. This has also been ported to `maint_12.x` and `maint_11.x`.

*Note* 
1. The validations in `Buildings.ThermalZones.EnergyPlus_24_2_0.BaseClasses.Validation`  work with OpenModelica, Optimica and with Dymola, but with Dymola, they work with the solver dassl but not with CVode. I don't know why they don't work with CVode as there is no error message.
2. Compiling the binaries gives a warning `EnergyPlus_24_2_0/C-Sources/cryptographicsHash.c:221:13: warning: ‘SHA1Transform’ reading 64 bytes from a region of size 7`. This seems to be OK based on see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106709 All other warnings are now removed.
3. I have not compiled the Windows binaries. We can do this once everything works on Linux.

Please go ahead and merge it if the changes are fine.

### DIFF
--- a/Buildings/ThermalZones/EnergyPlus_24_2_0/ThermalZone.mo
+++ b/Buildings/ThermalZones/EnergyPlus_24_2_0/ThermalZone.mo
@@ -3,6 +3,15 @@ model ThermalZone
   "Model to connect to an EnergyPlus thermal zone"
   extends
     Buildings.ThermalZones.EnergyPlus_24_2_0.BaseClasses.PartialEnergyPlusObject;
+  ////////////////////////////////////////////////////////////////////////////
+  // Buildings.Media declaration. This is identical to
+  // Buildings.Fluid.Interfaces.LumpedVolumeDeclarations, except
+  // that the comments have been changed to avoid a confusion about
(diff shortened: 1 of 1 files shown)


dhblum:
@mwetter Ah ok I see now, thank you!

PR #4356, opened 2025-09-19
Merge Ibpsa issue2043 two way flow element

mwetter:
This merges optimization of the code for the multizone airflow models.


mwetter:
This closes https://github.com/lbl-srg/modelica-buildings/issues/4360

PR #3627, opened 2024-01-05
Issue2885 fan coil unit controls

JayHuLBL:
This closes #2885.

PR #4357, opened 2025-09-20
Issue2885 fan coil unit controls 060325

karthikeyad-pnnl:
@mwetter I have applied the changes we discussed over the call. The simulation time has reduced even further to about 2 seconds in Dymola. Please review these changes when you get a chance. Thanks!


karthikeyad-pnnl:
@mwetter This needs to be merged before further testing #3627

PR #4393, opened 2025-10-08
Issue4378 autotuning pid

### LINKED ISSUE #4378: Autotuning PI controller obtains negative control gains
@SenHuang19 : I was trying to set up a small model with the autotuning PI controller, building on top of `Buildings.Fluid.HeatExchangers.Examples.WetCoilCounterFlowPIControlAutoTuning`.

I created a new branch with the above model slightly modified so that the 2nd tuning does not coincide with the set point change. When running the model, the control gains for the P and the I part switch from positive to negative, and control is lost.

JayHuLBL:
This closes #4378.


JayHuLBL:

PR #4415, opened 2025-11-20
Created air filter package

### LINKED ISSUE #3618: Air filter models
This issue is to develop a new component model for approximating the dynamic behaviors of air filters in general.
See details in [Generic Air filter Model.pdf](https://github.com/lbl-srg/modelica-buildings/files/13753595/Generic.Air.filter.Model.pdf)

JayHuLBL:
This replaces PR #4411.

It closes #3618.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ env:
     - DOCKER_REPONAME=lbnlblum
     - BUILDINGSPY_VERSION=BuildingsPy@v5.2.0
     - OMC_VERSION=ubuntu-2204-omc:1.24.0-1
-    - OPTIMICA_VERSION=travis-ubuntu-2204-optimica:1.55.11
+    - OPTIMICA_VERSION=travis-ubuntu-2204-optimica:1.55.11_rev-1
     - DYMOLA_VERSION=travis_ubuntu-2004_dymola:2025x-x86_64_rev-2
     - ENERGYPLUS_9_6_0=EnergyPlus-9.6.0-f420c06a69-Linux-Ubuntu20.04-x86_64
     - ENERGYPLUS_23_1_0=EnergyPlus-23.1.0-87ed9199d4-Linux-Ubuntu20.04-x86_64


JayHuLBL:

mwetter:
Todo
- [ ] In data record, clearly explain what `b` is.
- [ ] In data record, change arrays so there is one index for the substances, and then an array with the performance for that substance.
- [ ] Update the Validation model as `mMax=10` is not realistic, and why not run it for a year with realistic data?

Buildings/Fluid/AirFilters/BaseClasses/FiltrationEfficiency.mo
```diff
@@ -0,0 +1,74 @@
+within Buildings.Fluid.AirFilters.BaseClasses;
+model FiltrationEfficiency
+  "Filtration efficiencies for capturing each contaminant"
+  extends Modelica.Blocks.Icons.Block;
+
+  parameter Modelica.Units.SI.Mass mCon_max
+    "Maximum mass of the contaminant that can be captured by the filter";
+  parameter String namCon[:]
+    "Name of trace substance";
+  parameter Buildings.Fluid.AirFilters.Data.Characteristics.FiltrationEfficiencyParameters filEffPar[nConSub]
```
JayHuLBL, line 10: @mwetter After changing the filtration efficiency parameters to one set for each contaminant, running the validation model `BaseClasses.Validation.FiltrationEfficiency` with Dymola gives the error:
Screenshot 2025-12-23 at 1 17 31 PM
However, there is no error with OpenModelica and Optimica. Could it be a Dymola bug?

JayHuLBL: The issue is being addressed by the Dymola development team:
Screenshot 2026-01-08 at 10 35 02 AM

---
There is a workaround as below to bypass the issue:
In the class `Buildings.Fluid.AirFilters.BaseClasses.FiltrationEfficiency`, we could create the protected parameters:
```modelica
  parameter Real filRat[:,:]=filEffPar.rat
    "Relative mass of captured contaminant";
  parameter Real filEps[:,:]=filEffPar.eps
    "Filtration efficiency";
```

and then in the for loop in the equation section, replace the `filEffPar[i].rat` and `filEffPar[i].eps` with `filRat[i,:]` and `filEps[i,:]`:
```modelica
  for i in 1:nConSub loop
     y[i] = Buildings.Utilities.Math.Functions.smoothLimit(
        Buildings.Utilities.Math.Functions.smoothInterpolation(
                x=rat,
                xSup=filRat[i,:],
                ySup=filEps[i,:]),
        0,
        1,
        1E-3)
     "Calculate the filtration efficiency";
  end for;
```

JayHuLBL:

mwetter:
@JayHuLBL : I revised the models, primarily
- revised documentation
- removed `min` block that avoided the dimensionless mass to exceed 1 if the filter is not replaced. I think this value should keep increasing if the filter is not replaced.
- removed the attributes `rat(max=1)`  because of the above item
- removed a real delay in an example (and tested it with OpenModelica)

Can you please update the reference results, which may have slightly changed due to the removal of the `min` block.
Then it is ready to merge from my view. (Note that the CI tests with Optimica will fail until the image is updated.)

Issue #4407, opened 2025-11-17
The transation of theThe translation of the model Buildings.Templates.Components.Validation.ChillersCompression fails in OpenModelica because of cyclically dependent parameters


AndreaBartolini:
The translation of the model `Buildings.Templates.Components.Validation.ChillersCompression` fails in **OpenModelica** because the following parameters of the chiller instance `chiWatCoo` have a cyclical dependency:
- `TConEnt_nominal`
- `cpCon_default`
- `staCon_default`

The `chiWatCoo` is instantiated at **TopLevel** with the following parameters and modifiers:
```modelica
parameter Buildings.Templates.Components.Data.Chiller datChiWatCoo(
  ...
  per(
    ...
    use_TConOutForTab=true))
"Water-cooled chiller parameters" annotation(...);

Buildings.Templates.Components.Chillers.Compression chiWatCoo(
  ...
  final dat=datChiWatCoo,
  ...)
  "Water-cooled chiller - CHW/CW pressure drops computed externally" annotation (...);
```

and the class `Buildings.Templates.Components.Chillers.Compression` is defined as following:
```modelica
partial model PartialChiller
  ...
  parameter Modelica.Units.SI.Temperature TConEnt_nominal
    "Design condenser entering fluid temperature";
 ...
  final parameter MediumCon.SpecificHeatCapacity cpCon_default=MediumCon.specificHeatCapacityCp(staCon_default)
    "Condenser cooling fluid default specific heat capacity";
	
  final parameter MediumCon.ThermodynamicState staCon_default=MediumCon.setState_pTX(
    T=TConEnt_nominal,
    p=MediumCon.p_default,
    X=MediumCon.X_default)
    "Condenser cooling fluid default state";
  ...
end PartialChiller

model Compression
  "Compression chiller"
  extends Buildings.Templates.Components.Interfaces.PartialChiller(
    ...	
    final TConEnt_nominal 
      = if dat.per.use_TConOutForTab 
         then dat.TCon_nominal - QCon_flow_nominal / mCon_flow_nominal / cpCon_default
         else dat.TCon_nominal,
   ...);
  ...
end Compression
```

[...]

PR #4171, opened 2025-03-20
Templates • Refactor AWHP plant template with modular HP model

AntoineGautier:
This addresses #4152.

TODO before merging:
- [x] Merge #4170 and the master into `issue4152_HpTemplateWithModular`
- [x] Fix Dymola warnings `Failed to evaluate and check the size declaration of pla.hp.dat.perHeaHp.PLRSor`
- [x] Add release notes

PR #4416, opened 2025-11-21
Fix cyclic parameter bindings

AntoineGautier:
This addresses #4407.


AntoineGautier:
@mwetter This is ready to merge.

The change from this PR cannot be applied to `maint_12.x`. It fixes an issue that was introduced with https://github.com/lbl-srg/modelica-buildings/pull/4171, which is not merged into `maint_12.x`. In its current state, `maint_12.x` does not include the cyclic parameter bindings and `Buildings.Templates.Components.Chillers.Validation.Compression` can be simulated with OMC.

AntoineGautier:
@mwetter Good point. I've removed the addition to the release notes. I've also removed the revision note in `PartialChiller`.

Issue #4396, opened 2025-10-23
Model Buildings.Examples.FanCoils.FourPipe fails verification in omc


AndreaBartolini:
The following variables of the model `Buildings.Examples.FanCoils.FourPipe` fail the verification vs the dymola reference file when simulated by using omc:
- `conFCU2.yHeaCoi`
- `conFCU1.yCooCoi`

Let's focus on the first, here below the results from the omc CI:

<img width="2100" height="1307" alt="Image" src="https://github.com/user-attachments/assets/cc0e27ef-8070-4bf2-9df3-6106c8dab581" />

and by zooming:

<img width="2090" height="1296" alt="Image" src="https://github.com/user-attachments/assets/491b05ee-5610-4306-abac-67305f8f9bf6" />

it is possible to see that there is some high-frequency transient that doesn't match the verification, turning on the row data we can see the following:

<img width="2097" height="1303" alt="Image" src="https://github.com/user-attachments/assets/f4af5760-28bd-4922-995d-8e3fed920ac4" />

so the first note is that the default communication interval (500 intervals) is too coarse to properly match the transient with the reference data.

To better analyze the transient, I reproduced the issue on my win11-pro machine, by using:
- Buildings master branch 4c81fe31cca
- dymola Version 2025x, 2024-10-11
- OpenModelica v1.26.0-dev-344-g81f39f4f0e under win
- OpenModelica 1.26.0~dev-421-gd95044f with ubuntu 24.04 under WSL

Both the OpenModelica versions return the same values, so in the following only the results from win are reported.

Here below the same transient simulated without the equidistant time grid:

<img width="1044" height="681" alt="Image" src="https://github.com/user-attachments/assets/4214d188-a072-4a63-815f-131e2ab6565a" />

The results from dymola and omc in any case don't match, one cause may be the high sensitivity of the PI controller `conFCU2.heaPI`:

<img width="1186" height="916" alt="Image" src="https://github.com/user-attachments/assets/b96932d0-df75-4584-9ab4-62d78c1731a0" />

an error of 0.05 K on the controlled temperature produces a control output variation of about 0.3 PerUnit.
Is there a particular reason for which this high-gain loop is needed?

AndreaBartolini:
Keep @casella in the loop.

karthikeyad-pnnl:
[...]

PR #4442, opened 2025-12-16
Issue4396 omc verification error v2

mwetter:
This closes https://github.com/lbl-srg/modelica-buildings/issues/4396

### DIFF
--- a/Buildings/Fluid/FixedResistances/BaseClasses/PlugFlowPipe.mo
+++ b/Buildings/Fluid/FixedResistances/BaseClasses/PlugFlowPipe.mo
@@ -273,27 +273,34 @@ equation
     annotation (Line(points={{-50,10},{-50,86},{0,86},{0,100}},
       color={191,0,0}));
 
-  connect(heaLos_b.port_b, vol_b.ports[1])
-    annotation (Line(points={{60,0},{78,0},{78,20}}, color={0,127,255}));
-  connect(vol_b.ports[2], port_b)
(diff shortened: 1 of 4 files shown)


Buildings/Controls/OBC/ASHRAE/G36/FanCoilUnits/Subsequences/SupplyAirTemperature.mo
```diff
@@ -215,7 +215,8 @@ block SupplyAirTemperature
     final controllerType=heaCoiConTyp,
     final k=kHeaCoi,
     final Ti=TiHeaCoi,
-    final Td=TdHeaCoi) if have_heaCoi
+    final Td=TdHeaCoi,
+    final r=12) if have_heaCoi
```
mwetter: @karthikeyad-pnnl : I think this `12` and the `11` below are specific to your example. I suggest to just set `10` for both, and not make them final. Let me know if you agree and I can make this change. Similarly for the `4` in Buildings/Controls/OBC/ASHRAE/G36/FanCoilUnits/Controller.mo

mwetter (commented):
@karthikeyad-pnnl : Please see the inline-question

Buildings/Controls/OBC/ASHRAE/G36/FanCoilUnits/Subsequences/SupplyAirTemperature.mo
karthikeyad-pnnl: @mwetter I dont have any objections to the changes you are suggesting. Thanks, Michael!

Issue #4432, opened 2025-12-02
Update load model in HP plant template validation


AntoineGautier:
This ticket is to update the load model in `Buildings.Templates.Plants.HeatPumps.Validation.AirToWater` based on the comments https://github.com/lbl-srg/modelica-buildings/pull/2700#issuecomment-3419756828 and https://github.com/lbl-srg/modelica-buildings/pull/2700#issuecomment-3601559208 that I reproduce here for reference.

> The load model should be revisited and the heating heat flow rate should be tracked, rather than the HW mass flow rate. The reason is that, contrary to, e.g., the chiller plant, there is no ∆p reset for the boiler plant, only a supply temperature reset. So, at low load, the control intent is to reset down the HWST to maximize the opening of the terminal control valves. Controlling for a mass flow rate taken from another simulation makes little sense because at low load the plant gets disabled due to the terminal control valves closing. If you switch to using, e.g., an explicit radiator model and track the heating heat flow rate, the HWST reset at low load keeps the control valve more open and the simulation exhibits the actual control intent, which is to operate at low temperature and high flow rate. The plant only gets disabled near zero load.

With this update, the load model will be consistent across all plant templates (chiller, boiler and HP).

[EDIT] A delay is also added to account for the total fluid volume in the distribution system.

[EDIT] While addressing this issue, another issue appeared related to equipment availability logic and concurrent stage transitions between heating and cooling modes. Event sequencing during stage transitions introduces a delay between when a heat pump is required to run in a given mode, and when it is actually commanded on in this mode. During this delay, the heat pump should not be considered available for the opposite mode. This ticket will refactor `Buildings.Templates.Plants.Controls.StagingRotation.EquipmentAvailability` to prevent this race condition.

PR #4450, opened 2026-01-06
Update load model in plant template validation

AntoineGautier:
This closes https://github.com/lbl-srg/modelica-buildings/issues/4432.


Buildings/Templates/Components/Loads/Load.mo
```diff
@@ -0,0 +1,381 @@
...
+      TLiqEnt_nominal,
+      X=MediumLiq.X_default) - MediumLiq.specificEnthalpy_pTX(
+      MediumLiq.p_default,
+      TLiqLvg_nominal,
+      X=MediumLiq.X_default)) * mLiq_flow_nominal
+    "Transmitted heat flow rate at design conditions";
+  parameter Real u_min(
+    max=1,
+    min=0,
+    unit="1")=0.1 "Fan minimum speed";
```
JayHuLBL: Since it is the minimum load that compares with the `u` (Fraction of design load)? Should we revise the comment?

Buildings/Templates/Components/Loads/LoadTwoWayValve.mo
```diff
@@ -0,0 +1,256 @@
...
+        group="Conservation equations"));
+  Buildings.Controls.OBC.CDL.Interfaces.RealInput u
+    "Fraction of design load"
+    annotation(Placement(transformation(extent={{-140,60},{-100,100}}),
+      iconTransformation(extent={{-140,60},{-100,100}})));
+  Buildings.Controls.OBC.CDL.Interfaces.BooleanInput u1
+    "System enable"
+    annotation(Placement(transformation(extent={{-140,20},{-100,60}}),
+      iconTransformation(extent={{-140,20},{-100,60}})));
+  Buildings.Templates.Components.Loads.Load loa(
```
JayHuLBL, line 100: The parameter `u_min` was not propagated. Should we do so?

Buildings/Templates/Components/Loads/Load.mo
AntoineGautier: I refactored the load component to logically use the PI output to modulate the air flow. 
This matches actual FCU controls in case of modulating valves and ECM fans.
image

"Minimum fan speed" is then adequate, even if we use here an ideal mass flow source on the air side – because of the constant circuit flow characteristic.

Doing so, an issue appeared with the equipment availability logic. I've updated the issue page to describe it and included a fix in [35d9582](https://github.com/lbl-srg/modelica-buildings/pull/4450/commits/35d9582bec30526920c9ee0ae09f12b1012d10d9).

Buildings/Templates/Components/Loads/LoadTwoWayValve.mo
AntoineGautier: Yes, done!

PR #4542, opened 2026-04-13
Do not assume the library is installed in a folder named Buildings.

adrpo:
It can include version numbers, etc.


mwetter:
@adrpo : Thanks, I will run it through or CI and also port it to the maintenance branches.

PR #4490, opened 2026-03-04
Issue4304 hybrid air source heat pump plant controls

karthikeyad-pnnl:
@AntoineGautier I am submitting this to resolve the comments on #4369

PR #4369, opened 2025-09-25
Hybrid air-source heat pump plant controls

AntoineGautier:
This is for #4304.

PR #4566, opened 2026-04-24
Issue4304 hybrid air source heat pump plant controls april2026

karthikeyad-pnnl:


karthikeyad-pnnl:
@AntoineGautier I am submitting this to address the comments on both #4369 and #4490. Additionally, please consider this a replacement for #4490, which can be closed without merge. I had to create this fresh PR since I made some undesired merges to my development branch on #4490.
Regarding the compilation issues on #4490, I was able to run the unit tests for OpenModelica, with the only reported errors being timeseries deviations. Can you please let me know if there are any issues reported on OCT, which I could not test out?

AntoineGautier:
@karthikeyad-pnnl Yes, the model `Templates.Plants.Controls.HeatPumps.Validation.AirToWater` now simulates with both OCT and OMC.
However, I've noticed that the feature branch does not include the latest changes from MBL master, particularly the ones from https://github.com/lbl-srg/modelica-buildings/pull/4513 that modified the input connectors of the modular heat pump models (see my email from March 24, 2026). Can you refactor the controller consistently?

karthikeyad-pnnl:
@AntoineGautier I have implemented the changes we discussed on 05/11. I can now confirm both the plant template validation models pass pedantic translation. However, `Buildings.Templates.Plants.HeatPumps.Validation.AirToWater` is failing the pedantic model check during the unit test due to overconstrained equations. I am working on identifying the root-cause, but would also appreciate any pointers if the error looks familiar to you. Thanks!

AntoineGautier:
> @AntoineGautier I have implemented the changes we discussed on 05/11. I can now confirm both the plant template validation models pass pedantic translation. However, `Buildings.Templates.Plants.HeatPumps.Validation.AirToWater` is failing the pedantic model check during the unit test due to overconstrained equations. I am working on identifying the root-cause, but would also appreciate any pointers if the error looks familiar to you. Thanks!

You need to propagate `energyDynamics = Modelica.Fluid.Types.Dynamics.FixedInitial` to all components that use a dynamic energy balance. 
The default for all these components in MBL is `DynamicFreeInitial`, which means the tools use the start attribute as a _fallback_ for initializing the state variables – which triggers a warning. 
Using `FixedInitial` instead makes the choice of the start attribute for initialization explicitly assumed – it's equivalent to adding `fixed=true` to each variable.

Also, why model the fluid junctions in steady state? If there is no strong reason, best is to uniformly propagate `energyDynamics` to all components of the model.

One last request: can you merge the latest commit of MBL master? That would help for the review as I will focus on the diff, which is currently quite large.

AntoineGautier:
@karthikeyad-pnnl I'd like to merge this branch into 
lbl-srg:issue4304_HybridAirSourceHeatPumpControls and do the detailed review in https://github.com/lbl-srg/modelica-buildings/pull/4369. 
This way, we have a clean diff against the target master branch as a review basis.

Does that sound good to you?

Before I do so, could you:

- [ ] Propagate `energyDynamics = Modelica.Fluid.Types.Dynamics.FixedInitial` to _**all components that use a dynamic energy balance**_ in `Buildings.Templates.Plants.HeatPumps.Validation.HybridAirToWater`. (Only the fluid junction parameters were updated via [b61ed86](https://github.com/lbl-srg/modelica-buildings/pull/4566/commits/b61ed861a052e733455227c0ccdd96d9130eb36a), so the initialization warnings persist.)
- [ ] Refactor `Buildings.Templates.Plants.Controls.HeatPumps.AirToWater` to use the same connector variables for the 4-pipe HP as the ones from `Buildings.Fluid.HeatPumps.ModularReversible.TableData2DLoadDepSHC`. The later was refactored to use the control points specified in G36 Addendum L, see https://github.com/lbl-srg/modelica-buildings/issues/4507. In `issue4304_HybridAirSourceHeatPumpPlantControls_April2026`, the controller still uses an integer signal for the 4-pipe HP operating mode command, and the conversion into Booleans is done in `Buildings.Templates.Components.BaseClasses.PartialHeatPumpTableData2DLoadDepSHC`. This creates inconsistencies between the controller output points (HRC VS 4-pipe HP), and is also inconsistent with G36.

karthikeyad-pnnl:
@AntoineGautier Sorry for the late response. Yes, please feel free to merge the branch for the review. I have updated the control modules and the template validation as suggested. I am still seeing pedantic check errors in `Buildings.Templates.Plants.HeatPumps.Validation.AirToWater`, and would appreciate any guidance on identifying the root cause. Thanks!

AntoineGautier:
>  I am still seeing pedantic check errors in `Buildings.Templates.Plants.HeatPumps.Validation.AirToWater`

This is another issue. The changes solved the initialization warnings reported before.

This new issue is peculiar: I'm guessing a false positive by Dymola pedantic check, because there is no error or warning when running "Pedantic check and simulate":
image
The error only appears when running "Pedantic check" alone...
OCT can properly translate that model.

PR #4571, opened 2026-04-29
Merged correction to weather data interpolation

### LINKED ISSUE #4417: Incorrect interpolation of wind direction in ReaderTMY3
I found an issue in Buildings.BoundaryConditions.WeatherData.ReaderTMY3 related to the handling of wind direction from TMY3 weather files.
EPW/MOS file is reported hourly and the wind direction is within the range 0–360 degrees. In the ReaderTMY3 implementation, the wind direction column is passed through a continuous interpolation. This causes incorrect interpolation whenever the wind direction crosses the 0°/360° boundary.
For example, if in the weather file contains first hour 10° and second hour 350° for the wind direction, it will be more reasonable to interpolate as rotating counter-clockwise by −20° (10° → 350° via 360° → 350°) for the shortest distance, but current Reader performs direct linear interpolation in raw degrees and produce 10° → 180° → 350°. This will produce a large non-physical jump in wind direction, and will affect wind pressure calculations.

Please see below screenshots for this behavior.

Image

**Expected behavior**
Wind direction should be interpolated using circular interpolation, not linear interpolation in degrees.

**Version**
 - Modelica Buildings Library: 12.1.0

mwetter:
This merges IBPSA issue2068_weatherFileConversion.
This closes #4417


mwetter:
@EttoreZ : Can you please check the reference results, and update them as needed. There are some substantial changes for which we need an explanation. For example, `EnergyPlus_24_2_0.Examples.SmallOffice.ASHRAE2006Spring` has a non-trivial change in `EHea` (blue):
image

Also, `IdealHeatingCoolingSpring` has a large change in `EHea` (brown):
image

The following command runs the examples that have different results:
```
../bin/runUnitTests.py -s "Buildings.{Obsolete.ThermalZones.EnergyPlus_9_6_0.Examples,ThermalZones.EnergyPlus_24_2_0.Examples,ThermalZones.Detailed.Validation,Examples,Utilities.Plotters.Examples,Airflow.Multizone.Examples,Air.Systems.SingleZone.VAV.Examples,Fluid.Sources.Examples,Controls.Predictors.Examples,Fluid.FixedResistances.Validation.PlugFlowPipes}"
```

mwetter:
@EttoreZ : This error may be caused by the functions `windPressureLowRise` and `windPressureProfile`. Their unit tests should be changed to test for a larger range of wind angles, and see if these functions are correct for the wind angle that is now received with the new implementation of the weather data.

EttoreZ:
@mwetter thank you for the heads up. I will take a look at their unit tests, and see if the behavior is as intended.

PR #4607, opened 2026-05-18
Remove all remaining occurence of cardinality.

qlambert-pro:
`cardinality` is a deprecated construct in the modelica language and all remaining use case in the library are covered by either `mustBeConnected` or `mayOnlyConnectOnce`.

### DIFF
--- a/Buildings/Fluid/Sources/BaseClasses/PartialAirSource.mo
+++ b/Buildings/Fluid/Sources/BaseClasses/PartialAirSource.mo
@@ -21,7 +21,12 @@ partial model PartialAirSource
     each h_outflow(nominal=Medium.h_default),
     each Xi_outflow(each nominal=0.01))
     "Fluid ports"
-    annotation (Placement(transformation(extent={{90,40},{110,-40}})));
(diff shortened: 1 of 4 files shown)


qlambert-pro:
I just noticed that because the semantics of `cardinality` and `mayOnlyConnectOnce` are slightly different, this PR introduces a regression in `Buildings.DHC.Plants.Combined.AllElectricCWStorage`, because of the following statements:

```
  connect(bouConWatCoo.ports[1], pumConWatCoo.port_a);
  connect(coo.port_b, pumConWatCoo.port_a);
```

the `cardinality(bouConWatCoo.ports)` does return `1`, because it only appears once in a `connect` equation, and 
the `assert` in `Buildings.Fluid.Sources.BaseClasses.PartialSource` doesn't trigger.

But for the purpose of `mayOnlyConnectOnce`, `bouConWatCoo.ports` appears in a set of `3` which is then considered an error.

mwetter:
Thanks for the PR. We will further process it on branch `issue4607_cardinality`.

PR #4608, opened 2026-05-19
Remove use of cardinality

mwetter:
This removes the use of the depracated `cardinality` operator.

Todo:
- [x] Correct the regression reported in https://github.com/lbl-srg/modelica-buildings/pull/4607#issuecomment-4476460369


mwetter:
For `Sources.BaseClasses.{PartialSource,PartialAirSource}`, the `cardinality` operator cannot be replaced with `mustOnlyConnectOnce` (https://specification.modelica.org/maint/3.6/annotations.html#S5.SS2.p5) because this component is often used to set a reference pressure, in which case it is connected to a connection set with three ports. If we were to add `mustOnlyConnectOnce` then for each use of the pressure source, one would need to add a junction volume.

This PR also removes the protected parameter in the above models, which is only used to set the `max` attribute to `inf` and the `min` attribute to `-inf` and hence has no effect.

Issue #4609, opened 2026-05-19
Hardcode issue of the ASHRAE2006 AHU model


howedo:
**Describe the bug**

The bug is for Buildings.Examples.VAVReheat.BaseClasses.ASHRAE2006.

When this ASHAREA2006 AHU model is adapted for a system with a different number of thermal zones other than **FIVE**, a translation error occurs, stating that the zone number and AHU dimension do not match.  

Although the parameter '_numZon_' is declared at the beginning of this model and used in other components (e.g., Controls.RoomVAV, Buildings.Controls.OBC.CDL.Routing.RealScalarReplicator, etc), the number of inputs is hardcoded as '_nin=5_' in the component 
'Controls.DuctStaticPressureSetpoint' (Line 34 in the Text View Model).

**System information** 

- Modelica Buildings Library: 12.1.0
- Modelica simulation platform: Dymola 2025x Refresh 1
- Operating system: Windows 64-bit

**Solution**

Changing the declaration from 'nin=5' to 'nin=numZon' in the PressureSetpoint component solves the issue.

mwetter:
Thanks for reporting. I will correct it on the `maint_12.x` and the `master` branch.

PR #4617, opened 2026-05-22
Changed assignment for number of zones in ASHRAE2006. (maint_11.x)

mwetter:
For #4609


mwetter (commented):
Ready to merge when the tests pass.
It looks like the Dymola image needs to be updated first on `maint_11.x`.

PR #4653, opened 2026-06-26
Added nominal attributes to state variables

mwetter:
This merges IBPSA, issue2132_dynamicStates


mwetter:
@AntoineGautier : This PR essentially sets missing nominal attributes. For some models, the code generation now selects different iteration variables. I run all tests and they succeeded except for the models
`Buildings.Templates.Plants.HeatPumps.Components.Validation.PumpsPrimaryDedicated`
`Buildings.Templates.AirHandlersFans.Validation.VAVMultiZone`
`Buildings.Templates.Components.Validation.Pumps`

They all fail with the message
```
Pedantic translation of Buildings.Templates.Components.Validation.Pumps

The sorting of parameter bindings indicates that the binding
pumMul.dat = datPumMul;
shall be used to solve for datPumMul.per[1].pressure.V_flow[1].
```
Can you please look into what may cause this. I would think `datPumMul.per[1].pressure.V_flow[1]` should be a parameter expression.

Once that is resolved, the PR is ready to merge.

AntoineGautier:
@mwetter Definitely a false positive. It appears to be due to the change in `Buildings.Fluid.Interfaces.ConservationEquation` and the following equation in `Buildings.Fluid.Movers.BaseClasses.PartialFlowMachine`:
```mo
  parameter Modelica.Units.SI.MassFlowRate _m_flow_nominal=
    max(eff.per.pressure.V_flow)*rho_default
    "Nominal mass flow rate";
```
As you said, `datPumMul.per[1].pressure.V_flow[1]` is an explicit parameter expression with nothing to "solve for".
OCT has no problem translating these models.
I've pushed a fix in 146e975. I'll report the issue to DS and we can revert it once that's solved.

mwetter:
@AntoineGautier : Thanks for the fix https://github.com/lbl-srg/modelica-buildings/commit/146e975c04c91739e6109779b06a0a1be3c8a00b. But unfortunately, with this change, the models
`Buildings.Fluid.HeatPumps.ModularReversible.Examples.TableData2DLoadDepSHC1Only` and
`Buildings.Fluid.HeatPumps.ModularReversible.Examples.TableData2DLoadDepSHC1And2` won't translate with omc and also not display in OMEdit. 
If I delete in `TableData2DLoadDepSHC1Only` the two instances of `Templates.Components.Pumps.Multiple` then the model displays in OMEdit, so the problem is caused by this model.

AntoineGautier:
@mwetter I've pushed an even more compiler-friendly version of the fix in fb15b4e. Hopefully, this will avoid any regression for all tools.

mwetter:
@AntoineGautier : That's a hard one! Now, `Buildings.Templates.Plants.HeatPumps.Validation.AirToWater` freezes dymola on the model check and one the translation. The message is
"Failed to evaluate and check the size declaration of datAll.pla.pumChiWatHrc.pressure.dp"

AntoineGautier:
@mwetter That's because we now have multiple hacks piling up on each other. In `Buildings.Templates.Plants.HeatPumps.Data.HeatPumpPlant` we have the following:
```mo
  /* HACK(AntoineGautier):
  The bindings for per.pressure in the single pump record declarations below
  are identical to the default bindings in the PumpSingle class definition.
  However, they are needed for Dymola (version 2024x Refresh 1) that fails
  to evaluate the size of per.pressure.dp otherwise.
  */
  parameter Buildings.Templates.Components.Data.PumpSingle pumChiWatHrc(
    per(pressure(
      V_flow=if pumChiWatHrc.typ<>Buildings.Templates.Components.Types.Pump.None then 
      {0, 1, 2} * pumChiWatHrc.m_flow_nominal / cfg.rhoChiWat_default else {0,0,0},
      dp=if pumChiWatHrc.typ<>Buildings.Templates.Components.Types.Pump.None then 
      {1.14, 1, 0.42} * pumChiWatHrc.dp_nominal else {0,0,0})))
```
I tried to extract the inline array expressions and use a propagated record instance instead (as in `PumpSingle`) but the translation of `Buildings.Templates.Plants.HeatPumps.Validation.AirToWater` still freezes.
I also tried to keep inline expressions in both locations, since Dymola accepts them in the single-pump validation. This didn't resolve the plant template issue.

The cleanest solution would be to make `_m_flow_nominal` a public ("Advanced") parameter in `Buildings.Fluid.Movers.BaseClasses.PartialFlowMachine`. Then I can simply bind it to the nominal values that the template components require anyway. This breaks the relationship to the pressure curve record and the bindings resolve easily – the hack from my previous commits is then not needed. Is that an option?

AntoineGautier:
@mwetter All things considered, I would like to go back to db772f6 and introduce the following model for use in the Templates package only.

```mo
within Buildings.Templates.Components.BaseClasses;
model MoverSpeedControlled_y
  extends Buildings.Fluid.Movers.SpeedControlled_y(
    final _m_flow_nominal=m_flow_nominal);
  parameter Modelica.Units.SI.MassFlowRate m_flow_nominal;
annotation(Documentation(
  info="<html>
<p>
  This is a compiler-friendly version of
  <a href=\"modelica://Buildings.Fluid.Movers.SpeedControlled_y\">
    Buildings.Fluid.Movers.SpeedControlled_y</a> which facilitates the
  propagation of the mover performance curves by binding
  <code>_m_flow_nominal</code> to a top-level parameter rather than to an
  expression that depends on the performance data record. See
  <a href=\"https://github.com/lbl-srg/modelica-buildings/pull/4653\">
    pull request #4653</a> for a discussion of the motivation for this model.
</p>
</html>"));
end MoverSpeedControlled_y;
```

This will isolate the issue inside the Templates package. This model can be removed and `Buildings.Fluid.Movers.SpeedControlled_y` can be used instead once Dymola's issue is addressed.
All the proposed hacks become unnecessary as a result.

On top of that, would you agree to declare `dp` inside `Buildings.Fluid.Movers.BaseClasses.Characteristics.flowParameters` as an unknown size array:
```mo
  parameter Modelica.Units.SI.PressureDifference dp[:]  // <--- instead of dp[size(V_flow, 1)] currently
```
and add a statement inside `Buildings.Fluid.Movers.BaseClasses.PartialFlowMachine` to assert size equality.
The declaration with the specified size is challenging for compilers on complex models because they must assess the size at translation time. Dymola fails during that assessment, hence the original hack within `Buildings.Templates.Plants.HeatPumps.Data.HeatPumpPlant`:
```mo
  /* HACK(AntoineGautier):
  The bindings for per.pressure in the single pump record declarations below
  are identical to the default bindings in the PumpSingle class definition.
  However, they are needed for Dymola (version 2024x Refresh 1) that fails
  to evaluate the size of per.pressure.dp otherwise.
  */
  parameter Buildings.Templates.Components.Data.PumpSingle pumChiWatHrc(
    per(pressure(
      V_flow=if pumChiWatHrc.typ<>Buildings.Templates.Components.Types.Pump.None then
      {0, 1, 2} * pumChiWatHrc.m_flow_nominal / cfg.rhoChiWat_default else {0,0,0},
      dp=if pumChiWatHrc.typ<>Buildings.Templates.Components.Types.Pump.None then
      {1.14, 1, 0.42} * pumChiWatHrc.dp_nominal else {0,0,0})))
```

mwetter:
@AntoineGautier : The proposed changes look good to me, assuming they don't break anything unknown for the various tools. Please go ahead an implement them so we can run the tests.

AntoineGautier:
@mwetter I've committed the changes in 71d068a and 5af3619.
Let's see if all tests pass.
The latter commit, with the unknown-size array declaration, touches classes shared with the IBPSA Library, and should probably be made there first instead.

PR #4693, opened 2026-08-17
Issue4692 staging stuck

### LINKED ISSUE #4692: Chiller staging process stuck
In the chiller plant template model test (af09940), running the model `Buildings.Templates.Plants.Chillers.Validation.AirCooled` shows that the staging process stuck as below:

Image

The expected behaviour is as below:

Image

This issue is to correct the bug.

JayHuLBL:
This closes #4692.


AntoineGautier (approved):
That looks good. 
Just pushed a commit that fixes some typos and an incorrect component description.

Not related to this PR, but I've noticed several Edge+Latch occurrences e.g. in `Buildings.Controls.OBC.ASHRAE.G36.Plants.Chillers.Generic.PlantEnable.EnableDevices`. Latch is already inherently edge-sensitive: chaining an explicit Edge block upstream just reproduces a transition that Latch detects itself.
For clarity and readability, all these occurrences should be simplified, using only a Latch component.

JayHuLBL:
@AntoineGautier We can simplify the implementation in another issue.

@mwetter The PR is ready to merge if there is no more comments.

